PR #2, opened 2015-11-26
Create DnD menu only with supported actions

palinek:
see [comment](https://github.com/lxde/lxqt/issues/446#issuecomment-159903604)
> - if the DND is originated in plugin-mainmenu (QMenu), then after the drop anything can be selected, but the action is always Qt::CopyAction in dropMimeData()

OT: @PCMan shouldn't the lxde/libfm-qt repo be accesible for r/w also for other lxqt people?

### DIFF
--- a/src/dndactionmenu.cpp
+++ b/src/dndactionmenu.cpp
@@ -22,10 +22,19 @@
 
 namespace Fm {
 
-DndActionMenu::DndActionMenu(QWidget* parent): QMenu(parent) {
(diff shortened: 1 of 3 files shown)


PCMan:
@palinek Forgot to set the permissions. Will do it. Thanks!

PCMan:
Merged into d5bfd04c9f2c2dff890a3d964adc29fed4cddc09.

PR #8, opened 2015-12-19
Fix memory leak if reading stream for image failed

philippwiesemann:
If an error occurred while reading a thumbnail image from its stream then the allocated buffer was not deleted. This was fixed by using a smart pointer which will always delete the buffer later.

Thanks!

### DIFF
--- a/src/thumbnailloader.cpp
+++ b/src/thumbnailloader.cpp
@@ -21,6 +21,7 @@
 #include "thumbnailloader.h"
 #include <new>
 #include <QByteArray>
+#include <QScopedArrayPointer>
 
 namespace Fm {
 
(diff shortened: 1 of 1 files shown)


palinek:
GTM

paulolieuthier:
The image may be still used after the end of the function. Are you sure we should delete the buffer? From the [docs](http://doc.qt.io/qt-5/qimage.html#QImage-3):

> The buffer must remain valid throughout the life of the QImage and all copies that have not been modified or otherwise detached from the original buffer.

We must provide a cleanup function, since QImage destructor won't delete it.

philippwiesemann:
The documentation for `loadFromData()` does not have this note.

paulolieuthier:
You are correct.

PR #9, opened 2015-12-26
Fix target language in Lithuanian translation file

philippwiesemann:
The file's target language was set to "zh_TW" instead of "lt_LT".

Thanks!


paulolieuthier:
Thanks!

PR #10, opened 2015-12-31
Fix file search with smaller than option

philippwiesemann:
It was searching for files larger than the provided size.

Thanks!


palinek:
GTM

PCMan:
Merged. Thanks!

PR #11, opened 2016-01-01
Let the style engine draw the text selection rectangle.

tsujan:
More flexible Qt style engines, like QtCurve and Kvantum, have more elegant ways of drawing the text selection rectangle than just filling it with the highlight color. They may also have hovering/inactiveness effects. Besides, according to my experience with QStyle, it's better to leave the style related stuff to the active engine as far as possible (it's a long story).

Here is an example with Kvantum:
selected-items

If this PR is accepted, I'll make another one for Desktop.

### DIFF
--- a/src/folderitemdelegate.cpp
+++ b/src/folderitemdelegate.cpp
@@ -178,11 +178,23 @@ void FolderItemDelegate::drawText(QPainter* painter, QStyleOptionViewItemV4& opt
 
   QPalette::ColorGroup cg = opt.state & QStyle::State_Enabled ? QPalette::Normal : QPalette::Disabled;
   if(opt.state & QStyle::State_Selected) {
-    painter->fillRect(selRect, opt.palette.highlight());
+    if(!opt.widget)
+      painter->fillRect(selRect, opt.palette.highlight());
     painter->setPen(opt.palette.color(cg, QPalette::HighlightedText));
   }
   else
     painter->setPen(opt.palette.color(cg, QPalette::Text));
+
+  if (opt.state & QStyle::State_Selected || opt.state & QStyle::State_MouseOver) {
+    if (const QWidget* widget = opt.widget) { // let the style engine do it
+      QStyle* style = widget->style() ? widget->style() : qApp->style();
+      QStyleOptionViewItemV4 o(opt);
+      o.text = QString();
+      o.rect = selRect.toAlignedRect().intersected(opt.rect); // due to clipping and rounding, we might lose 1px
+      o.showDecorationSelected = true;
+      style->drawPrimitive(QStyle::PE_PanelItemViewItem, &o, painter, widget);
+    }
+  }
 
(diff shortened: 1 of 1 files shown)


src/folderitemdelegate.cpp
```diff
@@ -178,12 +178,24 @@ void FolderItemDelegate::drawText(QPainter* painter, QStyleOptionViewItemV4& opt
...
-    painter->fillRect(selRect, opt.palette.highlight());
+    if(!opt.widget)
+      painter->fillRect(selRect, opt.palette.highlight());
     painter->setPen(opt.palette.color(cg, QPalette::HighlightedText));
   }
   else
     painter->setPen(opt.palette.color(cg, QPalette::Text));
 
+  if (opt.state & QStyle::State_Selected || opt.state & QStyle::State_MouseOver) {
+    if (const QWidget* widget = opt.widget) { // let the style engine do it
```
f2404: What does this  condition mean?

tsujan: The first condition says, "if the item is selected or the mouse cursor is over it".
The second one says, "if there is a real widget". QStyleOptionViewItem has a public member `const QWidget *widget` (-> qstyleoption.h). Just as a precaution, if there isn't a real widget, the rect will just be filled as before.

f2404: I'm just concerned that the second condition will always be true due to the assignment
```const QWidget* widget = opt.widget```

Comment (tsujan) [inline src/folderitemdelegate.cpp]: I don't know of a situation, where it's not true but it's a necessary precaution PCMan has used a few lines later (near the end of the file) and *the assignment doesn't make it true automatically*.

Comment (f2404) [inline src/folderitemdelegate.cpp]: How can it be not true?
You can always assign a value to a pointer, and this will result in 'true' as the condition statement result.

@PCMan Could you please explain the meaning of the following statement?
```c++
if (const QWidget* widget = opt.widget) {
```

tsujan: @f2404 
In C++, `if (const QWidget* widget = opt.widget)` means:
````
if (opt.widget != NULL) {
  const QWidget* widget = opt.widget;
  .....
}
````
However, this discussion isn't related to the present commit.

f2404: Ok, it's not related - but could you guide me to a standard where it's described, for my education?
I was sure that the behavior was similar to C - i.e. the assignment wouldn't imply checking against NULL.

tsujan: It's OK. Count on Google and you'll easily find nice explanations (I don't know of a specific one). I remove the comments just for more readability.

jleclanche:
gtm

paulolieuthier:
GTM.

PCMan:
Merged. Thank you for working on this.
I really hate the QStyle painting thing. lol

PR #13, opened 2016-01-14
FolderView methods for getting and setting item delegate margins

tsujan:
This simple commit paves the way for adding an option for (minimum) icon spacing to PCManfm-qt and other apps that subclass FolderView. For now, I've used it with success to prevent large gaps from taking shape at desktop bottom (https://github.com/lxde/pcmanfm-qt/issues/97#issuecomment-170671223).

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -389,7 +389,8 @@ FolderView::FolderView(ViewMode _mode, QWidget* parent):
   autoSelectionTimer_(nullptr),
   selChangedTimer_(nullptr),
   fileLauncher_(nullptr),
-  model_(nullptr) {
+  model_(nullptr),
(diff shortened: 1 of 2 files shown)


jleclanche:
gtm

tsujan:
@jleclanche Thanks! So fast :)

paulolieuthier:
GTM.

tsujan:
Just a note: Don't worry if you've compiled and *installed* libfm-qt with this commit and then, desktop icons disappear after restarting pcmanfm-qt because they will appear again if you recompile the latter against the new libfm-qt. *Someone please test this too!*

tsujan:
> Just a note: Don't worry if ...

Oh, that was only a Debian issue -- some incomplete updates.

PCMan:
Merged. Thanks!

PR #14, opened 2016-01-24
Add (folder) custom actions to foldermenu.

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/297.
The new class `CustomFolderAction` is a duplication of `CustomAction` in `filemenu_p.h` because including that header in `foldermenu.cpp` didn't seem wise. The new private header `foldermenu_p.h` is just for it.

### DIFF
--- a/src/customaction_p.h
+++ b/src/customaction_p.h
@@ -17,37 +17,31 @@
  *
  */
 
-#ifndef FM_FILEMENU_P_H
-#define FM_FILEMENU_P_H
-
-#include "icontheme.h"
-#include <QDebug>
+#ifndef FM_CUSTOMACTION_P_H
+#define FM_CUSTOMACTION_P_H
 
 namespace Fm {
 
-class AppInfoAction : public QAction {
-  Q_OBJECT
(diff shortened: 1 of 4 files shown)


tsujan:
BTW, IMO it's better to add something like the lines with the comment `// before all custom actions` to filemenu too. In this way, custom actions would be more easily recognized.

PCMan:
Thanks.
1. Including filemenu_p.h in foldermenu.cpp is OK since it's a private header file which is only used in our internal implementation. This breaks nothing. So there's no need for the duplication.
2. For your purpose, I'd rather create customaction_p.h, and put the CustomAction class in it, and then, include that file in both filemenu.cpp and foldermenu.cpp.
What do you think?

tsujan:
Thanks!
I agree about `customaction_p.h`. Should I do that here?

PCMan:
Sounds good.

tsujan:
Done! Added a separator before custom actions of filemenu too.

PCMan:
merged.

PR #15, opened 2016-01-25
Fix compiling with Qt < 5.4

philippwiesemann:
Thanks!

### DIFF
--- a/src/xdndworkaround.cpp
+++ b/src/xdndworkaround.cpp
@@ -24,12 +24,12 @@
 #include <QX11Info>
 #include <QMimeData>
 #include <QCursor>
+#include <QWidget>
 
 // This part is for Qt >= 5.4 only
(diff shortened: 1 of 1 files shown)


src/xdndworkaround.cpp
```diff
@@ -43,15 +43,17 @@
 
 #endif // (QT_VERSION >= QT_VERSION_CHECK(5, 4, 0))
 
-XdndWorkaround::XdndWorkaround():
-  lastDrag_(nullptr) {
```
paulolieuthier: `lastDrag_` is now only initialized when Qt > 5.4?

palinek: Yes, because it isn't defined for Qt < 5.4 -> https://github.com/lxde/libfm-qt/blob/master/src/xdndworkaround.h#L74

PR #16, opened 2016-02-03
Focus automatically on the first field of the filesearch dialog

npmiller:
This makes it possible to use the file search dialog only with the keyboard for simple searches.


PCMan:
Merged. Thanks!

PR #18, opened 2016-02-14
Redirect focus operation on the folderview to the actual widget

npmiller:
With this addition, focus operation used on the `FolderView` widget will be properly redirected to the underlying widget.

Related documentation: https://doc.qt.io/qt-5.5/qwidget.html#setFocusProxy


PCMan:
Merged. Thanks!

PR #19, opened 2016-03-07
Enable renaming in Properties dialog

tsujan:
This commit fixes https://github.com/lxde/pcmanfm-qt/issues/324.

### DIFF
--- a/src/filepropsdialog.cpp
+++ b/src/filepropsdialog.cpp
@@ -423,6 +423,29 @@ void FilePropsDialog::accept() {
     op->run();
   }
 
+  // Renaming
+  if(singleFile) {
+    QString new_name = ui->fileName->text();
+    if(QString::fromUtf8(fm_file_info_get_disp_name(fileInfo)) != new_name) {
+      FmPath* path = fm_file_info_get_path(fileInfo);
+      GFile* gf = fm_path_to_gfile(path);
+      GFile* parent_gf = g_file_get_parent(gf);
+      GFile* dest = g_file_get_child(G_FILE(parent_gf), new_name.toLocal8Bit().data());
+      g_object_unref(parent_gf);
+      GError* err = NULL;
+      if(!g_file_move(gf, dest,
+                      GFileCopyFlags(G_FILE_COPY_ALL_METADATA |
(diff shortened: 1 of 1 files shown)


PCMan:
Thanks for the new feature. Other file manager I know, such as Dolphin, do support this, but they don't add a "Rename" button. Instead, they remember if the name is edited, and rename the file when the "OK" button is clicked as needed. This seems to be a more natural way. What do you think?

tsujan:
> Other file manager I know, such as Dolphin, do support this, but they don't add a "Rename" button...

Yes, I know that but I intentionally made the lineedit read-only and added the Rename button to prevent any inadvertent renaming.

Since name is the most important property a file has, in file managers, renaming is done only after pressing a shortcut or right-clicking; IMO, renaming so easily by adding a character to a lineedit and pressing OK defeats the main purpose of a properties dialog.

That said, if you still insist on similarity to other FMs, I'll change it. Please tell me your final decision!

npmiller:
The UI for this is a bit strange, to do the same thing, having a checkbox disabling/enabling the QLineEdit would be much clearer, instead of playing on read-only and focus.

Actually if you really want to protect the file name, maybe a simple label and using the button to call the rename dialog would be simpler (though maybe that's a bit too many dialogs).

I personally don't mind the QLineEdit being editable right away as long as:

* It doesn't get the focus straight away
* Ctrl+Z works as expected to get back the original file name if I mistype

tsujan:
@npmiller
Checkboxes are suitable for settings, not this job.
I wait for PCMan's decision. If he prefers a prop dialog similar to that of other DEs, I'll make the lineedit always editable.

PCMan:
@tsujan The proposed UI seems to be reasonable.
However, almost all of the existing file managers do not work this way.
Windows explorer, KDE dolphin, Gnome nautilus, and XFCE thunar all rename the file from the property dialog without any additional button clicks.
So, IMHO, it might still be more familiar for the users if we have the same behavior. Is this OK for you.

tsujan:
@PCMan
OK, removed the Rename button.

PCMan:
Merged.

PR #20, opened 2016-03-19
Cancel indexing job when closing filepropsdialog

npmiller:
Fixes: https://github.com/lxde/pcmanfm-qt/issues/328

`fm_job_cancel` is enough clean up because it will trigger the finished signal and the usual job cleanup will then take place.

### DIFF
--- a/src/filepropsdialog.cpp
+++ b/src/filepropsdialog.cpp
@@ -69,17 +69,26 @@ FilePropsDialog::FilePropsDialog(FmFileInfoList* files, QWidget* parent, Qt::Win
 }
 
 FilePropsDialog::~FilePropsDialog() {
-  delete ui;
-
   if(fileInfos_)
     fm_file_info_list_unref(fileInfos_);
-  if(deepCountJob)
-    g_object_unref(deepCountJob);
+
+  // Stop the timer if it's still running
   if(fileSizeTimer) {
     fileSizeTimer->stop();
     delete fileSizeTimer;
     fileSizeTimer = NULL;
   }
+
+  // Cancel the indexing job if it hasn't finished
+  if(deepCountJob) {
+    g_signal_handlers_disconnect_by_func(deepCountJob, (gpointer)G_CALLBACK(onDeepCountJobFinished), this);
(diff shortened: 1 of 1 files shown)


tsujan:
Please change it to this:
````
  if(deepCountJob) {
    g_signal_handlers_disconnect_by_func(deepCountJob, (gpointer)G_CALLBACK(onDeepCountJobFinished), this);
    fm_job_cancel(FM_JOB(deepCountJob)); // emits finished signal
    g_object_unref(deepCountJob);
  }
````
https://github.com/lxde/pcmanfm-qt/issues/328#issuecomment-198870272 may be fixed in this way

And would you please test the commit afterward? Here, I can't reproduce the crash, perhaps, because my laptop is too fast.

tsujan:
`g_object_unref(deepCountJob);` my not bee needed.

Ilya87:
How to reproduce: open your terminal app, killall pcmanfm-qt, gdb pcmanfm-qt, run, properties on your home folder or other folder with many files, press Esc button as soon as possible after clicking on properties.

npmiller:
Ok, I've updated the fix, it should be good now, there's several things going on:

* Stop the timer before cancelling the job, should be fine the other way around since the timer callback checks if the job is cancelled, but I think it's safer this way.
* @tsujan suggestion and handling the cleanup in the destructor instead of using the finished callback
* Only delete the ui at the last step of the destructor

Both the two last point seem to fix the issue, this is because the finished callback accesses the filesprops dialog window which seems to be in a bad state after the ui has been deleted (not exactly surprising). So either deleting the ui after cancelling the job, or not accessing the window after cancelling the job seem to work. Doing both here just to be on the safe side.

tsujan:
OK, pressing `Esc` was the solution (as soon as I wanted to press a dialog button, indexing was finished). I saw the crash but after https://github.com/lxde/libfm-qt/pull/20#issuecomment-198908970, I can't see it anymore. @npmiller's new commit should fix it too. Please test it!

Ilya87:
No crash!

PR #23, opened 2016-04-23
Fix four typos in messages

philippwiesemann:
They were also fixed in the translation files to keep the translations valid.

Thanks!


paulolieuthier:
Thanks!

PR #25, opened 2016-05-06
Fix memory leaks in file menu

philippwiesemann:
Its sub-menu and one entry had no parent and were therefore not deleted.

Found by Valgrind.

Thanks!


tsujan:
Thanks! As Qt doc says, "The ownership of action is not transferred to this QWidget" (in this case, QMenu) with `addAction()`. So, GTM.

tsujan:
@philippwiesemann Please keep up the good work with Valgrind!

PR #27, opened 2016-05-12
Fix DND on view mode change

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/351.

`FolderView::setViewMode()` uses `QListView::setMovement(QListView::Static)`. If listView is already visible, the latter function not only will disable the drop events for its viewport but also will cause a delayed re-layout, which will prevent accepting of drops even after enabling drop events for its parent view in the former function. This commit re-enables the drop events when listView is visible.


agaida:
Tested and GTM

PR #28, opened 2016-05-15
build: Use external translations

palinek:
Also use the lxqt_translate_ts/lxqt_translate_desktop from the liblxqt. But this
is only a build dependency, no lxqt component is needed in runtime.


agaida:
Tested and GTM

palinek:
@PCMan I'm not going to merge this, as I remeber you were against the (build) dependency on liblxqt. But I don't remember the reasoning...

palinek:
https://github.com/lxde/pcmanfm-qt/pull/353#issuecomment-219615948 -> merging..

PR #29, opened 2016-05-27
Fix memory leak in file open menu

philippwiesemann:
`AppInfoAction` stores a `GAppInfo` and updates the reference count in the constructor and destructor. Nevertheless, the `GAppInfo`s were later not finalized because they still were referenced from their original list.

Found by Valgrind.

Thanks!


tsujan:
The logic seems sound but please don't merge it until I test it!

tsujan:
Tested. It's all right. GTM.

As @philippwiesemann explained, `AppInfoAction` increases the reference count with `g_object_ref()` `in filemenu_p.h`, so we need to decrease it here and also free the memory used by the `GList`.

agaida:
GTM

PR #30, opened 2016-06-06
Feature: Adding ID3 'Title' 'Artist' Column

YWJamesLin:
*This is a testing feature, just wait for this able to work properly ,then be able to be accepted.

Since this is a file manager, I'd like to manage my music use pcmanfm, but I cannot view ID3Tag with this FM. Thus, I tried to make this feature come true. I use id3lib to perform ID3Tag reading, then add 2 testing column (Title, Artist) and finally show these tags. I'd like to get some opinions of this feature, since I've completed this code partly.

Known Issue:
1.Sort These column
2.CMake library including with libid3（I include library with absolute path, this is just a temporary solution.)
3.Acceleration of Reading Operation
4.ID3Tag Edit, Writing using filepropertydialog.

### DIFF
--- a/src/foldermodel.cpp
+++ b/src/foldermodel.cpp
@@ -223,6 +223,18 @@ QVariant FolderModel::data(const QModelIndex & index, int role/* = Qt::DisplayRo
           const char* name = fm_file_info_get_disp_owner(info);
           return QString::fromUtf8(name);
         }
+        case ColumnFileID3Title: {
+          ID3DataModel id3Data (fm_path_to_str(fm_file_info_get_path(info)));
+          return QString::fromStdString(id3Data.parseData(ID3DataModel::ColumnFileID3Title));
+        }
+        case ColumnFileID3Artist: {
+          ID3DataModel id3Data (fm_path_to_str(fm_file_info_get_path(info)));
+          return QString::fromStdString(id3Data.parseData(ID3DataModel::ColumnFileID3Artist));
+        }
+        case ColumnFileID3Album: {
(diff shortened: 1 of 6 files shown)


src/foldermodel.cpp
```diff
@@ -223,6 +223,18 @@ QVariant FolderModel::data(const QModelIndex & index, int role/* = Qt::DisplayRo
           const char* name = fm_file_info_get_disp_owner(info);
           return QString::fromUtf8(name);
         }
+        case ColumnFileID3Title: {
+          ID3DataModel id3Data (fm_path_to_str(fm_file_info_get_path(info)));
```
philippwiesemann: The strings from `fm_path_to_str()` should to be released with `g_free()` later (or it will be a memory leak).

PCMan:
Thanks for the PR. This is a great feature to have, but the implementation has some problems.
ID3DataModel object creation involves blocking disk I/O for each mp3 file.
The implement creates the model objectwhen a column value is needed, which generates unnecessary disk I/O every time.
For a GUI file manager, blocking I/O is only allowed in worker threads. Otherwise the UI will freeze if the amount of I/O increases. A better place to do this might be the async directory listing job of libfm.
For the design problem, I'm not able to merge the PR at the moment.

YWJamesLin:
@PCMan 
Thank for your advice. I admit that there is a  design problem there.  But I have think a lot and I will find other solution to make this happen, because this is maybe a good feature that I'd like to implement. Due to the design problem, I will implement with async and make this feature optional (i.e. as an option in settings.)

PR #32, opened 2016-07-28
Correct positioning of small icons in icon/thumbnail mode

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/377. All credit goes to @rafirafi.

When the icon of an item (e.g. a desktop file) is smaller than the big icon/thumbnail size, its text doesn't align with other item texts. This commit fixes the problem by using the original `QStyleOptionViewItem`, instead of the one initialized by `initStyleOption()`, to get the decoration size.


tsujan:
Could someone please confirm that the bug exists and is fixed by this commit?

Before:
before

After:
after

paulolieuthier:
Fix confirmed. I don't see python's or mutt's icons here, but I could test it with KDE Connect's icon. I'd just add a short comment explaning why we use the original `option` instead of `opt`.

tsujan:
@paulolieuthier Thanks! I'll add a short comment.

luis-pereira:
Bug and fix confirmed. GTM.

PR #35, opened 2016-08-03
Use lxqt compiler settings

luis-pereira:
@PCMan Can you pls take a look at https://github.com/lxde/libfm-qt/commit/eab53132431899bf8a944a4948ab0abf9b655847 ? You know the code better.


tsujan:
@luis-pereira 
Theoretically speaking, https://github.com/lxde/libfm-qt/commit/eab53132431899bf8a944a4948ab0abf9b655847 seems more appropriate to me.

luis-pereira:
@tsujan If FTBFS without https://github.com/lxde/libfm-qt/commit/eab53132431899bf8a944a4948ab0abf9b655847 because of https://github.com/lxde/liblxqt/blob/master/cmake/modules/LXQtCompilerSettings.cmake#L68.

tsujan:
@luis-pereira 
On second thought, maybe `toUtf8()` is needed there because of "copy\n".

luis-pereira:
@tsujan converting "copy\n" to UTF-8 it's a no operation as it's US-ASCII/Latin-1 encoded. The issue might be in `urilist`. What's `urilist` encoding ?

tsujan:
We have `data->setData("text/uri-list", urilist);` below it, so if you're sure that there's no need to convert "copy\n" to UTF-8, the commit couldn't do any harm.

tsujan:
@luis-pereira 
The hierarchy is

`fm_path_list_to_uri_list() → fm_path_list_write_uri_list() → fm_path_to_uri() → fm_path_to_str() and g_filename_to_uri()`

which means it consists of escaped ASCII-encoded URIs, with the components following Section 3.3. of RFC 2396 (https://developer.gnome.org/glib/stable/glib-URI-Functions.html#g-filename-to-uri).

luis-pereira:
@tsujan Thaks for the digging. So, there is no doubt, it's good to go.

tsujan:
> So, there is no doubt, it's good to go.

Yes, it is :)

paulolieuthier:
GTM.

luis-pereira:
@tsujan @paulolieuthier This commit will make a lot of warning popup. Hoping to fix them before the release.

tsujan:
> This commit will make a lot of warning popup. Hoping to fix them before the release.

I'll also check them tomorrow.

luis-pereira:
@tsujan That's great.

PR #36, opened 2016-08-04
Silence new compiler warnings

tsujan:
All new warnings and some old ones are silenced. However, I didn't try to silence some *old* warnings about `filepropsdialog.cpp` because I saw odd assignments in it.

### DIFF
--- a/src/appchoosercombobox.cpp
+++ b/src/appchoosercombobox.cpp
@@ -26,11 +26,11 @@ namespace Fm {
 
 AppChooserComboBox::AppChooserComboBox(QWidget* parent):
   QComboBox(parent),
-  defaultApp_(NULL),
+  mimeType_(NULL),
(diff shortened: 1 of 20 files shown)


paulolieuthier:
GTM.

luis-pereira:
GTM

tsujan:
@paulolieuthier and @luis-pereira

What do you think about conditions like `newUid != -1` or assignments like `newGid = -1;` in `filepropsdialog.cpp`, while newUid and newGid are `guint32`?

luis-pereira:
@tsujan I think that it should be rewritten. That kind of code is screaming: danger.

tsujan:
@luis-pereira I think so too. I'll work on it but, fortunately, it works in its current state ;)

paulolieuthier:
I second that.

PR #38, opened 2016-09-07
Fix Custom Actions Submenu

tsujan:
Fixes https://github.com/lxde/lxqt/issues/1135

A screenshot:

submenu


tsujan:
Please see http://www.nautilus-actions.org/?q=node/377 for a detailed explanation of how to create custom actions in submenus. The custom actions folder is `~/.local/share/file-manager/actions`.

tsujan:
For some reason unknown to me, https://github.com/lxde/lxqt/issues/1135 is deleted. So, I briefly describe the issue that's fixed by this PR:

Custom actions can be grouped on a submenu but only one of them was shown before. The cause was that separate submenus were created per action in the code (a typo?).

paulolieuthier:
GTM.

agaida:
GTM

PR #40, opened 2016-09-24
Remove empty files

jubalh:
All of these header files are empty. I don't see why and where they are
needed.


palinek:
Indeed... but IMO @PCMan should have a look and say if these files are present intentionaly.

tsujan:
GTM.

Perhaps these empty files were created because of an old plan that wasn't followed later.

paulolieuthier:
They were probably auto generated, I believe @PCMan used some tool to ease the porting process.

PCMan:
Yes, it's safe to remove them.
They are automatically generated by my C to C++ wrapper script.
The original plan was to add the wrappers manually to these files, but I'm don't have the time to do it.
Please feel free to delete them. Thanks!

palinek:
landed in @a421dbd, thanks

PR #41, opened 2016-09-27
Also Consider GEmblemedIcon

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/400

GEmblemedIcon wasn't handled by `IconTheme::convertFromGIcon()` (in icontheme.cpp).

### DIFF
--- a/src/icontheme.cpp
+++ b/src/icontheme.cpp
@@ -89,6 +89,8 @@ QIcon IconTheme::iconFromNames(const char* const* names) {
 }
 
 QIcon IconTheme::convertFromGIcon(GIcon* gicon) {
+  if(G_IS_EMBLEMED_ICON(gicon))
+    gicon = g_emblemed_icon_get_icon(G_EMBLEMED_ICON(gicon));
   if(G_IS_THEMED_ICON(gicon)) {
     const gchar * const * names = g_themed_icon_get_names(G_THEMED_ICON(gicon));
     QIcon icon = iconFromNames(names);


tsujan:
In the future, we should add emblem support too.

pmattern:
When lxde/pcmanfm-qt@1b59b89df is compiled against lxde/libfm-qt@ccd0257 that generic "unknown" icon gets replaced by a drive icon. Tested with icon themes Faience, GNOME and Oxygen in an Openbox only session (no desktop environment).

There's one thing left confusing me, though.
The icon is a generic drive icon only, not an icon featuring a padlock. PCManFM (GTK) does display an icon featuring a padlock on the same system but looking through the icon themes I could not find that icon.
So does PCManFM (GTK) use something hard-coded or is there something else missing?

tsujan:
@pmattern 
Thanks for yout thorough test! You didn't need to recompile pcmanfm-qt though.

That padlock is an emblem. As I mentioned above, emblem support should be added later.

pmattern:
> That padlock is an emblem. As I mentioned above, emblem support should be added later.

I see. Hadn't heard of that emblem concept before.

GTM then.

tsujan:
> Hadn't heard of that emblem concept before.

It's better known by GTK users but, for example, the small icon you see on the edge of a symlink is an emblem.

In ancient times, when I used Nautilus, I made the best of emblems with an action script. Optional emblems are very useful for marking files (and not only folders). Now, in the Qt world, I miss them. Maybe, this is a good opportunity to think about adding them to pcmanfm-qt :)

palinek:
Just FYI... here I've used something like the said embedding https://github.com/palinek/nm-tray/blob/master/src/menuview.cpp#L54

tsujan:
@palinek, thank you! I'll surely consider it when I start to add emblem support.
My goal is not only a lock emblem for the encrypted volume icon (which doesn't have any emblem in Dolphin either) but also optional emblems for files.

tsujan:
I experimented with emblems for two hours but the results weren't encouraging:

Unfortunately, it seems that gvfs emblem info is ignored by libfm, so that my dream of optional emblems couldn't come true unless large changes are made to libfm :(

On the other hand, although adding emblems to encrypted volumes is easy in the way @palinek suggested, there's a serious problem that can't be solved with the current structure of the code: items don't have a size hint. In simple words, only *fixed size* emblems are possible, which are senseless because the icon size can be changed by pcmanfm-qt's preferences.

I may attack these problems later again but, for now, this PR seems to be the best solution.

PCMan:
1. Emblemed icons:
@tsujan Showing emblems has nothing to do with libfm. The problem is QIcon::pixmap does not know emblems. To create an icon with emblems, you have to create QIcon for the icon itself, and create one QIcon object for each emblem attached to the icon.
Then, generate pixmaps for all of these QIcons. Next, composite them into a new QPixmap with proper emblem sizes and layouts which are not defined by any spec. This QPixmap then is used to paint the final icon.
The only sane place to do this is the QIcon engine. Otherwise we need to handle all the pixmap generation and cache stuff ourselves bypassing the whole QIcon stack.

2. File emblems:
Not supporting emblem is a design decision made in the early days of LXDE. Supporting per-file emblem will create several additional I/O operations for every single file listed in the directory, which is not we want. The scale of the impact is much more than that of the per-folder setting stuff.
Nautilus does it in a very inefficient way, creating one config file for every file that has an emblem. This is never what we want to do. Thunar does it a little bit better by storing the emblem information using tdb database developed by samba developers so it can be more efficient.
However, maintaining the information can be a problem, just like the per-folder setting issue. How can you delete the emblems when the files no longer exist? There is no reliable and efficient way to do it.
So, there was no plan to support the per-file emblem from the very beginning. However, if someone can figure out an efficient way to tackle the issues, I'm happy to see it implemented.

PCMan:
Supporting file emblems is easy, but supporting it in an efficient way is not.
QIcon is generic and is designed for all platforms and is not so good for use with the xdg specs.
GTK+ does a much better job in this field and I think the GIcon stuff is much more flexible than QIcon.
When handling GIcon, gtk+ tries different fallbacks defined by the programmer to get a usable icon.
https://developer.gnome.org/gio/stable/GThemedIcon.html#g-themed-icon-new-with-default-fallbacks
Every GIcon has several fallback icon names attached to it. After you change the icon theme, the icon name specified previously might no longer be preent in the current theme, so the icon loader will search in all of the fallback icon names again to find another usable icon. In QIcon, there is no fallback icon names, so the icon will simply disappear in this case.
In pcmanfm-qt, I did some dirty hacks by storing the GIcon information in FmIcon object, and try to reload the icons manually on theme changes. So in most cases it still works but that's done manually rather than automatically by QIcon. In gtk+, this work automatically.
We have our own icon engine already, but there is no way to store additional information in the QIcon objects. Even if we can do it, there is no way to pass some fallback icon names or emblem information to the QIcon instance since QIcon::fromTheme() only accept one argument, an icon name.
I haven't found any way to extend it without creating our own new library to replace QIcon.

PCMan:
@luis-pereira BTW I'm aware that XdgIcon provides a version of fromTheme() which accepts a list of icon names. However, this only handles the fallbacks when you create the icon the first time. If the icon theme is changed, the previously chosen icon name can become absent in current theme and the QIcon is broken. To make this useful, there needs to be a way to store the fallback icon names in the QIcon or QIconEngine object.

tsujan:
> Showing emblems has nothing to do with libfm

I meant that for *files*, not for encrypted volume icons.

> The only sane place to do this is the QIcon engine.

Yes, I also thought of it as the last resort but it would be a lot of work for a simple thing.

> However, maintaining the information can be a problem...

It seems so. However, I'm not completely disappointed.

Anyway, this PR simply detects and ignores emblems.

palinek:
> BTW I'm aware that XdgIcon provides a version of fromTheme() which accepts a list of icon names. However, this only handles the fallbacks when you create the icon the first time

This is truth... and you need to handle the theme changes by yourself... I've described the behaviour recently in the lxde/liblxqt#106

tsujan:
I also liked to add this:

It doesn't seem to me that the problems I encountered today are caused by Qt. pcmanfm-qt isn't *rewritten* in Qt; it's structured around GTK but written in Qt. This can be seen as good thing (benefiting from potentials of both worlds) but creates its own problems that, perhaps, should be solved in unorthodox ways.

PCMan:
@tsujan GTM for this PR.
But to solve the lack of fallback icon problems of QIcon, it's better to hack our XdgIcon and make it support multiple fallback icon names and even emblems. In this way everyone using the library can have nice themed icons that really work.

tsujan:
@PCMan
After trying for one more hour, I gave up on my emblem dream ;) I admit it wasn't as easy as I thought and may not worth the amount of required work. So far, so good.

luis-pereira:
@PCMan IMO hacking `XdgIcon` does not solve the fallback problem properly. It can be done for `XdgIcon` but all other stuff using `QIcon` (which are the vast majority) will show a different behavior That's bad IMO. For the average user, consistency is gold.
The main issue is the `QIcon::fromTheme()` interface itself. It accepts only an `QIcon` as a fallback. If the primary icon doesn't exist it uses the fallback. That decision is made by the `QIcon` instance. The IconEngine does not know anything about fallback icons.

tsujan:
I really don’t know why I didn't see the obvious! Maybe I was too tired and made chain of mistakes. Anyway, PCManFM-Qt already supports file emblems just because it's based on GLib. It only needed a simple way to show them:

emblem

I'll make a PR soon (a PR for volume icon emblems is already made).

PR #43, opened 2016-09-30
Show File Emblems

tsujan:
Being based on GLib, PCManFM-Qt can show `GFile` emblems as Nautilus/Nemo/Thunar can. If an emblem `ICON_NAME` is added to a file/folder `FILE` with:
````
gvfs-set-attribute FILE -t stringv metadata::emblems ICON_NAME
````
PCManFM-Qt will show it with this commit. The emblem can be removed with:
````
gvfs-set-attribute FILE -t unset metadata::emblems
````

For simplicity's sake, if a file has more than one emblem, only the first emblem will be shown.

Emblems can be set/removed easily with a file-manager action.

### DIFF
--- a/src/folderitemdelegate.cpp
+++ b/src/folderitemdelegate.cpp
@@ -79,7 +79,23 @@ QIcon::Mode FolderItemDelegate::iconModeFromState(const QStyle::State state) {
 void FolderItemDelegate::paint(QPainter* painter, const QStyleOptionViewItem& option, const QModelIndex& index) const {
   Q_ASSERT(index.isValid());
   FmFileInfo* file = static_cast<FmFileInfo*>(index.data(FolderModel::FileInfoRole).value<void*>());
-  bool isSymlink = file && fm_file_info_is_symlink(file);
+  bool isSymlink(false);
+  QIcon emblem;
+  if(file) {
+    isSymlink = fm_file_info_is_symlink(file);
+    FmPath* path = fm_file_info_get_path(file);
+    GFile* gfile = fm_path_to_gfile(path);
+    GFileInfo *info = g_file_query_info(gfile, "metadata::emblems", G_FILE_QUERY_INFO_NONE, NULL, NULL);
+    char *att = g_file_info_get_attribute_as_string(info, "metadata::emblems");
+    QString emblemName(att);
+    g_free(att);
+    g_object_unref(info);
+    g_object_unref(gfile);
+    if(!emblemName.isEmpty()) { // only get the first emblem
+      emblemName = emblemName.section(QRegExp("\\[|\\]|,"), 0, 0, QString::SectionSkipEmpty);
+      emblem = QIcon::fromTheme(emblemName);
+    }
+  }
 
   if(option.decorationPosition == QStyleOptionViewItem::Top ||
     option.decorationPosition == QStyleOptionViewItem::Bottom) {
@@ -100,9 +116,12 @@ void FolderItemDelegate::paint(QPainter* painter, const QStyleOptionViewItem& op
     painter->drawPixmap(iconPos + QPoint(margin.width(), margin.height()), pixmap);
 
     // draw some emblems for the item if needed
-    // we only support symlink emblem at the moment
     if(isSymlink)
       painter->drawPixmap(iconPos, symlinkIcon_.pixmap(option.decorationSize / 2, iconMode));
+    if(!emblem.isNull()) {
+      QPoint emblemPos(opt.rect.x() + opt.rect.width() / 2, opt.rect.y() + option.decorationSize.height() / 2);
+      painter->drawPixmap(emblemPos, emblem.pixmap(option.decorationSize / 2, iconMode));
+    }
 
     // draw the text
     // The text rect dimensions should be exactly as they were in sizeHint()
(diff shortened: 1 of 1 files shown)


PCMan:
@tsujan At first glance this code works, but I could not give a GTM.
Please note that g_file_query_info is a blocking call involved with disk I/O.
If you are on a remote filesystem, blocking for seconds is possible.
That means, your current implementation could block for seconds for painting "each icon", which is totally unacceptable. This can freeze the UI on slow devices and network filesystems.

The reason why the GIcon object you got from FmIcon does not contain emblems is not so obvious. I did it deliberately. The GIcon does not comes from GFileInfo. In libfm, I did optimizations for local filesystems. So all FmFileInfo of the local files are generated by my hand-coded faster code instead of the inefficient GFileInfo stuff. You can do some simple benchmarks. Getting file information with GFileInfo is quite slow. That partly explains the slowness of Nautilus. So in libfm I totally bypassed this path for local files that's why pcmanfm is faster than nautilus. Calling g_file_info actually defeats the purpose of this hard work.

The emblems of Thunar are not handled by gio if I remembered correctly. They use their own tdb to handle their own emblems info. Not sure if that's changed as I haven't read the code of Thunar in recent two years.

PCMan:
However, you can still support painting the emblemed GIcons as we discussed in this PR.
https://github.com/lxde/libfm-qt/pull/42
If the code is encapsulated in Fm::Icon and the item delegates, once we find a proper way to add emblems to FmFileInfo, everything should work automatically without any code change.
This is why encapsulation matters a lot. :)

tsujan:
@PCMan 
Please see https://github.com/lxde/libfm-qt/pull/42#issuecomment-250683139

PCMan:
@tsujan About this one, sorry it's my fault.Please accept my apology.
Just check the latest source code of Nautilus, gvfs, and Thunar and my info is out dated.
I studied their code years ago, but much has been changed since then.
Nautilus no longer creates scattered xml files for metadata and now gio use a more efficient binary format. Thunar already removed samba tdb and now also uses gio directly to replace their own metadata.
OK, I'll merge this one after #42, but maybe we can support this directly in libfm.
I'll handle this PR on the weekend.

tsujan:
> but maybe we can support this directly in libfm.

That'll be a better choice if you do it. My libfm PR's will be ignored.

PCMan:
@tsujan This piece of code is still useful. We just do the metadata query inside libfm, so the GIcon you get can contain emblems and you don't need to query it again here.
The main difference is, the query done in libfm runs in its own thread, so there is no UI blocking and the result is cached. Your pixmap painting code is still required, but this is also required by the volume icon as well, so it's better to move it to its own method to avoid code duplication.
That's what I mean by putting the code in the best place. In general your contribution is great, but I want to minimize code duplication and API change, and make sure it won't slow down anything.
So, thank you for your understanding and the time spent on this.

PCMan:
Merged and did some refactor.

PCMan:
File emblems information is read and stored in libfm rather than queried during painting every icon.

PCMan:
FYI: https://github.com/lxde/libfm/commit/bfbd8820fd8d45dd744bde87a821514a4e46d64f
With this commit, the gtk+ version of pcmanfm gets emblems automatically.
Both worlds are happy now.

tsujan:
@PCMan 
I tested the new libfm-qt. Since yesterday, I've emblemized some folders and files for testing. I'm afraid I see no file emblem with your commits. When I go back to my commit, emblems reappear.

tsujan:
My bad! I compiled the wrong source. Now all emblems are shown correctly. There's just one thing:

If a file has more than one emblem, its last emblem is shown. In my commit, I chose the first one. Is that intentional?

PCMan:
@tsujan No it's not. The new Fm::IconTheme API provides access to "all emblems", but the item delegate needs to be modified to render multiples emblems for a given file. That's the next step we have to do. For now to make the current code compile, I just take the first emblem from the list without other checks. The order is the original order provided by GFileInfo.

PCMan:
For icons with less than four emblems, it's OK to put them in the corners of the icon.
For symlinks, only three slots are left for showing other emblems.
If there are more emblems, it might not be helpful to show them all since they will overlap and the original main icon will become nearly invisible.
Anyway, with the current code base, the only place needs to be changed is the Fm::FolderItemDelegate::paint() and all emblemed icons should work.

PCMan:
@tsujan Please let me know if you're interested in working on the item delegate painting code.
If the answer is yes, I'll leave it to you and won't touch that part in the near future.
Thanks!

tsujan:
@PCMan
IMO, it's OK to show only the last emblem, in the way it's shown now, and It's better to reserve the other two places (top right and bottom left) for the future -- for example, to emblemize read-only files in the code.

As for the item delegate painting code, I prefer cooperation rather than exclusion :)

P.S. As I've told on various occasions, my only goal is the improvement of pcmanfm-qt. It's now one of the best and could become even better. Our intense differences are OK to me and, as it can be seen from our latest arguments, good things could come out of them. IMO, you have absolute right to reject, change or accept my PR's. When I say it's *your* code, I mean it. I wouldn't like anybody to change *my* programs freely. So, let's have our differences and yet cooperate!

PR #45, opened 2016-10-02
Check for FolderModelItem info (and FmPath)

tsujan:
This is a shot in the dark to prevent rare crashes (https://github.com/lxde/pcmanfm-qt/issues/397).

### DIFF
--- a/src/foldermodel.cpp
+++ b/src/foldermodel.cpp
@@ -366,11 +366,15 @@ QMimeData* FolderModel::mimeData(const QModelIndexList& indexes) const {
   for(const auto &index : indexes) {
     FolderModelItem* item = itemFromIndex(index);
     if(item) {
-      FmPath* path = fm_file_info_get_path(item->info);
-      char* uri = fm_path_to_uri(path);
-      urilist.append(uri);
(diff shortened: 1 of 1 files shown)


src/foldermodel.cpp
```diff
@@ -366,11 +366,15 @@ QMimeData* FolderModel::mimeData(const QModelIndexList& indexes) const {
   for(const auto &index : indexes) {
     FolderModelItem* item = itemFromIndex(index);
     if(item) {
-      FmPath* path = fm_file_info_get_path(item->info);
-      char* uri = fm_path_to_uri(path);
-      urilist.append(uri);
-      urilist.append('\n');
-      g_free(uri);
+      if(item->info) {
```
palinek: How about condensing the conditional `if (item) && item->info)` ?

tsujan: Yes, I know. I just wanted to emphasize that `item->info` may be null theoretically (I still don't know if that's possible in practice),

PCMan:
Merged into b577385bdc6143635bd36b9db92eae8e12a751a1

PR #47, opened 2016-10-08
Restore symlink emblem

tsujan:
The symlink emblem was gone (because of a typo) after implementation of GLib emblems in libfm.

Strange that none of us noticed that!


PCMan:
Merged. Thank you.

PR #49, opened 2016-11-10
CMake: Drop the use of link_directories()

luis-pereira:
pkconfig does the work. This shouldn't be needed and I see no reason for
using it.


luis-pereira:
@palinek @agaida @pmattern
Can you guys check this ? It works on Arch and I see no reason to use `link_directories()`.

agaida:
@luis-pereira - looks good, will test it today

palinek:
Testing with debian unstable:
- when using packaged `libfm-dev`, everything works
- with `libfm` build from sources and installed into non-default location, it fails:
```
[  2%] Linking CXX shared library libfm-qt.so
/usr/bin/ld: cannot find -lfm
clang: error: linker command failed with exit code 1 (use -v to see invocation)
src/CMakeFiles/fm-qt.dir/build.make:1352: recipe for target 'src/libfm-qt.so.3.0.0' failed
make[2]: *** [src/libfm-qt.so.3.0.0] Error 1
CMakeFiles/Makefile2:86: recipe for target 'src/CMakeFiles/fm-qt.dir/all' failed
make[1]: *** [src/CMakeFiles/fm-qt.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2
```

The problem here is the usage of \*_LIBRARIES in `target_link_libraries` of the `pkg-config`'s modules. We need to use the \*_LDFLAGS (acording to [doc](https://cmake.org/cmake/help/v3.0/module/FindPkgConfig.html)), that means following change:
```diff
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 0c45f98..44e4c61 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -100,9 +100,9 @@ install(EXPORT
 target_link_libraries(${LIBFM_QT_LIBRARY_NAME}
     Qt5::Widgets
     Qt5::X11Extras
-    ${LIBFM_LIBRARIES}
-    ${LIBMENUCACHE_LIBRARIES}
-    ${SYSTEM_LIBS_LIBRARIES}
+    ${LIBFM_LDFLAGS}
+    ${LIBMENUCACHE_LDFLAGS}
+    ${SYSTEM_LIBS_LDFLAGS}
     ${XCB_LIBRARIES}
 )
 ```

With this change the scenario 2. builds here successfully.

**Comment by luis-pereira** [comment]
@palinek. With your patch it doesn't export the `INCLUDE_DIRECTORIES` and building pcmamfm-qt will fail.See [here](https://gist.github.com/luis-pereira/2b4ca3c1bc2d0aa39dd2b318e765dddb). You probably don't see it, yet we (actually me) are not exporting it right on the current libfm-qt master.

Can you run https://github.com/lxde/libfm-qt/pull/49/commits/38646d8d7d65d07611a48d738c5d20c816f3eb74 and post the added debug stuff ? And also a verbose log of the linking error ?

Mine is
```
-- LIBFM_LIBRARY_DIRS: 
-- LIBMENUCACHE_LIBRARY_DIRS: 
-- SYSTEM_LIBS_LIBRARY_DIRS: 
-- LIBFM_LDFLAGS: -lfm;-lgthread-2.0;-pthread;-lgio-2.0;-lgobject-2.0;-lglib-2.0
-- LIBMENUCACHE_LDFLAGS: -lmenu-cache;-lglib-2.0
-- SYSTEM_LIBS_LDFLAGS: -lgio-2.0;-lgobject-2.0;-lglib-2.0
-- LIBFM_LIBRARIES: fm;gthread-2.0;gio-2.0;gobject-2.0;glib-2.0
-- LIBMENUCACHE_LIBRARIES: menu-cache;glib-2.0
-- SYSTEM_LIBS_LIBRARIES: gio-2.0;gobject-2.0;glib-2.0
```

**Comment by palinek** [comment]
Here's the debug output:
```
-- LIBFM_LIBRARY_DIRS: /home/palco/opt/lxqt/lib
-- LIBMENUCACHE_LIBRARY_DIRS: 
-- SYSTEM_LIBS_LIBRARY_DIRS: 
-- LIBFM_LDFLAGS: -L/home/palco/opt/lxqt/lib;-lfm;-lgthread-2.0;-pthread;-lgio-2.0;-lgobject-2.0;-lglib-2.0
-- LIBMENUCACHE_LDFLAGS: -lmenu-cache;-lglib-2.0
-- SYSTEM_LIBS_LDFLAGS: -lgio-2.0;-lgobject-2.0;-lglib-2.0
-- LIBFM_LIBRARIES: fm;gthread-2.0;gio-2.0;gobject-2.0;glib-2.0
-- LIBMENUCACHE_LIBRARIES: menu-cache;glib-2.0
-- SYSTEM_LIBS_LIBRARIES: gio-2.0;gobject-2.0;glib-2.0
-- Building libfm-qt with Qt 5.7.1
```
..and the failing output:
```
[  2%] Linking CXX shared library libfm-qt.so
cd /home/palco/work/oss/lxqt/libfm-qt/build/src && /usr/bin/cmake -E cmake_link_script CMakeFiles/fm-qt.dir/link.txt --verbose=1
/usr/bin/c++  -fPIC  -fno-exceptions -Wall -std=c++11 -O3 -DNDEBUG -Wl,--no-undefined  -shared -Wl,-soname,libfm-qt.so.3 -o libfm-qt.so.3.0.0 CMakeFiles/fm-qt.dir/libfmqt.cpp.o CMakeFiles/fm-qt.dir/bookmarkaction.cpp.o CMakeFiles/fm-qt.dir/sidepane.cpp.o CMakeFiles/fm-qt.dir/icontheme.cpp.o CMakeFiles/fm-qt.dir/filelauncher.cpp.o CMakeFiles/fm-qt.dir/foldermodel.cpp.o CMakeFiles/fm-qt.dir/foldermodelitem.cpp.o CMakeFiles/fm-qt.dir/cachedfoldermodel.cpp.o CMakeFiles/fm-qt.dir/proxyfoldermodel.cpp.o CMakeFiles/fm-qt.dir/folderview.cpp.o CMakeFiles/fm-qt.dir/folderitemdelegate.cpp.o CMakeFiles/fm-qt.dir/createnewmenu.cpp.o CMakeFiles/fm-qt.dir/filemenu.cpp.o CMakeFiles/fm-qt.dir/foldermenu.cpp.o CMakeFiles/fm-qt.dir/filepropsdialog.cpp.o CMakeFiles/fm-qt.dir/applaunchcontext.cpp.o CMakeFiles/fm-qt.dir/placesview.cpp.o CMakeFiles/fm-qt.dir/placesmodel.cpp.o CMakeFiles/fm-qt.dir/placesmodelitem.cpp.o CMakeFiles/fm-qt.dir/dirtreeview.cpp.o CMakeFiles/fm-qt.dir/dirtreemodel.cpp.o CMakeFiles/fm-qt.dir/dirtreemodelitem.cpp.o CMakeFiles/fm-qt.dir/dnddest.cpp.o CMakeFiles/fm-qt.dir/mountoperation.cpp.o CMakeFiles/fm-qt.dir/mountoperationpassworddialog.cpp.o CMakeFiles/fm-qt.dir/mountoperationquestiondialog.cpp.o CMakeFiles/fm-qt.dir/fileoperation.cpp.o CMakeFiles/fm-qt.dir/fileoperationdialog.cpp.o CMakeFiles/fm-qt.dir/renamedialog.cpp.o CMakeFiles/fm-qt.dir/pathedit.cpp.o CMakeFiles/fm-qt.dir/colorbutton.cpp.o CMakeFiles/fm-qt.dir/fontbutton.cpp.o CMakeFiles/fm-qt.dir/browsehistory.cpp.o CMakeFiles/fm-qt.dir/utilities.cpp.o CMakeFiles/fm-qt.dir/dndactionmenu.cpp.o CMakeFiles/fm-qt.dir/editbookmarksdialog.cpp.o CMakeFiles/fm-qt.dir/thumbnailloader.cpp.o CMakeFiles/fm-qt.dir/execfiledialog.cpp.o CMakeFiles/fm-qt.dir/appchoosercombobox.cpp.o CMakeFiles/fm-qt.dir/appmenuview.cpp.o CMakeFiles/fm-qt.dir/appchooserdialog.cpp.o CMakeFiles/fm-qt.dir/filesearchdialog.cpp.o CMakeFiles/fm-qt.dir/fm-search.c.o CMakeFiles/fm-qt.dir/xdndworkaround.cpp.o CMakeFiles/fm-qt.dir/fm-qt_automoc.cpp.o /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5.7.1 /usr/lib/x86_64-linux-gnu/libQt5X11Extras.so.5.7.1 -lfm -lgthread-2.0 -lgio-2.0 -lgobject-2.0 -lglib-2.0 -lmenu-cache -lglib-2.0 -lgio-2.0 -lgobject-2.0 -lglib-2.0 -lxcb /usr/lib/x86_64-linux-gnu/libQt5Gui.so.5.7.1 /usr/lib/x86_64-linux-gnu/libQt5Core.so.5.7.1 -lmenu-cache -lxcb 
/usr/bin/ld: cannot find -lfm
clang: error: linker command failed with exit code 1 (use -v to see invocation)
src/CMakeFiles/fm-qt.dir/build.make:1352: recipe for target 'src/libfm-qt.so.3.0.0' failed
make[2]: *** [src/libfm-qt.so.3.0.0] Error 1
```

PCMan:
According to the document of cmake pkg_check_modules().

It sets the following variables.

    PKG_CONFIG_FOUND          ... if pkg-config executable was found
    PKG_CONFIG_EXECUTABLE     ... pathname of the pkg-config program
    PKG_CONFIG_VERSION_STRING ... the version of the pkg-config program found
                                  (since CMake 2.8.8)

 For the following variables two sets of values exist; first one is the
 common one and has the given PREFIX.  The second set contains flags
 which are given out when ``pkg-config`` was called with the ``--static``
 option. ::

    <XPREFIX>_FOUND          ... set to 1 if module(s) exist
    <XPREFIX>_LIBRARIES      ... only the libraries (w/o the '-l')
    <XPREFIX>_LIBRARY_DIRS   ... the paths of the libraries (w/o the '-L')
    <XPREFIX>_LDFLAGS        ... all required linker flags
    <XPREFIX>_LDFLAGS_OTHER  ... all other linker flags
    <XPREFIX>_INCLUDE_DIRS   ... the '-I' preprocessor flags (w/o the '-I')
    <XPREFIX>_CFLAGS         ... all required cflags
    <XPREFIX>_CFLAGS_OTHER   ... the other compiler flags

So, _LIBRARY_DIRS contain the path of the libraries, which is not always the default path. That's why I put it in link_directories(). By default most libraries should built without it since they are in the default location, but to be safe it's better to leave them there IMHO.

luis-pereira:
@palinek Thanks for the debug info. I hadn't considered custom locations.
@PCMan I'm aware of that. But IMO if one has to use `link_directories()` it's on the wrong path.

I've written Find modules for Fm and MenuCache. Check it [here](https://github.com/lxde/lxqt-build-tools/pull/8).
The accompaning libfm-qt PR it's [here](https://github.com/lxde/libfm-qt/pull/50).

Handle GLib stuff: in progress.

PR #50, opened 2016-11-22
Use cmake modules drop link directories

luis-pereira:


PCMan:
tested and it works.
LGTM after https://github.com/lxde/lxqt-build-tools/pull/8 is merged.

agaida:
GTM with lxde/lxqt-build-tools#8

PR #51, opened 2016-12-03
Align Path Buttons

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/414.

The toolbar icon size is also respected.

### DIFF
--- a/src/pathbar_p.h
+++ b/src/pathbar_p.h
@@ -21,7 +21,6 @@
 #define FM_PATHBAR_P_H
 
 #include <QToolButton>
-#include <QApplication>
 #include <QStyle>
 #include "path.h"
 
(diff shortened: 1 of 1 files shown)


tsujan:
A screenshot with Fusion:

pathbuttons

src/pathbar_p.h
```diff
@@ -32,10 +34,13 @@ class PathButton: public QToolButton {
     QToolButton(parent),
     pathElement_(pathElement) {
 
-      setSizePolicy(QSizePolicy::Fixed, QSizePolicy::Fixed);
+      setSizePolicy(QSizePolicy::Fixed, QSizePolicy::MinimumExpanding);
       setCheckable(true);
       setAutoExclusive(true);
       setToolButtonStyle(Qt::ToolButtonTextBesideIcon);
+      /* respect the toolbar icon size (can be set with some styles) */
+      int icnSize = qApp->style()->pixelMetric(QStyle::PM_ToolBarIconSize);
```
palinek: Just a note: by the use of `QApplication::style()` you're ignoring the `QWidget::style()` (if overriden...) and also, if style is changed (or some stylesheet used..) the toolbutton widget does not adapt.

tsujan: You're right. Thanks! I'll see what's best.

tsujan:
@palinek 
It made no difference. After changing the syle, the size won't be updated until the folder is changed. But I'll replace `qApp->style()` with `style()` for now.

tsujan:
@palinek, I think `PathBar` needs additional code to be completely style aware but the current situation is not a big deal.

PCMan:
Merged into master. I also made it aware of style changes.
Thanks.

PR #52, opened 2016-12-06
Scroll Pathbar with Mouse Wheel

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/419.

Also arrow buttons are fixed for RTL and their enabled state is set correctly. In addition, they are made hidden on editing pathbar.

The one-step scrolling by clicking on the arrow buttons is kept because wheel scrolling is possible now.

### DIFF
--- a/src/pathbar.cpp
+++ b/src/pathbar.cpp
@@ -239,8 +239,10 @@ void PathBar::closeEditor() {
     return;
   layout()->replaceWidget(tempPathEdit_, scrollArea_, Qt::FindDirectChildrenOnly);
   scrollArea_->show();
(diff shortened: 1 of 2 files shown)


PCMan:
Merged into c789bec658d82684a67153a2d087b2ab117ef57b.
I also did some variable renaming to make the code more readable.
leftArraw->setArrowType(Right) does not look right. So I rename the buttons.

tsujan:
> I also did some variable renaming to make the code more readable.

Yes, that's better! I also wanted to change them but supposed that the code reader would take LTR as default.

PR #54, opened 2016-12-10
Fix Pathbar Paint on Menu Pop-Up

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/422 by hiding path-edit before deleting it.


PCMan:
Merged.

PR #55, opened 2016-12-16
Fixes QString to char conversion dangling pointers

luis-pereira:
Explanation at https://github.com/lxde/pcmanfm-qt/issues/427.
Closes https://github.com/lxde/pcmanfm-qt/issues/427.

### DIFF
--- a/src/utilities.cpp
+++ b/src/utilities.cpp
@@ -136,7 +136,8 @@ void renameFile(FmFileInfo *file, QWidget *parent) {
 
   GFile* gf = fm_path_to_gfile(path);
   GFile* parent_gf = g_file_get_parent(gf);
-  GFile* dest = g_file_get_child(G_FILE(parent_gf), new_name.toLocal8Bit().data());
+  QByteArray nm = new_name.toLocal8Bit();
+  GFile* dest = g_file_get_child(G_FILE(parent_gf), nm.constData());
   g_object_unref(parent_gf);
 
   GError* err = NULL;
@@ -193,7 +194,8 @@ _retry:
     return;
 
   GFile* parent_gf = fm_path_to_gfile(parentDir);
-  GFile* dest_gf = g_file_get_child(G_FILE(parent_gf), new_name.toLocal8Bit().data());
+  QByteArray nm = new_name.toLocal8Bit();
+  GFile* dest_gf = g_file_get_child(G_FILE(parent_gf), nm.constData());
   g_object_unref(parent_gf);
 
   GError* err = NULL;


tsujan:
GTM.

src/utilities.cpp
```diff
@@ -136,7 +136,8 @@ void renameFile(FmFileInfo *file, QWidget *parent) {
 
   GFile* gf = fm_path_to_gfile(path);
   GFile* parent_gf = g_file_get_parent(gf);
-  GFile* dest = g_file_get_child(G_FILE(parent_gf), new_name.toLocal8Bit().data());
```
palinek: What's the problem here? Doesn't the `g_file_get_child` make a copy of the received string if it needs/stores it in the `GFile`? If so, then there is no problem with the original call as the string/pointer is valid throughout the whole call of the `g_file_get_child`..

tsujan: @palinek, I also think there was no problem in the first place at all. However, I think @luis-pereira's PR is *logically* good because `g_file_get_child()` gets a const char*.

agaida:
GTM

tsujan:
@luis-pereira, `QByteArray nm` isn't needed. IMHO, just replace `data()` with `nm.constData()` and merge it!

luis-pereira:
@tsujan @palinek I did a quick check and ended up with the impression that `g_file_get_child` did not perform a copy. That was the point. If that's not the case, there's no point.

Can somebody do it ? I'm away!

tsujan:
@luis-pereira 

I don't get why `g_file_get_child` should copy anything. We give a `const char*` to a GLib function; there's nothing wrong with the code. `constData()` is just more standard.

tsujan:
Qt doc says, "For read-only access, constData() is faster because it never causes a deep copy to occur," and it explicitly says "This function is mostly useful to pass a byte array to a function that accepts a const char *."

luis-pereira:
@tsujan 
```GFile *g_file_get_child (GFile *file, const char *name);```
At some point it will use [g_build_filename()](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Utility-Functions.html#g-build-filename) and the contents are copied. This way name can go out of scope without any issue. In my quick check I failed to recognize that. That's the motivation for the patch.

In the doc the sentence before that one is:

> The pointer remains valid as long as the byte array isn't reallocated or destroyed.

No doubt that the `constData()` method should be used.

tsujan:
> In the doc the sentence before that one is:
> The pointer remains valid as long as the byte array isn't reallocated or destroyed.
> No doubt that the constData() method should be used.

Yes, the pointer might be invalid at some point but that doesn't concern use as we get our data (char*) *immediately*. `QByteArray nm` doesn't do anything here. If you've added it for more readability, I'm not against it. `constData()` is surely better -- and that was why I gave my "GTM" -- but not necessary.

tsujan:
Please also see that the whole https://github.com/lxde/pcmanfm-qt/issues/427 was based on a conceptual mistake. However, it lead to `constData()` quite indirectly, which is good.

luis-pereira:
@tsujan There is no doubt that `nm` is not needed. 
It was not for readability.
It doesn't matter if the data (char *) is taken immediately. What matter's is that data is copied while still valid. That's why `nm` is not needed.

tsujan:
@luis-pereira 
If going out of scope was a problem (and I don't know how it could be), `QByteArray nm = new_name.toLocal8Bit();` wouldn't make any difference.

Anyhow, apart from theoretical differences, we both agree that it's good to have `constData()` instead of `data()`.

luis-pereira:
@tsujan `nm` scope is different from the scope of an temporary object. Or at least __can__ be.

tsujan:
>  nm scope is different from the scope of an temporary object.

Yes, but the temporary object scope is good for `g_file_get_child()`. I don't see any reason to worry about a dangling pointer. This started only with a misconception at lxde/pcmanfm-qt#427. The good thing was `constData()`.

agaida:
@tsujan and @luis-pereira - feel free to descide and merge the PR once you are finished your discussion :)

tsujan:
@agaida, the PR is good, especially without `nm` (which I hope @luis-pereira will remove). Our discussion was only theoretical.

luis-pereira:
@tsujan I didn't do it, yet, because I'm away from my dev machine.

luis-pereira:
@tsujan Yes, the temporary scope is good for `g_file_get_child()`. That was my mistake.

luis-pereira:
New PR: https://github.com/lxde/libfm-qt/pull/57

PR #57, opened 2016-12-17
Use QByteArray::constData() where we can

luis-pereira:
Faster than data() because it never causes a deep copy to occur.

### DIFF
--- a/src/utilities.cpp
+++ b/src/utilities.cpp
@@ -136,7 +136,7 @@ void renameFile(FmFileInfo *file, QWidget *parent) {
 
   GFile* gf = fm_path_to_gfile(path);
   GFile* parent_gf = g_file_get_parent(gf);
-  GFile* dest = g_file_get_child(G_FILE(parent_gf), new_name.toLocal8Bit().data());
+  GFile* dest = g_file_get_child(G_FILE(parent_gf), new_name.toLocal8Bit().constData());
   g_object_unref(parent_gf);
 
   GError* err = NULL;
@@ -193,7 +193,7 @@ _retry:
     return;
 
   GFile* parent_gf = fm_path_to_gfile(parentDir);
-  GFile* dest_gf = g_file_get_child(G_FILE(parent_gf), new_name.toLocal8Bit().data());
+  GFile* dest_gf = g_file_get_child(G_FILE(parent_gf), new_name.toLocal8Bit().constData());
   g_object_unref(parent_gf);
 
   GError* err = NULL;
--- a/src/filepropsdialog.cpp
+++ b/src/filepropsdialog.cpp
@@ -439,7 +439,7 @@ void FilePropsDialog::accept() {
       FmPath* path = fm_file_info_get_path(fileInfo);
       GFile* gf = fm_path_to_gfile(path);
       GFile* parent_gf = g_file_get_parent(gf);
-      GFile* dest = g_file_get_child(G_FILE(parent_gf), new_name.toLocal8Bit().data());
+      GFile* dest = g_file_get_child(G_FILE(parent_gf), new_name.toLocal8Bit().constData());
       g_object_unref(parent_gf);
       GError* err = NULL;
       if(!g_file_move(gf, dest,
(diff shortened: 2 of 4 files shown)


tsujan:
GTM.

tsujan:
Oh, my bad! There is a problem:
````
error: invalid conversion from ‘const char*’ to ‘char*’ [-fpermissive]
     char* eol = strchr(pdata, '\n');
````
@luis-pereira, please revert to `char* pdata = gnomeData.data();` and it'll be OK.

File: src/utilities.cpp [hunk 1]
```diff
@@ -59,7 +59,7 @@ void pasteFilesFromClipboard(FmPath* destPath, QWidget* parent) {
   if(data->hasFormat("x-special/gnome-copied-files")) {
     // Gnome, LXDE, and XFCE
     QByteArray gnomeData = data->data("x-special/gnome-copied-files");
-    char* pdata = gnomeData.data();
+    const char* pdata = gnomeData.constData();
```
tsujan: This needs to be reversed because of `strchr()`.

luis-pereira: @tsujan I was able to build it with clang. Did you use gcc ?
The prototype is `char *strchr(const char *s, int c);` anyway. Will check it later.

tsujan: Yes, gcc. See [here](http://www.cplusplus.com/reference/cstring/strchr/). It's logical that either both of them or none of them is const.

tsujan: I forgot to tell this: `man strchr` says `char *strchr(const char *s, int c);` and I only saw the problem when I tried to compile.

tsujan: Also see http://en.cppreference.com/w/cpp/string/byte/strchr

agaida:
how is the current state?

tsujan:
Ok, this is what I've found:

Clang follows `man strchr` and considers the prototype as `char *strchr(const char *s, int c);`. Hence a successful compilation.

However, GCC considers these prototypes:
````
const char *strchr(const char *s, int c);
char *strchr(char *s, int c);
````
They seem more standard to me. So compilation fails.

Therefore. it seems that `const char* pdata = gnomeData.constData();` should be reversed to char* pdata = gnomeData.data();``.

tsujan:
BTW, `constData()` could be kept by using `strdup()` as a workaround but, personally, I don't like it. IMO, the original code is better.

agaida:
if would be nice if this could be finished finished soon - i want to start the second batch of point release (including libfmqt, lximage-qt and pcmanfm-qt) and would like to have this in.

tsujan:
@agaida 
I think @luis-pereira will fix this tomorrow. But don't you want to wait for lximage-qt's DND and pcmafm-qt's slide-show?

luis-pereira:
@tsujan @agaida I already pushed a correction.

agaida:
@tsujan - i want both in - as soon as possible if they work and don't wreck things. And i want to release the current things soon, so if any problems occour we could react in a timely manner before Debian freezes. If we miss some things now, we have the chance to fix it approx. until Jan. 20 - after this date it will be hard to get any changes into Debian/Ubuntu for their next releases. And i want our software in a shape that makes our bugtracker don't explode once Stretch or 17.04 is released :D

luis-pereira:
@agaida  this one is ready. It's tested with both gcc and clang.
You have to update your copy. I did a `git push -f`.

@tsujan It seems that somehow clang is considering the prototypes defined at stdlib and gcc the ones at cstdlib. gcc is right, this is a c++ project. I will take a look to the source of that clang behavior.

I have a large embedded C programming. The C prototypes are "always in my mind".

tsujan:
@luis-pereira, thanks for the explanation!

This is ready indeed! So, I merge it.

>  i want both in - as soon as possible if they work and don't wreck things. 

@agaida, both work. I'm using them and you've tried them too. I don't see any problem in the code but this is a known fact that fresh eyes may see all sorts of things ;) So, I prefer to wait for another GTM in both cases.

tsujan:
@agaida 
BTW, this may be a stupid question but I remember you had a bleeding edge LXQt Debian repository of your own. What was its address? I don't want to compile everything when I restart to my Debian system once a month.

agaida:
https;/packages.4sid.eu

tsujan:
Thanks :)

PR #59, opened 2016-12-19
Fix erasing of filter string with path buttons

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/431.


tsujan:
Sorry, closed the correct PR. This should be closed and replaced by https://github.com/lxde/pcmanfm-qt/pull/432

PR #60, opened 2017-01-09
Arrange Custom Actions

tsujan:
The patch arranges custom actions according to the standard explained at http://www.nautilus-actions.org/?q=node/377.

The items that are included in `~/.local/share/file-manager/actions/level-zero.directory` come first and in the order specified in that file. Other items follow them in the alphabetical order.

Fixes "Issue 2" from https://github.com/lxde/pcmanfm-qt/issues/438.

P.S. GKeyFile is used intentionally instead of QSettings because key values may contain semicolon. A purely Qt solution was possible but this seemed cleaner -- after all, it's about a GLib function.

### DIFF
--- a/src/filemenu.cpp
+++ b/src/filemenu.cpp
@@ -29,12 +29,33 @@
 #ifdef CUSTOM_ACTIONS
 #include <libfm/fm-actions.h>
 #include "customaction_p.h"
-#endif
 #include <QMessageBox>
+#endif
 #include <QDebug>
 #include "filemenu_p.h"
 
 namespace Fm {
+
+static QStringList itemNameList_;
+
+static int compare_items(FmFileActionItem* a, FmFileActionItem* b)
+{
+  const gchar* a_name = fm_file_action_item_get_name(FM_FILE_ACTION_ITEM(a));
+  const gchar* b_name = fm_file_action_item_get_name(FM_FILE_ACTION_ITEM(b));
+  if(!itemNameList_.isEmpty()) {
+    int first = itemNameList_.indexOf(QString::fromUtf8(a_name));
+    int second = itemNameList_.indexOf(QString::fromUtf8(b_name));
+    if(first > -1) {
+      if(second > -1)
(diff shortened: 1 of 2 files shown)


pmattern:
Basically lxde/pcmanfm-qt@1d38bbd compiled against 53da05c behaves as expected.
In absence of file `level-zero.directory` items are sorted in alphabetical order. The order is referring to the values of key `Name`, not to the filenames which I think is the right thing to do. When `level-zero.directory` is available, the topics stated by its key `ItemsList` are displayed first and followed by the others in alphabetical order.

There seems to be an issue with localized keys, though.
When key `Name` is localized like in `Name[de]`, there's still the same correct alphabetical order as long as `level-zero.directory` is absent. But `level-zero.directory` just seems to be ignored, when it's present an impact on the order cannot be seen.

Seen with `LANG=de_DE.UTF-8` and trivial action files `{a,b,c,d}.desktop` stating letters in reverse alphabetical order like
```
$ cat a.desktop
[Desktop Entry]
Type=Action
Name=D
Name[de]=Z
Description=D description
Tooltip=D tooltip
Icon=face-laugh
Profiles=test

[X-Action-Profile test]
Exec=dovecot

$ cat d.desktop 
[Desktop Entry]
Type=Action
Name=A
Name[de]=W
Description=A description
Tooltip=A tooltip
Icon=face-laugh
Profiles=test

[X-Action-Profile test]
Exec=aplay
```

On a side-note menus seem to fix order. Personally I wasn't aware of this so far but it seems to be according to those specs at http://www.nautilus-actions.org/?q=node/377.

tsujan:
@pmattern You're right. I forgot localization in this PR.
Anyhow, libfm-qt and pcmanfm-qt PR's can't be merged due to the fundamental changes that are going to be made. I leave them as they are.

tsujan:
This PR will be rewritten after https://github.com/lxde/libfm-qt/pull/63 is merged. Closing it now.

PR #22, opened 2016-04-12
Fix memory leak if restoring files from trash

philippwiesemann:
Thanks!

PR #17, opened 2016-02-03
Use grid layout to have proper properties alignment

npmiller:
The fields in the file search form were not aligned properly.

PR #63, opened 2017-02-08
Port libfm C API to C++ 11 with some API redesign/rewrite

PCMan:
This is a gigantic PR that requires more testing.
Previously I tried to add C++ wrapper for the libfm C APIs, but after spending a lot of time debugging the mixed C/C++ code across different libraries, tracing some memory errors or race condition, it's became less efficient as expected. After some evaluation, I started experimenting with porting the C APIs to C++-11 instead. This is a work in progress, but most of the parts has been ported.

The main focuses of this port/rewrite:
* Use C++-11 smart pointers and move semantics whenever possible.
* Strict const correctness.
* Make careful use of Qt/QObject APIs whenever possible so it integrates with Qt well
* Replace Qt-specific code with equivalent C++ STL containers and algorithms whenever possible
* Examine all APIs and try to redesign some of them, making them more modern C++
* Ensure thread safety and avoid race conditions
* Remove dependency on the libfm C library (still a work in progress)

The remaining parts that are not yet fully ported:
* Fm::FileOpsJob (main class handling file operations copy, move, unlink, ...etc)
* Archiver integration
* File templates
* DES-EMA nautilus extension support.
* FmConfig and FmFolderConfig
* FileLaunher

Otherwise the major parts have been ported to C++-11.
Please give your comments and help tests.
Thank you all.


tsujan:
@PCMan This needs at least a month of testing. Frankly, I can't read this amount of code to see if it has a problem. The most I can do is testing and also trying to find the relevent codes if a problem arises.

PCMan:
@tsujan Thanks. Please test against this one.
https://github.com/lxde/pcmanfm-qt/pull/450

palinek:
As @tsujan said, this is a big chunk of changes... so really hard for one to review.
I'm a (really) sporadic user of pcmanfm-qt, so my regular usage tests won't be much of use.

@PCMan would it be helpful for you, if I just review the changes of the commits (e.g one-by-one) and giving you some (nitpicking) comments?

PCMan:
@palinek Yes, any kind of help is appreciated. This is quite a huge change so I think it's good to be more careful. Previously I want to do this incrementally but later I realized that doing this touches nearly all files so it's not really possible to make the changes smaller.
It's still a work in progress, but after it's finished, I believe this may help long-term development.
So thank you all who help the review or testing.

tsujan:
@PCMan 
I read parts of this huge change and played with it a little. I have no doubt that you fully know what you're doing. However, after this change, I don't think that I could contribute to pcmanfm-qt anymore. Let me honestly explain my situation:

Previously, I could clearly see the relation between libfm-qt and libm. Everything was transparent and if I wanted to fix a bug or add a feature, I could quickly find the relevant libfm function(s) or see that purely Qt methods would be enough. Of course, it took me a while to get familiar with those codes but after that, everything became clear. The first C++ wrappers changed that but just a little.

Now, what I see is a very thick wall between libfm-qt and libfm; nothing is transparent to me and there's less room -- if any -- for maneuvers. I'm really sorry to say that I won't have enough free time or, maybe, enough programming skills to follow these changes.

PCMan:
@tsujan There's no wall between libfm-qt and libfm. They're in the same lib now. The old libfm C API is ported to C++ 11 and are in the src/core subdirectory. This is the place you can find the non-GUI libfm part.
The naming are mostly the same, but they are real C++ classes, not wrappers.
Error-prone manual reference counting and raw malloc/free are all replaced by C++ 11 smart pointers. Data structures which should not be changed by users are now correctly marked as constants.
While working on the old libfm C API, I discovered some design flaws which I did not noticed when I started the project 8 years ago. For example, some data structures are shared among threads but their contents can be changed at many places without enough protection. This causes hard-to-find bugs in many places. The new C++ implementation tries to put emphasis on these areas utilizing the new facilities provided by the standard C++-11. Cross-thread communications are handled by the safer mechanism provided by Qt. So at the moment it breaks source code compatibility a lot and causes problems, but in the long run this should help future development.
For the PRs you already provided, I'll port them to the new APIs if this PR can be merged.

src/core/gobjectptr.h
```diff
@@ -0,0 +1,74 @@
...
+#define FM2_GOBJECTPTR_H
+
+#include <glib.h>
+#include <glib-object.h>
+#include <cstddef>
+
+namespace Fm2 {
+
+template <typename T>
+class GObjectPtr {
```
palinek: Didn't you consider simplification of the `GObjectPtr` by using smart pointers? E.g.:
```c++
template <typename T>
class GObjectPtr : private std::shared_ptr<T> {
public:
    GObjectPtr(T* gobj = nullptr, bool add_ref = true) : std::shared_ptr<T>{gobj, [this] { if (*this) g_object_unref(get()); } } {
        if (add_ref && *this)
            g_object_ref(get());
    }
    GObjectPtr(const GObjectPtr& r) : std::shared_ptr<T>(r) {};
    GObjectPtr(const GObjectPtr&& r) : std::shared_ptr<T>(r) {};
    using std::shared_ptr<T>::get;
    using std::shared_ptr<T>::operator =;
    using std::shared_ptr<T>::operator ==;
    using std::shared_ptr<T>::operator !=;
    using std::shared_ptr<T>::operator bool;
};
```

Yes, this way `GObjectPtr` family increases the g_object_ref only once (the first time the shared_ptr is constructed), but the shared_ptr implementation assures that the g_object_unref is called also only when the last one is destructed.

PCMan:
@palinek Yes I'd like to use std::shared_ptr here, but that would be a waste. GObject already does thread-safe reference counting internally and using shared_ptr will introduce one additional memory allocation per object. It needs to allocates memory for the reference count and the GObject pointer. So after some more consideration, I created a specialized smart pointer for GObject.

tsujan:
> There's no wall between libfm-qt and libfm. ...

OK. I'm getting familiar with the new code by "translating" some patches I'd made for my personal use into it. So far so good. The wall exists but I'm trying to ignore it ;)

tsujan:
@PCMan https://github.com/lxde/pcmanfm-qt/pull/450 Cant't be compiled with this:
````
/usr/include/libfm-qt/core/filepath.h:4:28: fatal error: libfmqtglobals.h: No such file or directory
 #include "libfmqtglobals.h"
````

**Comment by PCMan** [comment]
@tsujan Thank you for the testing and finding the bugs. I also noticed them during testing. I already fixed these issues.
BTW, after testing this branch for nearly one month, I personally did not see major issues. May I move forward and merge this into master?

**Comment by tsujan** [comment]
@PCMan 
I like these branches to be merged as soon as possible but would please wait until I test them?

Before I test them, I'd like to ask another question: did you fix the compilation issues (https://github.com/lxde/libfm-qt/pull/63#issuecomment-279232456)? I guess you did the tests over already installed files. The above issues were encountered with a clean installation.

**Comment by tsujan** [comment]
@PCMan 
When I do a "makepkg", I get:
````
CMake Error in src/CMakeLists.txt:                                                                                    
  Target "fm-qt" INTERFACE_INCLUDE_DIRECTORIES property contains path:                                                
                                                                                                                      
    "/home/pedram/Files/libfm-qt/libfm-qt-pcman-libfm-c-11-port-merge/src/"                                           
                                                                                                                      
  which is prefixed in the source directory.
````
I remember I got the same error when I tested a month ago but somehow worked around it.

**Comment by tsujan** [comment]
After working around the above-mentioned issue, with pcmanfm-qt I still get:
````
In file included from /home/pedram/Files/pcmanfm-qt/pcmanfm-qt-pcman-libfm-c-11-port/pcmanfm/mainwindow.cpp:41:0:
/usr/include/libfm-qt/fileoperation.h:28:27: fatal error: core/filepath.h: No such file or directory
 #include <core/filepath.h>    
````

**Comment by tsujan** [comment]
After working around the last issue by replacing `<core/filepath.h>` with `"core/filepath.h"` in `fileoperation.h`, pcmanfm-qt seems to work fine at first glance.

However, when it's run without its desktop functionality from terminal, a *huge* amount of lines like this is created with right-clicking:
````
FileAction.match:  ........
````
Apparently they are about custom actions I have but, IMO, they're redundant and make it hard to see relevant lines.

**Comment by tsujan** [comment]
Sorry for comment after comment ;)

Custom actions have totally disappeared (in contrast to those lines in terminal)!

**Comment by luis-pereira** [comment]
@tsujan @PCMan 
Consider these two: https://github.com/lxde/libfm-qt/pull/64 https://github.com/lxde/libfm-qt/pull/65
I need them to build it.

**Comment by tsujan** [comment]
@luis-pereira, thanks!

**Comment by palinek** [comment]
Just my poor 2 cents....

Building/running on current debian testing:
- build of libfm-qt OK
- build of pcmanfm fails with the aforementioned <> vs "" include ... fixing that in libfm-qt -> pcmanfm builds correctly

I made a simple test with running under valgrind... and it seems there are some leaks. I'll try to address them in PR.

**Comment by palinek** [comment]
Trying to find the leaks... I think I found a crucial problem with the `Fm::IconInfo` .... give me a moment for more investigation and I'll post the description here..

**Comment by tsujan** [comment]
@palinek, now that you've delved into the code, could you also investigate the bug that hides custom actions?

**Comment by palinek** [comment]
Here is the description:
- the `Fm::IconInfo` (which should be the `shared_ptr` all the time) contains
```c++
    mutable QIcon qicon_;
```
which is (optionally) constructed only if `Fm::IconInfo:qicon()` is called:
```c++
QIcon IconInfo::qicon() const {
      if(Q_UNLIKELY(qicon_.isNull() && gicon_)) {
          if(!G_IS_FILE_ICON(gicon_.get())) {
              qicon_ = QIcon(new IconEngine{shared_from_this()});
          }
          else {
              qicon_ = internalQicon_;
          }
      }
      return qicon_;
  }
```
=> so the `qicon_` takes ownership of the newly constructed `Fm::IconEngine` (which is based on then parent `Fm::IconInfo`)
- the `Fm::IconEngine` holds `shared_ptr` of the `Fm::IconEngine`:
```c++
private:
    std::shared_ptr<const Fm::IconInfo> info_;
```


So if the `Fm::IconInfo::qicon_` ever gets constructed with the underlying `Fm::IconEngine`, we have a circular reference of `Fm::IconInfo` object => it will be never deleted.

(shared_ptr<Fm::IconInfo> -> QIcon -> Fm::IconEngine -> shared_ptr<Fm::IconInfo>)

PCMan:
@tsujan @palinek @luis-pereira Thanks a lot for your hard work. Just fixed the #include bug and merged all the PRs.
@tsujan Would you please provide me your custom action files so I can debug this part? Thank you.

tsujan:
@PCMan 

There's a more important problem: desktop files show no icon (just the default icon "application-x-desktop"). I opened `/usr/share/applications` and all files had  "application-x-desktop". However, I did that with the compilation from yesterday.

> Would you please provide me your custom action files so I can debug this part?

They are many (more than 30) and each of them calls a script from `~/bin/`, so I don't think they would be of any help.

I do my tests on another (older) computer because I need my laptop for other dev jobs.

palinek:
@PCMan Not sure here... but I believe that all the IconInfo object should be constructed by as `shared_ptr`, consider this:
```diff
diff --git a/src/core/volumemanager.h b/src/core/volumemanager.h
index 4766a91..021f369 100644
--- a/src/core/volumemanager.h
+++ b/src/core/volumemanager.h
@@ -30,8 +30,8 @@ public:
         return CStrPtr{g_volume_get_uuid(get())};
     }
 
-    IconInfo icon() {
-        return IconInfo{GIconPtr{g_volume_get_icon(get()), false}};
+    std::shared_ptr<const IconInfo> icon() const {
+        return IconInfo::fromGIcon(GIconPtr{g_volume_get_icon(get()), false});
     }
 
     // GDrive *        g_volume_get_drive(get());
@@ -97,8 +97,8 @@ public:
         return CStrPtr{g_mount_get_uuid(get())};
     }
 
-    IconInfo icon() const {
-        return IconInfo{GIconPtr{g_mount_get_icon(get()), false}};
+    std::shared_ptr<const IconInfo> icon() const {
+        return IconInfo::fromGIcon(GIconPtr{g_mount_get_icon(get()), false});
     }
 
     // GIcon * g_mount_get_symbolic_icon(get());
```

palinek:
> There's a more important problem: desktop files show no icon (just the default icon "application-x-desktop"). I opened /usr/share/applications and all files had "application-x-desktop". However, I did that with the compilation from yesterday.

confirmed here... a really dummy workaround can be done in following way:
```diff
diff --git a/src/core/fileinfo.cpp b/src/core/fileinfo.cpp
index 3e1f4d0..6e65b57 100644
--- a/src/core/fileinfo.cpp
+++ b/src/core/fileinfo.cpp
@@ -1,6 +1,7 @@
 #include "fileinfo.h"
 #include "fileinfo_p.h"
 #include <gio/gio.h>
+#include <QSettings>
 
 namespace Fm {
 
@@ -153,10 +154,24 @@ _file_is_symlink:
         }
     }
 
-    /* try file-specific icon first */
-    gicon = g_file_info_get_icon(inf.get());
-    if(gicon) {
-        icon_ = IconInfo::fromGIcon(gicon);
+    if (isNative() && isDesktopEntry())
+    {
+        QString file_path = parentDirPath.toString().get();
+        file_path += '/';
+        file_path += name_.c_str();
+        QSettings d{file_path, QSettings::IniFormat};
+        QString icon_name = d.value(QStringLiteral("Desktop Entry/Icon")).toString();
+        icon_ = IconInfo::fromName(icon_name.toStdString().c_str());
+        // FIXME: localized values!!
+        dispName_ = d.value(QStringLiteral("Desktop Entry/Name"), dispName_).toString();
+
+    } else
+    {
+        /* try file-specific icon first */
+        gicon = g_file_info_get_icon(inf.get());
+        if(gicon) {
+            icon_ = IconInfo::fromGIcon(gicon);
+        }
     }
 
 #if 0
```

tsujan:
@palinek 
libfm-qt should get the desktop icons from libfm -- the job is done by libfm.

palinek:
> libfm-qt should get the desktop icons from libfm -- the job is done by libfm.

? This whole PR is about making libfm-qt standalone... all logic from former libfm should be covered by libfm-qt now.

tsujan:
@palinek 
Good question, indeed!

@PCMan 
Is a goal of this PR independency from libfm? If not, there are lots of code in `FileInfo::setFromGFileInfo()` that seem to get the same info libfm already provides (I just took a look). If yes, two things come to my mind: (1) Why? Can't libfm be used efficiently? (2) The changes aren't enough for that goal (`#include <libfm/fm.h>` in `fileinfo.h`, for example).

tsujan:
I just saw this: "Remove dependency on the libfm C library (still a work in progress)". So, the PR can't be considered complete yet.

tsujan:
And other question arise too:

Should we use QObjects like QSettings in the core directory or should we keep it GLib based as far as possible and use QObjects outside it?

Is this a transition stage to make libfm-qt independent from GLib or the goal is sticking with GLib (which I prefer)?

I'll wait until this is complete and, in the meantime, I test it.

PCMan:
@tsujan @palinek The desktop entry handling is fixed.
@tsujan Yes, the final goal is making it independent from libfm C API (but still depend on glib). That's quite a lot of work so there will be a transition period. Most parts are already done, though. Mixing C and C++ work well without problems, but debugging and keep them synchronized becomes more and more difficult overtime. The main problem is, it's not possible or very difficult to use many new C++ features due to the C API compatible design. Even if the wrapper works, which "was" my plan in the past, it creates some unavoidable overhead, such as UTF8 string conversion and the use of raw pointers everywhere.
In addition, so many years have passed and glib/gio gain much more new features which were not available when C libfm is started. At that time libfm does many things now glib already offers. So after thinking about this problem for months, I think building a new C++-11 based API sitting on top of glib/gio only is quite reasonable.

tsujan:
After applying @palinek's workaround (or solution), the only problem is that of custom actions. I haven't found another issue so far.

tsujan:
While reading the PR to know what's wrong with custom actions, I encountered a crash. Unfortunately, being busy, I didn't get the backtrace and couldn't reproduce the crash afterward but I'm sure it was about `Fm::Folder::fromPath()`. Sorry for the poor info!

tsujan:
Yesterday, I needed to use pcmanfm-qt with this patch on my old computer and pcmanfm-qt hung on opening a system folder. Again, sorry for the poor info -- I can't use this on a regular basis -- but it seems there are serious problems and a lot of debugging will be needed.

tsujan:
An update:

After 2 weeks, I tested this again. pcmanfm-qt seems to be stable. The problem with custom actions persists, of course. The bookmark crash, that @agaida reported elsewhere, doesn't happen. The crash I mentioned previously may have been caused by bad compilation, so please ignore it! All system folders are opened correctly.

tsujan:
@PCMan 
A  crash is confirmed by @agaida and me here: https://github.com/lxde/pcmanfm-qt/issues/474

tsujan:
Another crash on adding any directory to bookmarks:
````
#0  0x00007ffff5a55e61 in g_file_get_uri () at /usr/lib/libgio-2.0.so.0
#1  0x00007ffff78b7cbc in Fm::FilePath::uri() const (this=<optimized out>) at /home/pedram/Experiment/libfm-qt/libfm-qt-pcman-libfm-c-11-port-merge/src/core/filepath.h:63
        uri = <optimized out>
        __for_range = 
          std::vector of length 7, capacity 8 = {<error reading variable: Cannot access memory at address 0x99>, std::shared_ptr (count 3, weak 0) 0x852090, std::shared_ptr (count 3, weak 0) 0x74ea40, std::shared_ptr (count 3, weak 0) 0x917b40, std::shared_ptr (count 3, weak 0) 0x917bd0, std::shared_ptr (count 3, weak 0) 0x917c50, std::shared_ptr (count 3, weak 0) 0x7d5600}
        buf = ""
        err = 0x0
#2  0x00007ffff78b7cbc in Fm::Bookmarks::save() (this=0x7f4db0) at /home/pedram/Experiment/libfm-qt/libfm-qt-pcman-libfm-c-11-port-merge/src/core/bookmarks.cpp:96
        uri = <optimized out>
        __for_range = 
          std::vector of length 7, capacity 8 = {<error reading variable: Cannot access memory at address 0x99>, std::shared_ptr (count 3, weak 0) 0x852090, std::shared_ptr (count 3, weak 0) 0x74ea40, std::shared_ptr (count 3, weak 0) 0x917b40, std::shared_ptr (count 3, weak 0) 0x917bd0, std::shared_ptr (count 3, weak 0) 0x917c50, std::shared_ptr (count 3, weak 0) 0x7d5600}
        buf = ""
        err = 0x0
#3  0x00007ffff62a4476 in  () at /usr/lib/libQt5Core.so.5
#4  0x00007ffff6298483 in QObject::event(QEvent*) () at /usr/lib/libQt5Core.so.5
#5  0x00007ffff6f4be0c in QApplicationPrivate::notify_helper(QObject*, QEvent*) () at /usr/lib/libQt5Widgets.so.5
#6  0x00007ffff6f53581 in QApplication::notify(QObject*, QEvent*) () at /usr/lib/libQt5Widgets.so.5
#7  0x00007ffff626bdf0 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () at /usr/lib/libQt5Core.so.5
#8  0x00007ffff62bf6be in QTimerInfoList::activateTimers() () at /usr/lib/libQt5Core.so.5
#9  0x00007ffff62bfbe1 in  () at /usr/lib/libQt5Core.so.5
#10 0x00007ffff54f1587 in g_main_context_dispatch () at /usr/lib/libglib-2.0.so.0
#11 0x00007ffff54f17f0 in  () at /usr/lib/libglib-2.0.so.0
#12 0x00007ffff54f189c in g_main_context_iteration () at /usr/lib/libglib-2.0.so.0
#13 0x00007ffff62c071f in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) () at /usr/lib/libQt5Core.so.5
#14 0x00007ffff626a24a in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () at /usr/lib/libQt5Core.so.5
#15 0x00007ffff627274c in QCoreApplication::exec() () at /usr/lib/libQt5Core.so.5
````

**Comment by agaida** [comment]
confirmed:
```
Thread 1 "pcmanfm-qt" received signal SIGSEGV, Segmentation fault.
std::__shared_ptr<Fm::BookmarkItem const, (__gnu_cxx::_Lock_policy)2>::__shared_ptr(std::__shared_ptr<Fm::BookmarkItem const, (__gnu_cxx::_Lock_policy)2>&&) (__r=<optimized out>, this=<optimized out>)
    at /usr/include/c++/6/bits/shared_ptr_base.h:936
936           : _M_ptr(__r._M_ptr), _M_refcount()
(gdb) bt
#0  std::__shared_ptr<Fm::BookmarkItem const, (__gnu_cxx::_Lock_policy)2>::__shared_ptr(std::__shared_ptr<Fm::BookmarkItem const, (__gnu_cxx::_Lock_policy)2>&&) (__r=<optimized out>, this=<optimized out>)
    at /usr/include/c++/6/bits/shared_ptr_base.h:936
#1  std::shared_ptr<Fm::BookmarkItem const>::shared_ptr(std::shared_ptr<Fm::BookmarkItem const>&&) (__r=<optimized out>, this=<optimized out>) at /usr/include/c++/6/bits/shared_ptr.h:230
#2  std::_Construct<std::shared_ptr<Fm::BookmarkItem const>, std::shared_ptr<Fm::BookmarkItem const> >(std::shared_ptr<Fm::BookmarkItem const>*, std::shared_ptr<Fm::BookmarkItem const>&&) (__p=<optimized out>)
    at /usr/include/c++/6/bits/stl_construct.h:75
#3  std::__uninitialized_copy<false>::__uninit_copy<std::move_iterator<std::shared_ptr<Fm::BookmarkItem const>*>, std::shared_ptr<Fm::BookmarkItem const>*> (__result=<optimized out>, __last=..., __first=...)
    at /usr/include/c++/6/bits/stl_uninitialized.h:75
#4  std::uninitialized_copy<std::move_iterator<std::shared_ptr<Fm::BookmarkItem const>*>, std::shared_ptr<Fm::BookmarkItem const>*> (__result=<optimized out>, __last=..., __first=...) at /usr/include/c++/6/bits/stl_uninitialized.h:126
#5  std::__uninitialized_copy_a<std::move_iterator<std::shared_ptr<Fm::BookmarkItem const>*>, std::shared_ptr<Fm::BookmarkItem const>*, std::shared_ptr<Fm::BookmarkItem const> > (__result=<optimized out>, __last=..., __first=...)
    at /usr/include/c++/6/bits/stl_uninitialized.h:281
#6  std::__uninitialized_move_if_noexcept_a<std::shared_ptr<Fm::BookmarkItem const>*, std::shared_ptr<Fm::BookmarkItem const>*, std::allocator<std::shared_ptr<Fm::BookmarkItem const> > > (__alloc=..., __result=<optimized out>, 
    __last=0x740090, __first=0x7400a0) at /usr/include/c++/6/bits/stl_uninitialized.h:304
#7  std::vector<std::shared_ptr<Fm::BookmarkItem const>, std::allocator<std::shared_ptr<Fm::BookmarkItem const> > >::_M_insert_aux<std::shared_ptr<Fm::BookmarkItem const> >(__gnu_cxx::__normal_iterator<std::shared_ptr<Fm::BookmarkItem const>*, std::vector<std::shared_ptr<Fm::BookmarkItem const>, std::allocator<std::shared_ptr<Fm::BookmarkItem const> > > >, std::shared_ptr<Fm::BookmarkItem const>&&) (this=this@entry=0x73faf0, 
    __position=<error reading variable: Cannot access memory at address 0x29>, __args#0=<unknown type in /usr/lib/debug/.build-id/72/bc8985d1634d7ed79f7901d0a2bf82e4c02967.debug, CU 0x30949c, DIE 0x341748>)
    at /usr/include/c++/6/bits/vector.tcc:370
#8  0x00007ffff7b44146 in std::vector<std::shared_ptr<Fm::BookmarkItem const>, std::allocator<std::shared_ptr<Fm::BookmarkItem const> > >::emplace<std::shared_ptr<Fm::BookmarkItem const> >(__gnu_cxx::__normal_iterator<std::shared_ptr<Fm::BookmarkItem const> const*, std::vector<std::shared_ptr<Fm::BookmarkItem const>, std::allocator<std::shared_ptr<Fm::BookmarkItem const> > > >, std::shared_ptr<Fm::BookmarkItem const>&&) (this=this@entry=0x73faf0, __position=..., 
    __args#0=__args#0@entry=<unknown type in /usr/lib/debug/.build-id/72/bc8985d1634d7ed79f7901d0a2bf82e4c02967.debug, CU 0x30949c, DIE 0x341748>) at /usr/include/c++/6/bits/vector.tcc:313
#9  0x00007ffff7b42278 in std::vector<std::shared_ptr<Fm::BookmarkItem const>, std::allocator<std::shared_ptr<Fm::BookmarkItem const> > >::insert(__gnu_cxx::__normal_iterator<std::shared_ptr<Fm::BookmarkItem const> const*, std::vector<std::shared_ptr<Fm::BookmarkItem const>, std::allocator<std::shared_ptr<Fm::BookmarkItem const> > > >, std::shared_ptr<Fm::BookmarkItem const>&&) (
    __x=<unknown type in /usr/lib/debug/.build-id/72/bc8985d1634d7ed79f7901d0a2bf82e4c02967.debug, CU 0x30949c, DIE 0x3415d5>, __position=..., this=0x73faf0) at /usr/include/c++/6/bits/stl_vector.h:1016
#10 Fm::Bookmarks::insert (this=0x73fad0, path=..., name=..., pos=pos@entry=-1) at ./src/core/bookmarks.cpp:44
#11 0x00000000004282a7 in PCManFM::MainWindow::on_actionAddToBookmarks_triggered (this=0x73fb20) at ./pcmanfm/mainwindow.cpp:619
#12 0x000000000045f53f in PCManFM::MainWindow::qt_static_metacall (_o=_o@entry=0x73fb20, _c=_c@entry=QMetaObject::InvokeMetaMethod, _id=_id@entry=50, _a=_a@entry=0x7fffffffcd50)
    at ./obj-x86_64-linux-gnu/pcmanfm/pcmanfm-qt_automoc.dir/moc_mainwindow_L2CLFKCLUYECYY.cpp:464
#13 0x0000000000460185 in PCManFM::MainWindow::qt_metacall (this=0x73fb20, _c=QMetaObject::InvokeMetaMethod, _id=50, _a=0x7fffffffcd50) at ./obj-x86_64-linux-gnu/pcmanfm/pcmanfm-qt_automoc.dir/moc_mainwindow_L2CLFKCLUYECYY.cpp:591
#14 0x00007ffff6d21784 in QMetaObject::activate(QObject*, int, int, void**) () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#15 0x00007ffff75d21c2 in QAction::triggered(bool) () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#16 0x00007ffff75d4bb0 in QAction::activate(QAction::ActionEvent) () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#17 0x00007ffff77441ca in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#18 0x00007ffff774b854 in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#19 0x00007ffff774c826 in QMenu::mouseReleaseEvent(QMouseEvent*) () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#20 0x00007ffff7620278 in QWidget::event(QEvent*) () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#21 0x00007ffff774eb53 in QMenu::event(QEvent*) () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#22 0x00007ffff75d8b8c in QApplicationPrivate::notify_helper(QObject*, QEvent*) () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#23 0x00007ffff75e11fd in QApplication::notify(QObject*, QEvent*) () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#24 0x00007ffff6cf59e0 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#25 0x00007ffff75df8ad in QApplicationPrivate::sendMouseEvent(QWidget*, QMouseEvent*, QWidget*, QWidget*, QWidget**, QPointer<QWidget>&, bool) () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#26 0x00007ffff763ad16 in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#27 0x00007ffff763d313 in ?? () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#28 0x00007ffff75d8b8c in QApplicationPrivate::notify_helper(QObject*, QEvent*) () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#29 0x00007ffff75e0341 in QApplication::notify(QObject*, QEvent*) () from /usr/lib/x86_64-linux-gnu/libQt5Widgets.so.5
#30 0x00007ffff6cf59e0 in QCoreApplication::notifyInternal2(QObject*, QEvent*) () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#31 0x00007ffff703db03 in QGuiApplicationPrivate::processMouseEvent(QWindowSystemInterfacePrivate::MouseEvent*) () from /usr/lib/x86_64-linux-gnu/libQt5Gui.so.5
#32 0x00007ffff703f685 in QGuiApplicationPrivate::processWindowSystemEvent(QWindowSystemInterfacePrivate::WindowSystemEvent*) () from /usr/lib/x86_64-linux-gnu/libQt5Gui.so.5
#33 0x00007ffff701d61b in QWindowSystemInterface::sendWindowSystemEvents(QFlags<QEventLoop::ProcessEventsFlag>) () from /usr/lib/x86_64-linux-gnu/libQt5Gui.so.5
#34 0x00007fffedeae4c0 in ?? () from /usr/lib/x86_64-linux-gnu/libQt5XcbQpa.so.5
#35 0x00007ffff5f807f7 in g_main_context_dispatch () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#36 0x00007ffff5f80a60 in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#37 0x00007ffff5f80b0c in g_main_context_iteration () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#38 0x00007ffff6d4a04f in QEventDispatcherGlib::processEvents(QFlags<QEventLoop::ProcessEventsFlag>) () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#39 0x00007ffff6cf39ca in QEventLoop::exec(QFlags<QEventLoop::ProcessEventsFlag>) () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#40 0x00007ffff6cfc13c in QCoreApplication::exec() () from /usr/lib/x86_64-linux-gnu/libQt5Core.so.5
#41 0x000000000042504c in PCManFM::Application::exec (this=<optimized out>) at ./pcmanfm/application.cpp:334
#42 0x000000000041d941 in main (argc=<optimized out>, argv=0x7fffffffdea8) at ./pcmanfm/pcmanfm.cpp:12
```

**Comment by tsujan** [comment]
Briefly, 3 issues are detected by @agaida and me for now:

(1) https://github.com/lxde/pcmanfm-qt/issues/474;
(2) Crash on adding bookmarks;
(3) Custom actions are missing.

I have to go back to my laptop from this old comuter that I use for such jobs. So, no further report from me today.

**Comment by tsujan** [comment]
BTW, `#include "icontheme.h"` is redundant in `foldermodelitem.h`.

**Comment by palinek** [comment]
the @7cf8893 should fix @agaida's crashes.

**Comment by tsujan** [comment]
@palinek, now that you're looking into the code, what do you think about https://github.com/lxde/pcmanfm-qt/issues/474? It seems to me that the problem is not in `FolderModelItem` but in `IconInfo` and that `qicon_` (in `iconinfo.cpp`) is somehow destroyed before calling `IconInfo::qicon();`

**Comment by palinek** [comment]
..next one @1f02e0d should fix the crash with icon... but next work should be done to use some decent fallback icon.

**Comment by tsujan** [comment]
https://github.com/lxde/libfm-qt/commit/1f02e0d surely fixes the crash but fallback icons are set at `iconinfo.cpp`; I don't think it's a good idea to set them again in another place. There may be a structural problem here.

**Comment by tsujan** [comment]
Found a simple solution. Something like this at the end of `FileInfo::setFromGFileInfo()`:
````
    if (!icon_ && mimeType_) {
        icon_ = mimeType_->icon();
    }
````

**Comment by tsujan** [comment]
Before going to sleep I summarize the situation:

(1) Thanks to @palinek's and @PCMan's commits, it seems that all crashes are fixed now. A little more testing is needed though.
(2) A commit like the one I suggested above is needed for fallback icons.
(3) Apparently the only remaining issue is that of custom actions.

After (2) is done and (3) is fixed, and if no other crash/problem is detected in the meantime, the PR could be merged for further testing by bleeding edge users, IMO.

So far, so good :)

**Comment by tsujan** [comment]
Today, while "translating" my patches and PR's into the new code, I encountered another crash. It happens when pcmanfm-qt uses Engrampa to make a 7z archive with the Compress menu-item::
````
#0  0x00007f9dd74df446 in strlen () at /usr/lib/libc.so.6
#1  0x00007f9dda1b247e in std::char_traits<char>::length(char const*) (__s=0x0)
    at /usr/include/c++/6.2.1/bits/char_traits.h:267
#2  0x00007f9dda1b247e in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::assign(char const*) (__s=0x0, this=0x7f9d9c013048) at /usr/include/c++/6.2.1/bits/basic_string.h:1253
#3  0x00007f9dda1b247e in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::operator=(char const*) (__s=0x0, this=0x7f9d9c013048) at /usr/include/c++/6.2.1/bits/basic_string.h:590
#4  0x00007f9dda1b247e in Fm::FileInfo::setFromGFileInfo(Fm::GObjectPtr<_GFileInfo> const&, Fm::FilePath const&) (this=0x7f9d9c013040, inf=..., parentDirPath=...)
    at /home/pedram/Experiment/libfm-qt/libfm-qt-pcman-libfm-c-11-port-merge/src/core/fileinfo.cpp:31
#5  0x00007f9dda1bbccd in __gnu_cxx::new_allocator<Fm::FileInfo>::construct<Fm::FileInfo, Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&>(Fm::FileInfo*, Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&) (this=<optimized out>, __p=<optimized out>) at /usr/include/c++/6.2.1/ext/new_allocator.h:120
#6  0x00007f9dda1bbccd in std::allocator_traits<std::allocator<Fm::FileInfo> >::construct<Fm::FileInfo, Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&>(std::allocator<Fm::FileInfo>&, Fm::FileInfo*, Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&) (__a=..., __p=<optimized out>) at /usr/include/c++/6.2.1/bits/alloc_traits.h:455
#7  0x00007f9dda1bbccd in std::_Sp_counted_ptr_inplace<Fm::FileInfo, std::allocator<Fm::FileInfo>, (__gnu_cxx::_Lock_policy)2>::_Sp_counted_ptr_inplace<Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&>(std::allocator<Fm::FileInfo>, Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&) (__a=..., this=0x7f9d9c013030)
    at /usr/include/c++/6.2.1/bits/shared_ptr_base.h:520
#8  0x00007f9dda1bbccd in std::__shared_count<(__gnu_cxx::_Lock_policy)2>::__shared_count<Fm::FileInfo, std::allocator<Fm::FileInfo>, Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&>(std::_Sp_make_shared_tag, Fm::FileInfo*, std::allocator<Fm::FileInfo> const&, Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&) (__a=..., this=<optimized out>)
    at /usr/include/c++/6.2.1/bits/shared_ptr_base.h:615
#9  0x00007f9dda1bbccd in std::__shared_ptr<Fm::FileInfo, (__gnu_cxx::_Lock_policy)2>::__shared_ptr<std::allocator<Fm::FileInfo>, Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&>(std::_Sp_make_shared_tag, std::allocator<Fm::FileInfo> const&, Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&) (__a=..., __tag=..., this=<optimized out>)
    at /usr/include/c++/6.2.1/bits/shared_ptr_base.h:1100
#10 0x00007f9dda1bbccd in std::shared_ptr<Fm::FileInfo>::shared_ptr<std::allocator<Fm::FileInfo>, Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&>(std::_Sp_make_shared_tag, std::allocator<Fm::FileInfo> const&, Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&) (__a=..., __tag=..., this=<optimized out>) at /usr/include/c++/6.2.1/bits/shared_ptr.h:319
#11 0x00007f9dda1bbccd in std::allocate_shared<Fm::FileInfo, std::allocator<Fm::FileInfo>, Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&>(std::allocator<Fm::FileInfo> const&, Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&) (__a=...)
    at /usr/include/c++/6.2.1/bits/shared_ptr.h:620
#12 0x00007f9dda1bbccd in std::make_shared<Fm::FileInfo, Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&>(Fm::GObjectPtr<_GFileInfo>&, Fm::FilePath&) () at /usr/include/c++/6.2.1/bits/shared_ptr.h:636
#13 0x00007f9dda1bbccd in Fm::FileInfoJob::exec() (this=0x136be30)
    at /home/pedram/Experiment/libfm-qt/libfm-qt-pcman-libfm-c-11-port-merge/src/core/fileinfojob.cpp:24
#14 0x00007f9dda1b920a in Fm::Job::run() (this=0x136be30)
    at /home/pedram/Experiment/libfm-qt/libfm-qt-pcman-libfm-c-11-port-merge/src/core/job.cpp:32
#15 0x00007f9dd8998d78 in  () at /usr/lib/libQt5Core.so.5
#16 0x00007f9dd7249454 in start_thread () at /usr/lib/libpthread.so.0
#17 0x00007f9dd75477df in clone () at /usr/lib/libc.so.6
````
The crash happenes only on creating 7z archives and doesn't happen when Engrampa is used directly from the main menu.

**Comment by tsujan** [comment]
The 7z crash shows a check is missing in `FileInfoJob::exec()`. It can be fixed so:
````
void FileInfoJob::exec() {
.....
            GFileInfoPtr inf{
                g_file_query_info(path.gfile().get(), gfile_info_query_attribs,
                                  G_FILE_QUERY_INFO_NONE, cancellable().get(), &err),
                false
            };
            if(!inf)
                return;
.....
````

**Comment by tsujan** [comment]
As for the last challenge, namely custom actions, there is typo in `FileMenu::FileMenu()`. This line
````
auto custom_actions = FileActionItem::get_actions_for_files(files);
````
should be:
````
auto custom_actions = FileActionItem::get_actions_for_files(files_);
````
However, this doesn't fix the issue completely.

**Comment by tsujan** [comment]
Ah! It was about typos. Custom actions came back :)

All problems, that are found until now, can be fixed easily. I'll make a few commits tomorrow if I find some time.

**Comment by tsujan** [comment]
Now that all problems are solved, another one shows up: because of some changes in the code, desktop item texts aren't drawn correctly. This problem may not be visible to all but I think I can fix it.

File: src/core/fileinfo.cpp [hunk 2]
```diff
@@ -0,0 +1,326 @@
...
+                dispName_ = displayName.get();
+            }
+            /* handle 'Hidden' key to set hidden attribute */
+            if(!isHidden_) {
+                isHidden_ = g_key_file_get_boolean(kf, "Desktop Entry", "Hidden", nullptr);
+            }
+        }
+        g_key_file_free(kf);
+    }
+
```
tsujan: A fallback is needed here, like this:
````
    if(!icon_ && mimeType_)
        icon_ = mimeType_->icon();
````

File: src/core/fileinfojob.cpp [hunk 3]
```diff
@@ -0,0 +1,31 @@
...
+void FileInfoJob::exec() {
+    for(const auto& path: paths_) {
+        if(!isCancelled()) {
+            GErrorPtr err;
+            GFileInfoPtr inf{
+                g_file_query_info(path.gfile().get(), gfile_info_query_attribs,
+                                  G_FILE_QUERY_INFO_NONE, cancellable().get(), &err),
+                false
+            };
+
```
tsujan: Here, lack of the following check may cause a crash:
````
            if(!inf)
                return;
````

File: src/customactions/fileaction.cpp [hunk 4]
```diff
@@ -0,0 +1,557 @@
...
+    GErrorPtr err;
+    enabled = g_key_file_get_boolean(kf, "Desktop Entry", "Enabled", &err);
+    if(err) { // key not found, default to true
+        err.reset();
+        enabled = true;
+    }
+    hidden = g_key_file_get_boolean(kf, "Desktop Entry", "Hidden", nullptr);
+    suggested_shortcut = CStrPtr{g_key_file_get_string(kf, "Desktop Entry", "SuggestedShortcut", nullptr)};
+
+    condition = unique_ptr<FileActionCondition> {new FileActionCondition(kf, "Desktop Entry")};
```
tsujan: This line is missing here:
````
has_parent = false;
````

File: src/customactions/fileaction.h [hunk 5]
```diff
@@ -0,0 +1,155 @@
...
+
+    const char* get_id() const {
+        return action->id.get();
+    }
+
+    FileActionTarget get_target() const {
+        if(action->type == FileActionType::ACTION) {
+            return FileActionTarget(static_cast<FileAction*>(action.get())->target);
+        }
+        return FILE_ACTION_TARGET_NONE;
```
tsujan: The above line should be replaced by this one:
````
return FILE_ACTION_TARGET_CONTEXT;
````

File: src/filemenu.cpp [hunk 6]
```diff
@@ -26,383 +26,346 @@
...
-          }
-          if(archiver->extract_cmd) {
-            QAction* action = new QAction(tr("Extract Here"), this);
-            connect(action, &QAction::triggered, this, &FileMenu::onExtractHere);
-            addAction(action);
-          }
+
+    // DES-EMA custom actions integration
+    // FIXME: port these parts to Fm API
+    auto custom_actions = FileActionItem::get_actions_for_files(files);
```
tsujan: `files` is a typo in the above line. It should be so:
````
auto custom_actions = FileActionItem::get_actions_for_files(files_);
````

File: src/filemenu.cpp [hunk 7]
```diff
@@ -26,383 +26,346 @@
...
+    for(auto& item: custom_actions) {
+        if(item && !(item->get_target() & FILE_ACTION_TARGET_CONTEXT)) {
+            continue;  // this item is not for context menu
         }
-        else {
-          QAction* action = new QAction(tr("Compress"), this);
-          connect(action, &QAction::triggered, this, &FileMenu::onCompress);
-          addAction(action);
+        if(item == custom_actions.front() && item->is_action()) {
```
tsujan: I think the above line should be:
````
if(item == custom_actions.front() && !item->is_action()) {
````

File: src/foldermenu.cpp [hunk 8]
```diff
@@ -25,277 +25,272 @@
...
+
+    // DES-EMA custom actions integration
+    FileInfoList files;
+    files.push_back(view->folderInfo());
+    auto custom_actions = FileActionItem::get_actions_for_files(files);
+    for(auto& item: custom_actions) {
+        if(item && !(item->get_target() & FILE_ACTION_TARGET_CONTEXT)) {
+            continue;  // this item is not for context menu
+        }
+        if(item == custom_actions.front() && item && item->is_action()) {
```
tsujan: As in `filemenu.cpp`, I think the above line should be:
````
if(item == custom_actions.front() && item && !item->is_action()) {
````

tsujan:
An update:

All detected problems can be fixed easily; c++11 PR's could be merged after a few commits.

However, we shouldn't ignore lximage-qt; it won't work after these PR's.

palinek:
> However, we shouldn't ignore lximage-qt; it won't work after these PR's.

Is it a big deal to adapt lximage-qt for the new libfm-qt ?

tsujan:
> Is it a big deal to adapt lximage-qt to build with the new libfm-qt ?

Having been busy with libfm-qt and pcmanfm-qt, I don't have any estimation yet.

tsujan:
Now, GTM.

palinek:
> Now, GTM.

Hooray.

tsujan:
Wait a minute! Found another problem: On renaming files, their *display* name is shown/changed instead of their real name. That's bad for desktop files.

Will add another commit for it.

tsujan:
Fixed!

Yet another (random) problem. User's home and Desktop view-items on the Places pane sometimes don't work correctly.

agaida:
applied, built and uploaded - it installs and runs fine as far as i can see - good job

tsujan:
Please test this:

(1) Start from you home and go to a folder that has a subfolder with a subfolder;
(2) Go to the the first and then, to the second subfolder;
(3) Click on the back button (on toolbar) once;
(4) Click on the user's home button inside Places pane.

At (4), the directory isn't changed to user's home but to the parent folder.

tsujan:
Corrected the description of the above test.

tsujan:
OK, there's a problem with path buttons.

tsujan:
Ah, found it :) There was a strange mistake at `PathBar::setPath()`.

agaida:
applied, uploaded, installed - runs fine  :)

tsujan:
I encountered a case of high CPU usage by pcmanfm-qt. Its cause was unknown to me. It might be about thumbnail creation (but *i'm not sure at all*) because I was downloading a video onto the desktop by Firefox. However, Firefox uses `application/x-partial-download`, so that there can't be a video thumbnail until the download is finished. The CPU usage became normal after I restarted pcmanfm-qt (while the download in question was still in progress) and remained normal.

I have no clue about the cause and optimistically (or foolishly?) ignore this until it happens again ;)

tsujan:
Sorry for so  many alternating good news and bad news! This time, it's a mixture of bad news and magic. I hope it'll be followed by logical good news tomorrow. It's related to the above comment.

During any video download with Firefox, I start pcmanfm-qt with terminal in the desktop mode. Then, I open a windowed instance too and go to the desktop folder with it. Once I change directory to any desktop subfolder, I get millions of this line in terminal:
````
QSortFilterProxyModel: index from wrong model passed to mapFromSource
````
This doesn't happen after I stop download, remove the unfinished video from desktop and restart pcmanfm-qt in desktop mode with terminal. Always reproducible here.

tsujan:
Another issue: Thumbnailing is incorrect. When a thumbnail is created, it'll never be updated. For example, if an image is edited, its thumbnail isn't updated. Or if a image is put in a folder, which contained another image with the same name (but deleted), the new image will have the thumbnail of the old one. Reloading the folder or even restarting the app has no effect.

palinek:
> Another issue: Thumbnailing is incorrect. ...

I think we should merge this and open an issue for each (new) problem found. This here becomes too cluttered.

agaida:
+1, full ack

palinek:
> +1, full ack

Do we want to make a minor release before?

tsujan:
If this isn't merged, I might find and fix problems ad infinitum. I don't like the thumbnail problem and the high CPU usage it causes but I agree with @palinek that it's better to merge this and fix problems in separate PR's.

Shouldn't we wait for @PCMan?

tsujan:
As for the minor release, I have no idea because I don't know what it it. I just don't want this to cause trouble for ordinary users; it isn't ready for them.

tsujan:
Thumbnail update problem is fixed but there may be other problems in thumbnail making too.

agaida:
I would just merge this and work on it to bring it into a release ready state - so in that case only users of bleeding edge (AUR) or some people that run the latest master are will get into trouble.

tsujan:
> so in that case only users of bleeding edge (AUR) or some people that run the latest master are will get into trouble.

On second thought... maybe I pamper ordinary users when I say it isn't ready for them; can't be objective in this case... Anyhow, it's good to merge.

BTW, pcmanfm-qt's PR has conflicts; I don't know if there's a "merge by force".

agaida:
Ordinary users don't compile and install from source, not even the Gentoo ones - so no mercy :P

tsujan:
Preferably don't do this; just believe my words:

Go to `~/.cache/thumbnails/normal` and you CPU will get quite busy. The reason is that the thumbnailer tries to make thumbnails of thumbnails of thumbnails of....

palinek:
@tsujan @PCMan please, consider also the #68 before merging this.

tsujan:
The positions of items can't be changed by DND on desktop. "Stick to Current Position" works but the position doesn't change. I see this as an important bug and try to find its cause.

tsujan:
The file launcher makes pcmanfm-qt crash on launching a desktop entry shortcut that points to a nonexistent file. This bug should be fixed later, when @PCMan ports file launcher to the new API.

PCMan:
@tsujan @palinek I felt sorry about the mess I've created and really appreciate your hard work to clean it up. Just tested all the recent fixes and they seem to work well. @tsujan I tried to debug the file launcher bug but it seems that the bug is in libfm C library so it should not block this PR.
Shall we go on and merge this one as it is right now, and create new issues if new problems are found later?

PCMan:
Sorry guys. I'll not submit any huge and messy PR like this one in the future. Will do them in smaller pieces with more communication in advance.

tsujan:
@PCMan 
You did a great work here. The code structure is robust. Typos and mistakes are expected with such a big PR. Fortunately, all detected problems are fixed; new problems could be fixed in separate PR's. So, please merge this and https://github.com/lxde/pcmanfm-qt/pull/450 (the latter needs conflict resolving)!

Please also continue this line of work until libfm-qt is completely independent of libfm. Then, pcmanfm-qt will be the best FM on the planet. Thanks!

palinek:
After merging this, please, consider also the PR for lxqimage-qt lxde/lximage-qt#90

tsujan:
One word before merging this (I do NOT mean that the following problem should be fixed before merging): I'm pretty sure that the workarounds (for the old bug https://sourceforge.net/p/pcmanfm/bugs/686/) that are added to the ends of `Folder::eventFileAdded()` and `Folder::eventFileDeleted()` are wrong and cause various problems (on downloading files to a folder). I'll try to fix the problem after this is merged and might also open an issue for it.

agaida:
so - if we all aggree that we should work on remaining problems after merging this "monster" we should merge it i guess - any objections?

tsujan:
Merge them! I don't think anyone has an objection.

tsujan:
Self-talk:
Actually, the above-mentioned workaround aside, there should be some mistakes in `folder.cpp`. *Sometimes*, when I download `lxqt-build-tools-master.zip` onto Desktop with Qupzilla, for example, and then delete it, its icon remains on Desktop. I get lines like these by using qDebug (edited):
````
"lxqt-build-tools-master.zip.download" is added to paths_to_add
"lxqt-build-tools-master.zip" is added to paths_to_add
"lxqt-build-tools-master.zip.download" is added to paths_to_del
and removed
"lxqt-build-tools-master.zip" is added to paths_to_add
````
`xqt-build-tools-master.zip.download` is created, then `lxqt-build-tools-master.zip` is created and the former is deleted but the last line doesn't seem logical.

And when I delete `lxqt-build-tools-master.zip`:
````
"lxqt-build-tools-master.zip"
is NOT in files_
and is NOT added to paths_to_del
````
This isn't always reproducible but occurs enough to get the above result.

agaida:
built, uploaded and installed libfm-qt, pcmanfm-qt and lximgage-qt to snapshots - works fine so far, great job.

tsujan:
> works fine so far, great job.

To me, the above-mentioned issue is a serious bug. A few minutes ago I found out that the list of files on Desktop (the unordered map `files_` in `folder.cpp`) can contain an element with a null key -- as if a file-name is a null string! Yet another needle in the haystack...

tsujan:
I think I found the cause -- this time, it wasn't a typo. The pointer returned by `c_str()` can be invalidated easily, hence the possibility of elements with an *invalidated* key. I'll add a PR after being certain about that and some testing.

PR #65, opened 2017-02-22
Use const iterators

luis-pereira:
Prevent a detach and with the latest lxqt-build-tools prevent a FTBFS on
debug mode. The FTBFS is there on purpose so we can spot non needed
detachments.


PCMan:
Rebase and merged. Thanks!

PR #67, opened 2017-04-07
Inline Renaming

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/468.

This is one of the possible implementations of inline renaming as a property of list views (icon, thumbnail and compact views) and independent of pcmanfm-qt, so that it's available to lximage-qt too. It coexists with dialog renaming and shares the same error dialogs with it.

Renaming of the current item is started by the context-menu "Rename" item when only one file/folder is selected, or by `Ctrl+R`. It is finished by Enter/Return or by removing focus from the item, and canceled by Esc. (`F2` is still for dialog renaming.)

If several items are selected, the context-menu "Rename" item will behave as before, namely it will call the rename dialog.

NOTE: Due to a design problem in pcmanfm-qt, desktop has its own `QStyledItemDelegate`. Therefore, its inline renaming should be implemented separately. I will make another PR for it.

### DIFF
--- a/src/filemenu.cpp
+++ b/src/filemenu.cpp
@@ -38,14 +38,16 @@ namespace Fm {
 
 FileMenu::FileMenu(FmFileInfoList* files, FmFileInfo* info, FmPath* cwd, QWidget* parent):
   QMenu(parent),
-  fileLauncher_(NULL) {
+  fileLauncher_(NULL),
+  view_ (NULL) {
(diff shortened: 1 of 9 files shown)


tsujan:
This PR will be rewritten after https://github.com/lxde/libfm-qt/pull/63 is merged. Closing it now.

PR #68, opened 2017-04-13
GObjectPtr: Detect & handle "self-assignment"

palinek:
W/o checking if the we are assigning the "same" pointer, we can end in
letting the refcount go to zero. With "same" pointer check we also
eliminate unneed g_object_unref/g_object_ref pairs on the same pointer.

This simplified example shows the erroneous behavior and triggers the
"GLib-GObject-CRITICAL" messages:

```c++
Fm::GObjectPtr<GFile>& foo(Fm::GObjectPtr<GFile>& f)
{
    return f;
}

const Fm::GObjectPtr<GFile>& foo_const(const Fm::GObjectPtr<GFile>& f)
{
  return f;
}

int main(int argc, char **argv)
{
  std::cerr << "1\n";
  Fm::GObjectPtr<GFile> file{g_file_parse_name("test.txt"), false};
  GFile * f = file.get();
  file = f;

  std::cerr << "2\n";
  Fm::GObjectPtr<GFile> f2{g_file_parse_name("test2.txt"), false};
  f2 = foo(f2);

  std::cerr << "3\n";
  Fm::GObjectPtr<GFile> f3{g_file_parse_name("test3.txt"), false};
  const Fm::GObjectPtr<GFile> & f3_r{f3};
  f3 = foo_const(f3_r);

  std::cerr << "4\n";
  return 0;
}
```
Output:
```
1

(process:17367): GLib-GObject-CRITICAL **: g_object_ref: assertion 'G_IS_OBJECT (object)' failed
2

(process:17367): GLib-GObject-CRITICAL **: g_object_ref: assertion 'G_IS_OBJECT (object)' failed
3

(process:17367): GLib-GObject-CRITICAL **: g_object_ref: assertion 'G_IS_OBJECT (object)' failed
4
```

### DIFF
--- a/src/core/gobjectptr.h
+++ b/src/core/gobjectptr.h
@@ -21,12 +21,10 @@ public:
             g_object_ref(gobj_);
     }
 
-    GObjectPtr(const GObjectPtr& other): GObjectPtr{} {
-        *this = other;
+    GObjectPtr(const GObjectPtr& other): gobj_{other.gobj_ ? reinterpret_cast<T*>(g_object_ref(other.gobj_)) : nullptr} {
     }
 
-    GObjectPtr(GObjectPtr&& other): GObjectPtr{} {
-        *this = other;
+    GObjectPtr(GObjectPtr&& other): gobj_{other.release()} {
     }
 
     ~GObjectPtr() {
@@ -51,6 +49,9 @@ public:
     }
 
     GObjectPtr& operator = (const GObjectPtr& other) {
+        if (*this == other)
+            return *this;
+
         if(gobj_ != nullptr)
             g_object_unref(gobj_);
         gobj_ = other.gobj_ ? reinterpret_cast<T*>(g_object_ref(other.gobj_)) : nullptr;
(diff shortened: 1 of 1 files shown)


tsujan:
Interesting! Did you encounter a reproducible problem with pcmanfm-qt first?

tsujan:
@palinek 
This causes crash in `GAppInfoPtr` (used in `filemenu_p.h`) on right clicking any file.

tsujan:
I guess the problem is in `filemenu_p.h`, not here.

tsujan:
@palinek 
Would you fix, *in this PR*, `filemenu_p.h` (by replacing `app` with `appInfo_` wherever needed) and also find whether there are other instances elsewhere, or should I do so after this is merged?

PCMan:
@palinek Thanks a lot for the fix! I overlooked the issue. LGTM.
However, would it be simpler to just do this instead?
`
auto tmp_new_obj = object_ref(new_gobj);
if(gobj_ != nullptr)
    g_object_unref(gobj_);
gobj_ = tmp_new_obj;
`
By changing the order of ref/unref, it still prevents releasing the same pointer while no further checks are needed. But there should be no visible performance difference. :-)

PCMan:
Merged. Thanks a lot!

tsujan:
@PCMan 
As I mentioned above, this causes crash on right clicking any file because now the move ctor releases the resources. I'll add a fix.

palinek:
> By changing the order of ref/unref, it still prevents releasing the same pointer while no further checks are needed. But there should be no visible performance difference. :-)

Right, but by adding the ìf` before the ref/unref does omit the ref/unref pair at all. I have no knowledge how the ref/uref are implemented, but I think that one branch is cheaper than the ref/unref pair. We can also add the `Q_UNLIKELY` (or the glib counterpart) hint for the compiler...

palinek:
> and also find whether there are other instances elsewhere

(not at the devel machine right now) I had a look for the usage of `std::move` elsewhere and IIRC there is only one wrong usage in the return statement which avoids the copy elision optimization done by compilers.

PR #69, opened 2017-04-17
Fix invalid pointers

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/475 by using `const std::string` instead of `const char*` to keep track of files inside a folder because the pointer returned by `c_str()` may be invalidated easily and result in the view showing items whose corresponding files are deleted.

### DIFF
--- a/src/core/folder.cpp
+++ b/src/core/folder.cpp
@@ -129,7 +129,8 @@ bool Folder::isLoaded() const {
 }
 
 std::shared_ptr<const FileInfo> Folder::fileByName(const char* name) const {
-    auto it = files_.find(name);
+    std::string nameStr(name);
+    auto it = files_.find(nameStr);
     if(it != files_.end()) {
         return it->second;
     }
@@ -217,14 +218,14 @@ void Folder::onFileInfoFinished() {
             dirInfo_ = info;
         }
         else {
-            auto it = files_.find(info->name().c_str());
+            auto it = files_.find(info->name());
             if(it != files_.end()) { // the file already exists, update
                 files_to_update.push_back(std::make_pair(it->second, info));
             }
             else { // newly added
                 files_to_add.push_back(info);
             }
-            files_[info->name().c_str()] = info;
+            files_[info->name()] = info;
         }
 
         if(!files_to_add.empty()) {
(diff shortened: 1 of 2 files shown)


src/core/folder.h
```diff
@@ -168,7 +168,7 @@ private Q_SLOTS:
     bool wants_incremental;
     bool stop_emission; /* don't set it 1 bit to not lock other bits */
 
-    std::unordered_map<const char*, std::shared_ptr<const FileInfo>, CStrHash, CStrEqual> files_;
+    std::unordered_map<const std::string, std::shared_ptr<const FileInfo>, std::hash<std::string>> files_;
```
palinek: The `std::hash<Key>` is a default template parameter... Is there a good reason for stating it explicitly here?

src/core/folder.cpp
```diff
@@ -129,7 +129,8 @@ bool Folder::isLoaded() const {
 }
 
 std::shared_ptr<const FileInfo> Folder::fileByName(const char* name) const {
-    auto it = files_.find(name);
+    std::string nameStr(name);
+    auto it = files_.find(nameStr);
```
palinek: There is no need to create the temp `std::string` here. The `std::unordered_map::find(const Key &)` can handle it itself.

palinek:
Just a note: the problem here is that the Key in the `unordered_map` is the `name()` of the held `const FileInfo`... so it can't be invalidated easily (as the container holds the const instance of it). But the assignments like:
```cpp
            files_[info->name().c_str()] = info;
```
can cause that the the `info` will be held as data of "the old `const FileInfo` `name()`" which will go out of scope and the Key will get invalidated.
So the `std::string` as Key does solve this.

If we really want to avoid the string (const char*) copying.... we should do the assignments like:
```c++
            auto i_insert = files_.find(info->name().c_str());
              if (i_insert != files_.end())
              {
                  files_.erase(i_insert);
              }
              files_.insert({info->name().c_str(),  info});
```
...but it would be very error prone...

src/core/folder.h
tsujan: `std::hash<Key>` is for `std::string`, not for `const std::string`. Compilation would fail without adding it.

src/core/folder.cpp
tsujan: OK.

tsujan:
> So the std::string as Key does solve this.

Exactly, and without creating hassles.

agaida:
GTM

PR #70, opened 2017-04-18
Move fixes

palinek:
I had a look on the usage of std::move() (as @tsujan suggested in #68) and found theese...

### DIFF
--- a/src/placesview.cpp
+++ b/src/placesview.cpp
@@ -172,7 +172,7 @@ void PlacesView::onClicked(const QModelIndex& index) {
 
 void PlacesView::setCurrentPath(Fm::FilePath path) {
     currentPath_ = std::move(path);
-    if(path) {
+    if(currentPath_) {
         // TODO: search for item with the path in model_ and select it.
         PlacesModelItem* item = model_->itemFromPath(currentPath_);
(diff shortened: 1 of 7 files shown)


src/placesview.cpp
```diff
@@ -172,7 +172,7 @@ void PlacesView::onClicked(const QModelIndex& index) {
 
 void PlacesView::setCurrentPath(Fm::FilePath path) {
     currentPath_ = std::move(path);
-    if(path) {
+    if(currentPath_) {
         // TODO: search for item with the path in model_ and select it.
```
tsujan: There are more examples of this in the code and all of them are OK because the move c-ctor of FilePath doesn't free any resource. However, IMO, not using the moved-from object (with some exceptions) seems safer and better for code readability.

palinek: > move c-ctor of FilePath doesn't free any resource

No move c-tor/assignment should ever free resources (as the resources are moved to other object).

tsujan: By freeing resources I meant "freeing" them from the first object and hence, *moving* them to the other one -- that isn't done for FilePath; so, the moved-from object could be used without problem in its case.

PR #71, opened 2017-04-18
[WIP] Address compiler warnings

palinek:
Warnings left:
- [x] copy elision & std::move (addressed in #70)
- [x] usage of deprecated `Fm::_convertPathList`, `Fm::_convertPath` -> ~~we should either undeprecate them or do not use them~~
- [x] unused variable `trash_query` -> ~~the untrash job seems to be unimplemented !?~~

Update: the last two are valid/needed warnings as the porting process is ongoing...

### DIFF
--- a/src/appchooserdialog.cpp
+++ b/src/appchooserdialog.cpp
@@ -48,7 +48,7 @@ bool AppChooserDialog::isSetDefault() const {
     return ui->setDefault->isChecked();
 }
 
-static void on_temp_appinfo_destroy(gpointer data, GObject* objptr) {
+static void on_temp_appinfo_destroy(gpointer data, GObject* /*objptr*/) {
     char* filename = (char*)data;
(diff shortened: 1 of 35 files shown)


PCMan:
@palinek The deprecation marks for Fm::_convertPathList, Fm::_convertPath are deliberate. These are for the parts that still depend on C libfm. When the remaining parts are ported to C++ 11, we should delete the compatibility module altogether.
Not only the untrash job, but the remaining jobs are not completely ported either. Most of the current file operations still calls libfm C API internally at the moment.

palinek:
So can this be merged as is? All the warning are really disrupting... and potentially hiding real problems.

tsujan:
@palinek 
@PCMan's comment was the answer to your questions and the other warnings are (mostly?) in those areas that will be rewritten by him. This is GTM.

PR #72, opened 2017-04-18
Don't make items current on mouseover

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/470.

This is about a usability issue I encountered when comparing PCManFM-Qt to Dolphin. The "focus rectangle", which is for keyboard navigation, would be useless if items were made current on mouseover.


agaida:
GTM

PR #73, opened 2017-04-19
Replace start tilde in PathEdit

tsujan:
This follows https://github.com/lxde/pcmanfm-qt/pull/481.

Since neither `Fm::FilePath` nor autocompletion understands a starting tilde, this is the safest and easiest way of letting users type it in the location bar with autocompletion, IMO.

### DIFF
--- a/src/pathedit.cpp
+++ b/src/pathedit.cpp
@@ -26,6 +26,7 @@
 #include <QThread>
 #include <QDebug>
 #include <QKeyEvent>
+#include <QDir>
 #include <libfm/fm.h>
 
 namespace Fm {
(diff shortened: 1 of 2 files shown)


agaida:
GTM

src/pathedit.cpp
```diff
@@ -118,7 +120,23 @@ bool PathEdit::event(QEvent* e) {
     return QLineEdit::event(e);
 }
 
+void PathEdit::onTextEdited(const QString& text) {
+    // just replace start tilde with home path if text is changed by user
+    if(text == QLatin1String("~") || text.startsWith(QLatin1String("~/"))) {
+        QString txt(text);
+        txt.replace(0, 1, QDir::homePath());
```
palinek: Just a note... there is a API in libfm-qt for the home path -> `Fm::FilePath::homeDir()`

tsujan: I know but getting a `QString` directly is cleaner than getting a `FilePath` -- which is acquired from `const char*`, in the first place -- and then converting it to `QString`, not in one step.

PR #74, opened 2017-04-21
Really cancel pending thumbnail jobs on chdir

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/484.

### DIFF
--- a/src/foldermodel.cpp
+++ b/src/foldermodel.cpp
@@ -123,6 +123,7 @@ void FolderModel::loadPendingThumbnails() {
     for(auto& item: thumbnailData_) {
         if(!item.pendingThumbnails_.empty()) {
             auto job = new Fm::ThumbnailJob(item.pendingThumbnails_, item.size_);
+            pendingThumbnailJobs_.push_back(job);
             job->setAutoDelete(true);
             connect(job, &Fm::ThumbnailJob::thumbnailLoaded, this, &FolderModel::onThumbnailLoaded, Qt::BlockingQueuedConnection);
             connect(job, &Fm::ThumbnailJob::finished, this, &FolderModel::onThumbnailJobFinished, Qt::BlockingQueuedConnection);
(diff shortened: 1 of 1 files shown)


src/foldermodel.cpp
```diff
@@ -122,7 +122,7 @@ void FolderModel::loadPendingThumbnails() {
     hasPendingThumbnailHandler_ = false;
     for(auto& item: thumbnailData_) {
         if(!item.pendingThumbnails_.empty()) {
-            auto job = new Fm::ThumbnailJob(item.pendingThumbnails_, item.size_);
+            auto job = new Fm::ThumbnailJob(std::move(item.pendingThumbnails_), item.size_);
```
palinek: The `item.pendingThumbnails_` can't be referenced/used after this call? I don't have the the snippet of C++ standard to cite, but the [cppreference](http://en.cppreference.com/w/cpp/language/move_constructor) states:

> Move constructors typically "steal" the resources held by the argument (e.g. pointers to dynamically-allocated objects, file descriptors, TCP sockets, I/O streams, running threads, etc), rather than make copies of them, and leave the argument in some valid but otherwise indeterminate state. For example, moving from a std::string or from a std::vector may result in the argument being left empty. However, this behavior should not be relied upon.

So for the sake of correctness, the `item.pendingThumbnails_` should be reinitialized/cleared explicitly.

tsujan: No, `item.pendingThumbnails_` is cleared correctly here and it can be used again by adding infos to it later.

tsujan: Actually, I had a hard time to find the (second) cause of extra CPU usage until I saw the obvious ;)

palinek: > No, item.pendingThumbnails_ is cleared correctly here

Because you know the internal implementation of the `QVector` on which you rely upon here, but...

tsujan: This is *again* about our different (or maybe, opposite) viewpoints, not about something real...

agaida:
GTM

PR #75, opened 2017-04-24
Fixes a FTBFS in superbuild mode

luis-pereira:
It was introduced with the c++11-port merge.
Closes https://github.com/lxde/lxqt/issues/1281.


agaida:
works, LGTM

PR #76, opened 2017-04-27
Continue file selection on next/previous row

vc-01:
In folder view when left/right arrow key is pressed and current item is first/last on the row, selection continues on previous/next row accordingly.

Before this patch, selection stopped on the current row.

****
First, I'm not sure if it's done correctly, tell me if not.

Runtime check for current view mode is there because in "Compact View" left/right arrow keys are used differently.

Checks for first and last items overflow/underflow are done in FolderModel class, so not needed here.

This behaviour was missing at least to me, and is standard in pcmanfm(gtk), thunar and win explorer too.

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -182,6 +182,26 @@ void FolderViewListView::mouseDoubleClickEvent(QMouseEvent* event) {
     activationAllowed_ = activationWasAllowed;
 }
 
+QModelIndex FolderViewListView::moveCursor(CursorAction cursorAction, Qt::KeyboardModifiers modifiers) {
+    QAbstractItemModel* model_ = model();
+
+    if(model_ && currentIndex().isValid()) {
+        FolderView::ViewMode viewMode = static_cast<FolderView*>(parent())->viewMode();
+        if((viewMode == FolderView::IconMode) || (viewMode == FolderView::ThumbnailMode)) {
+            int next = (layoutDirection() == Qt::RightToLeft) ? - 1 : 1;
+
+            if(cursorAction == QAbstractItemView::MoveRight) {
+                return model_->index(currentIndex().row() + next, 0);
+            }
+            else if(cursorAction == QAbstractItemView::MoveLeft) {
+                return model_->index(currentIndex().row() - next, 0);
+            }
+        }
+    }
+
+    return QListView::moveCursor(cursorAction, modifiers);
+}
+
 void FolderViewListView::activation(const QModelIndex& index) {
     if(activationAllowed_) {
         Q_EMIT activatedFiltered(index);
--- a/src/folderview_p.h
+++ b/src/folderview_p.h
@@ -63,6 +63,9 @@ public:
 Q_SIGNALS:
   void activatedFiltered(const QModelIndex &index);
 
+protected:
+  virtual QModelIndex moveCursor(CursorAction cursorAction, Qt::KeyboardModifiers modifiers);
+
 private Q_SLOTS:
   void activation(const QModelIndex &index);
 


src/folderview.cpp
```diff
@@ -182,6 +182,27 @@ void FolderViewListView::mouseDoubleClickEvent(QMouseEvent* event) {
     activationAllowed_ = activationWasAllowed;
 }
 
+QModelIndex FolderViewListView::moveCursor(CursorAction cursorAction, Qt::KeyboardModifiers modifiers) {
+    QAbstractItemModel* model_ = model();
+
+    if(!model_ || !currentIndex().isValid()) {
+        return QModelIndex();
```
tsujan: Here, `return QModelIndex();` should be replaced by `return QListView::moveCursor(cursorAction, modifiers);`, otherwise there will be no action after opening a folder.

Preferably, you could put your code under the condition `if(model_ && currentIndex().isValid())` and have only one instance of `return QListView::moveCursor(cursorAction, modifiers);` at the end (outside the condition).

vc-01:
Thank you for this amazingly fast code review, I much appreciate it.

I pushed update as a new (squashed) commit - I hope this is the preferable way instead of github side commit squashing.

tsujan:
This is good but yet another problem (that may not be predictable easily) remains, namely RTL. If you run pcmanfm-qt with `pcmanfm-qt --reverse`, I think you'll see that right and left keys should be reversed (please test! I'm pretty sure about it but didn't have time to test).

Just a suggestion:
```cpp
            int next = (layoutDirection() == Qt::RightToLeft) ? - 1 : 1;
            if(cursorAction == QAbstractItemView::MoveRight) {
                return model_->index(currentIndex().row() + next, 0);
            }
            else if(cursorAction == QAbstractItemView::MoveLeft) {
                return model_->index(currentIndex().row() - next, 0);
            }
```
Although this should be OK, I haven't tried it. Please test it and if OK, add something like it -- in any way that seems appropriate to you -- to your code for RTL to be correct. Thanks in advance!

palinek:
> This is good but yet another problem (that may not be predictable easily) remains, namely RTL.

Note that in the Qt the `moveCursor()` functions are using something like:
```c++
    if (isRightToLeft()) {
        if (cursorAction == MoveLeft)
            cursorAction = MoveRight;
        else if (cursorAction == MoveRight)
            cursorAction = MoveLeft;
    }
```

tsujan:
> Note that in the Qt the moveCursor() functions are using something like...

No difference in the result.

tsujan:
@palinek If you meant that RTL is automatically considered by `cursorAction`, I tested it right now: No! It isn't.

vc-01:
@tsujan Yes, you were right with reversed arrow keys movement in RTL mode. I'll send a corrected patch.

tsujan:
> @tsujan Yes, you were right with reversed arrow keys movement in RTL mode.

That's because we're dealing with MoveRight/MoveLeft here, not with MoveNext/MovePrevious.

vc-01:
I tested it, your one-liner solved it.

tsujan:
Everything is OK. GTM.

tsujan:
@agaida, thanks for merging all these new PR's! Now, pcmanfm-qt is stable again and development can continue :)

agaida:
@all: Hmm - it's sometimes hard for me to decide, which things are good to merge. So in future it would be nice if the reviewers would state if GTM  - even if their own point of views are not fully satisfied. Please bare with me, i can read some of the code meanwhile, but mostly not understand it in that depth that others can. GTM means to me: It solves real problems and don't introduce new _real_ problems in near future.  So please be explicit about real downsides of current changes - that would help me and others to learn and follow. :)

Joke aside - we should do us all a favour and make reviews and merges as soon as possible. That reduce the time needed to fix glitches if they happend (the time needed to go into the code is reduced dramatically if fresh in ones memory - or 'Man, who wrote this? Cool,  was i clever a few weeks ago!") and dont block other development (needed rebases etc) - Thanks for all the hard work.

tsujan:
@agaida Here, there are 5 things, *in my humble opinion*. I list them according to their relative importance to me:

(1) Does a PR solve a problems / fix a bug / add a feature? This can be tested by using it. There's no doubt.

(2) Does it create another problem? The code should be read/reviewed. If there's a programmatic problem, it can be pinpointed objectively but (4) should also be taken into account. Tests are also important.

(3) Is there an easier way of doing the same thing? This is partly objective and partly subjective. When 20 lines of code are replaced by 5 or when less methods are used without creating a problem, it's objective. When someone sees method X simpler than Y, it *may* be subjective.

(4) Do I like this solution? Is it according to may habits/coding style? Oh, this is totally subjective. Of course, I can say many things about why I like/dislike it but when the number of persons ≥ 2, they can't have the same likes/dislikes in all cases.

(5) Mistakes are not the end of the world. Sometimes -- as in the case of @PCMan's huge PR -- they could be found and fixed later.

I told the above to make it clear why sometimes I "easily" accept a PR/suggestion and sometimes seem uncompromising/stubborn in discussions (have no plan to change this).

Anyway, it's great to make development faster by merging PRs wisely.

agaida:
Hihihi - i'm not the kind of  'if it compile roll it out' - but there is a little truth in it. Only things that are available without to much efforts needed will be used by an broader base. So i'm all for smaller steps in development - maybe i'm biased as as former Arch AUR and now Sid user.  In general i prefer to release faster and more often in smaller steps.

PR #77, opened 2017-04-29
Ensure one item per file

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/482.

When the directory list job is finished on loading a folder, the list may contain more than one info per file because files may have been updated meanwhile (SFTP can be an example). In such cases, `filesChanged()` should be emitted instead of `filesAdded()`, otherwise more than one item will be shown for each file. In other words, infos should always be checked for being either added or updated.

Such cases were handled (and described as "very rare") by `libfm` but forgotten in the new `libfm-qt`.

### DIFF
--- a/src/core/folder.cpp
+++ b/src/core/folder.cpp
@@ -463,11 +463,29 @@ void Folder::onDirListFinished() {
     }
     dirInfo_ = job->dirInfo();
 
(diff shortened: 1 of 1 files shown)


agaida:
GTM

PR #78, opened 2017-04-30
Fix the broken filesystem status (free disk space) display.

PCMan:
Fix the issue reported here.
https://github.com/lxde/pcmanfm-qt/issues/494


agaida:
GTM

PR #80, opened 2017-05-01
Fix broken folder unmount message caused by incorrect FilePath & GFile* comparison.

PCMan:
Add operator== for comparing FilePath and GFile* using g_file_equal() gio API.
This fixes the incorrect GFile* pointer comparison that caused broken folder unmount event.

### DIFF
--- a/src/core/filepath.h
+++ b/src/core/filepath.h
@@ -130,11 +130,15 @@ public:
     }
 
     bool operator == (const FilePath& other) const {
-        if(gfile_ == other.gfile_) {
+        return operator==(other.gfile_.get());
+    }
+
+    bool operator == (GFile* other_gfile) const {
+        if(gfile_ == other_gfile) {
             return true;
         }
-        if(gfile_ && other.gfile_) {
-            return g_file_equal(gfile_.get(), other.gfile_.get());
+        if(gfile_ && other_gfile) {
+            return g_file_equal(gfile_.get(), other_gfile);
         }
         return false;
     }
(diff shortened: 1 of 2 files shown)


tsujan:
In case anyone wants to test this before merging, it's about unmounting a mounted folder while it's opened by pcmanfm-qt. Without this PR, the folder-view wasn't closed.

src/core/filepath.h
```diff
@@ -130,11 +130,15 @@ class LIBFM_QT_API FilePath {
     }
 
     bool operator == (const FilePath& other) const {
-        if(gfile_ == other.gfile_) {
+        return operator==(other.gfile_.get());
+    }
+
+    bool operator == (GFile* other_gfile) const {
+        if(gfile_ == other_gfile) {
```
palinek: I'm not sure about the compiler optimizations of such branching.... but logically this:
```c++
return gfile_ == other_file || (gfile_ && other_gfile && g_file_equal(gfile_.get(), other_gfile));
```
is the same as those 2 branches w/o the actual (explicit) branching.

palinek:
> In case anyone wants to test this before merging, it's about unmounting a mounted folder while it's opened by pcmanfm-qt. Without this PR, the folder-view wasn't closed.

Testing just with the `mount -o loop foo.iso bar`...and works as expected.

GTM

Note: sometimes there are 2 (same) devices added in the left pane upon the mount (only the first one is selectable).

tsujan:
> Note: sometimes there are 2 (same) devices added in the left pane upon the mount (only the first one is selectable).

Would you please open an issue? That couldn't be related to this PR, IMO.

agaida:
GTM

PR #84, opened 2017-05-07
Check for nullity of IconInfo

tsujan:
Because GIcon may be NULL.

### DIFF
--- a/src/filemenu_p.h
+++ b/src/filemenu_p.h
@@ -35,8 +35,10 @@ public:
         appInfo_{std::move(app)} {
         setToolTip(QString::fromUtf8(g_app_info_get_description(appInfo_.get())));
         GIcon* gicon = g_app_info_get_icon(appInfo_.get());
-        QIcon icon = Fm::IconInfo::fromGIcon(gicon)->qicon();
-        setIcon(icon);
+        const auto icnInfo = Fm::IconInfo::fromGIcon(gicon);
+        if(icnInfo) {
+            setIcon(icnInfo->qicon());
+        }
     }
 
     virtual ~AppInfoAction() {


src/filemenu_p.h
```diff
@@ -35,8 +35,10 @@ class AppInfoAction : public QAction {
         appInfo_{std::move(app)} {
         setToolTip(QString::fromUtf8(g_app_info_get_description(appInfo_.get())));
         GIcon* gicon = g_app_info_get_icon(appInfo_.get());
-        QIcon icon = Fm::IconInfo::fromGIcon(gicon)->qicon();
-        setIcon(icon);
+        const auto icnInfo = Fm::IconInfo::fromGIcon(gicon);
+        if(icnInfo) {
```
yan12125: I guess it should be ```if(icnInfo.get())```? It checks the content of the shared pointer rather than the pointer itself.

tsujan: `IconInfo` is `std::shared_ptr<const IconInfo>{}` when `GIcon` is `NULL` (See `IconInfo::fromGIcon()` in `iconinfo.cpp`)

tsujan:
@yan12125 
Actually, I think this should never happen with zip or pdf, as @ferronrsmith reported at https://github.com/lxde/lxqt/issues/1293. However, GIO doc says `g_app_info_get_icon()` can return `NULL`.

src/filemenu_p.h
yan12125: Ah, I don't know ```shared_ptr``` has ```operator bool```. C++11 is too hard for me :/

tsujan:
Apparently, https://github.com/lxde/lxqt/issues/1293 was fixed by this.

agaida:
GTM

PR #85, opened 2017-05-07
Fix single items when seaching

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/505 by not updating infos but adding all of them in the case of searching when directory list job is finished.

Please note that seaching still depends on libfm and this might need change after @PCMan rewrites the search.


agaida:
GTM

PR #86, opened 2017-05-08
Added a missing change signal

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/506 and similar issues.

When a file is added (pasted) to an already loaded folder, it should be queued for changes if the monitor event is `G_FILE_MONITOR_EVENT_ATTRIBUTE_CHANGED` or `G_FILE_MONITOR_EVENT_CHANGED` and even if it isn't added to the files list yet. Otherwise, weird things like incomplete or missing thumbnails or incorrect size infos might happen on pasting files.


agaida:
GTM

PR #88, opened 2017-05-21
Grey out files that have been Ctrl-X'ed

vc-01:
Add visual feedback for files that have been 'cut' to clipboard.
---
Hello. Referring to [issue #503](https://github.com/lxde/pcmanfm-qt/issues/503).

I was looking on that and made some working draft. But before more work to be put into this I'd like to know if you think this is good way at all and if you are interested in it.

I put more details in source comments temporarily.

I can answer design decisions but mostly they were influenced by speed requirements (to the point I understood them as such).

This solution is based mostly on weak pointers serving as "observers" to current cut selection, and if present - adjusting icons transparency (icons/thumbnails view only now)

In this approach, two lines of code are necessary to add to *pcmanfm-qt* because actual CTRL-X key shortcut is defined and processed there - test it with mouse context menu -> "cut" for now.

If you see something very mysterious there, that's possible, this is my first meet with all this, smart pointers etc.

This is not a ready to commit PR yet but a designing proposal.

### DIFF
--- a/src/core/dirlistjob.cpp
+++ b/src/core/dirlistjob.cpp
@@ -6,7 +6,8 @@
 
 namespace Fm {
 
-DirListJob::DirListJob(const FilePath& path, Flags _flags): dir_path{path}, flags{_flags} {
+DirListJob::DirListJob(const FilePath& path, Flags _flags, const std::shared_ptr<const HashSet>& cutFilesHashSet):
+    dir_path{path}, flags{_flags}, cutFilesHashSet_{cutFilesHashSet} {
 }
 
 void DirListJob::exec() {
@@ -102,6 +103,12 @@ _retry:
                 if(emit_files_found) {
                     // Q_EMIT filesFound();
                 }
+
+                if(cutFilesHashSet_
+                        && cutFilesHashSet_->count(fileInfo->path().hash()) > 0) {
+                    fileInfo->bindCutFiles(cutFilesHashSet_);
+                }
+
                 foundFiles.push_back(std::move(fileInfo));
             }
             else {
--- a/src/core/fileinfojob.cpp
+++ b/src/core/fileinfojob.cpp
@@ -3,10 +3,11 @@
 
 namespace Fm {
 
-FileInfoJob::FileInfoJob(FilePathList paths, FilePath commonDirPath):
+FileInfoJob::FileInfoJob(FilePathList paths, FilePath commonDirPath, const std::shared_ptr<const HashSet>& cutFilesHashSet):
     Job(),
     paths_{std::move(paths)},
-    commonDirPath_{std::move(commonDirPath)} {
+    commonDirPath_{std::move(commonDirPath)},
+    cutFilesHashSet_{cutFilesHashSet} {
 }
 
 void FileInfoJob::exec() {
@@ -23,9 +24,17 @@ void FileInfoJob::exec() {
 
             // Reuse the same dirPath object when the path remains the same (optimize for files in the same dir)
             auto dirPath = commonDirPath_.isValid() ? commonDirPath_ : path.parent();
-            std::shared_ptr<const FileInfo> fileInfo = std::make_shared<FileInfo>(inf, dirPath);
-            results_.push_back(fileInfo);
-            Q_EMIT gotInfo(path, fileInfo);
+            FileInfo fileInfo(inf, dirPath);
+
+            if(cutFilesHashSet_
+                    && cutFilesHashSet_->count(fileInfo.path().hash())) {
+                fileInfo.bindCutFiles(cutFilesHashSet_);
+            }
+
+            auto fileInfoPtr = std::make_shared<const FileInfo>(fileInfo);
+
+            results_.push_back(fileInfoPtr);
+            Q_EMIT gotInfo(path, fileInfoPtr);
         }
     }
 }
--- a/src/core/folder.h
+++ b/src/core/folder.h
@@ -80,6 +80,12 @@ public:
 
     const std::shared_ptr<const FileInfo> &info() const;
 
+    bool hadCutFilesUnset();
+
+    bool hasCutFiles();
+
+    void setCutFiles(const std::shared_ptr<const HashSet>& cutFilesHashSet);
+
     void forEachFile(std::function<void (const std::shared_ptr<const FileInfo>&)> func) const {
         std::lock_guard<std::mutex> lock{mutex_};
         for(auto it = files_.begin(); it != files_.end(); ++it) {
@@ -179,6 +185,9 @@ private:
     bool defer_content_test : 1;
 
     static std::unordered_map<FilePath, std::weak_ptr<Folder>, FilePathHash> cache_;
+    static FilePath cutFilesDirPath_;
+    static FilePath lastCutFilesDirPath_;
+    static std::shared_ptr<const HashSet> cutFilesHashSet_;
     static std::mutex mutex_;
 };
 
--- a/src/foldermodel.cpp
+++ b/src/foldermodel.cpp
@@ -156,6 +156,21 @@ void FolderModel::insertFiles(int row, const Fm::FileInfoList& files) {
     endInsertRows();
 }
 
+void FolderModel::setCutFiles(const QItemSelection& selection) {
+    if(folder_) {
+        if(!selection.isEmpty()) {
+            auto cutFilesHashSet = std::make_shared<HashSet>();
+            folder_->setCutFiles(cutFilesHashSet);
+            for(const auto& index : selection.indexes()) {
+                auto item = itemFromIndex(index);
+                item->bindCutFiles(cutFilesHashSet);
+                cutFilesHashSet->insert(item->info->path().hash());
+            }
+        }
+        Q_EMIT dataChanged(index(0, 0), index(rowCount() - 1, 0));
+    }
+}
+
 void FolderModel::removeAll() {
     if(items.empty()) {
         return;
@@ -195,6 +210,11 @@ QVariant FolderModel::data(const QModelIndex& index, int role/* = Qt::DisplayRol
     FolderModelItem* item = itemFromIndex(index);
     auto info = item->info;
 
+    bool isCut = false;
+    if(folder_ && Q_UNLIKELY(folder_->hasCutFiles())) {
+        isCut = item->isCut();
+    }
+
     switch(role) {
     case Qt::ToolTipRole:
         return QVariant(item->displayName());
@@ -215,7 +235,7 @@ QVariant FolderModel::data(const QModelIndex& index, int role/* = Qt::DisplayRol
     }
     case Qt::DecorationRole: {
         if(index.column() == 0) {
-            return QVariant(item->icon());
+            return QVariant(item->icon(isCut));
         }
         break;
     }
@@ -229,6 +249,8 @@ QVariant FolderModel::data(const QModelIndex& index, int role/* = Qt::DisplayRol
         return QVariant::fromValue(info);
     case FileIsDirRole:
         return QVariant(info->isDir());
+    case FileIsCutRole:
+        return isCut;
     }
     return QVariant();
 }
@@ -480,8 +502,9 @@ void FolderModel::onThumbnailLoaded(const std::shared_ptr<const Fm::FileInfo>& f
         FolderModelItem& item = *it;
         QModelIndex index = createIndex(row, 0, (void*)&item);
         // store the image in the folder model item.
-        FolderModelItem::Thumbnail* thumbnail = item.findThumbnail(size);
+        FolderModelItem::Thumbnail* thumbnail = item.findThumbnail(size, false);
         thumbnail->image = image;
+        thumbnail->transparent = false;
         // qDebug("thumbnail loaded for: %s, size: %d", item.displayName.toUtf8().constData(), size);
         if(image.isNull()) {
             thumbnail->status = FolderModelItem::ThumbnailFailed;
@@ -501,7 +524,7 @@ void FolderModel::onThumbnailLoaded(const std::shared_ptr<const Fm::FileInfo>& f
 QImage FolderModel::thumbnailFromIndex(const QModelIndex& index, int size) {
     FolderModelItem* item = itemFromIndex(index);
     if(item) {
-        FolderModelItem::Thumbnail* thumbnail = item->findThumbnail(size);
+        FolderModelItem::Thumbnail* thumbnail = item->findThumbnail(size, item->isCut());
         // qDebug("FolderModel::thumbnailFromIndex: %d, %s", thumbnail->status, item->displayName.toUtf8().data());
         switch(thumbnail->status) {
         case FolderModelItem::ThumbnailNotChecked: {
--- a/src/core/dirlistjob.h
+++ b/src/core/dirlistjob.h
@@ -19,7 +19,7 @@ public:
         DETAILED = 1 << 1
     };
 
-    explicit DirListJob(const FilePath& path, Flags flags);
+    explicit DirListJob(const FilePath& path, Flags flags, const std::shared_ptr<const HashSet>& cutFilesHashSet = nullptr);
 
     FileInfoList& files() {
         return files_;
@@ -54,6 +54,7 @@ private:
     Flags flags;
     std::shared_ptr<const FileInfo> dir_fi;
     FileInfoList files_;
+    const std::shared_ptr<const HashSet> cutFilesHashSet_;
     bool emit_files_found;
     // guint delay_add_files_handler;
     // GSList* files_to_add;
--- a/src/core/fileinfo.cpp
+++ b/src/core/fileinfo.cpp
@@ -287,6 +287,10 @@ _file_is_symlink:
 #endif
 }
 
+void FileInfo::bindCutFiles(const std::shared_ptr<const HashSet>& cutFilesHashSet) {
+    cutFilesHashSet_ = cutFilesHashSet;
+}
+
 bool FileInfo::canThumbnail() const {
     /* We cannot use S_ISREG here as this exclude all symlinks */
     if(size_ == 0 ||  /* don't generate thumbnails for empty files */
--- a/src/core/fileinfo.h
+++ b/src/core/fileinfo.h
@@ -31,6 +31,7 @@
 #include <fcntl.h>
 
 #include <vector>
+#include <set>
 #include <utility>
 #include <string>
 #include <forward_list>
@@ -42,6 +43,9 @@
 
 
 namespace Fm {
+
+class FileInfoList;
+typedef std::set<unsigned int> HashSet;
 
 class LIBFM_QT_API FileInfo {
 public:
@@ -158,6 +162,10 @@ public:
         return dirPath_ ? dirPath_.isNative() : path().isNative();
     }
 
+    bool isCut() const {
+        return !cutFilesHashSet_.expired();
+    }
+
     mode_t mode() const {
         return mode_;
     }
@@ -187,6 +195,8 @@ public:
     }
 
     void setFromGFileInfo(const GFileInfoPtr& inf, const FilePath& parentDirPath);
+
+    void bindCutFiles(const std::shared_ptr<const HashSet>& cutFilesHashSet);
 
     const std::forward_list<std::shared_ptr<const IconInfo>>& emblems() const {
         return emblems_;
@@ -225,6 +235,7 @@ private:
     bool isHiddenChangeable_ : 1; /* TRUE if hidden can be changed */
     bool isReadOnly_ : 1; /* TRUE if host FS is R/O */
 
+    std::weak_ptr<const HashSet> cutFilesHashSet_;
     // std::vector<std::tuple<int, void*, void(void*)>> extraData_;
 };
 
--- a/src/core/fileinfojob.h
+++ b/src/core/fileinfojob.h
@@ -13,7 +13,7 @@ class LIBFM_QT_API FileInfoJob : public Job {
     Q_OBJECT
 public:
 
-    explicit FileInfoJob(FilePathList paths, FilePath commonDirPath = FilePath());
+    explicit FileInfoJob(FilePathList paths, FilePath commonDirPath = FilePath(), const std::shared_ptr<const HashSet>& cutFilesHashSet = nullptr);
 
     const FilePathList& paths() const {
         return paths_;
@@ -33,6 +33,7 @@ private:
     FilePathList paths_;
     FileInfoList results_;
     FilePath commonDirPath_;
+    const std::shared_ptr<const HashSet> cutFilesHashSet_;
 };
 
 } // namespace Fm
--- a/src/core/folder.cpp
+++ b/src/core/folder.cpp
@@ -34,6 +34,9 @@
 namespace Fm {
 
 std::unordered_map<FilePath, std::weak_ptr<Folder>, FilePathHash> Folder::cache_;
+FilePath Folder::cutFilesDirPath_;
+FilePath Folder::lastCutFilesDirPath_;
+std::shared_ptr<const HashSet> Folder::cutFilesHashSet_;
 std::mutex Folder::mutex_;
 
 Folder::Folder():
@@ -251,7 +254,8 @@ void Folder::processPendingChanges() {
         FilePathList paths;
         paths.insert(paths.end(), paths_to_add.cbegin(), paths_to_add.cend());
         paths.insert(paths.end(), paths_to_update.cbegin(), paths_to_update.cend());
-        info_job = new FileInfoJob{paths, dirPath_};
+        info_job = new FileInfoJob{paths, dirPath_,
+                hasCutFiles() ? cutFilesHashSet_ : nullptr};
         paths_to_update.clear();
         paths_to_add.clear();
     }
@@ -451,6 +455,30 @@ void Folder::onFileChangeEvents(GFileMonitor* /*monitor*/, GFile* gf, GFile* /*o
         }
         queueUpdate();
     }
+}
+
+// checks whether there were cut files here
+// and if there were, invalidates this last cut path
+bool Folder::hadCutFilesUnset() {
+    if(lastCutFilesDirPath_ == dirPath_) {
+        lastCutFilesDirPath_ = FilePath();
+        return true;
+    }
+    return false;
+}
+
+bool Folder::hasCutFiles() {
+    return cutFilesHashSet_
+            && !cutFilesHashSet_->empty()
+            && cutFilesDirPath_ == dirPath_;
+}
+
+void Folder::setCutFiles(const std::shared_ptr<const HashSet>& cutFilesHashSet) {
+    if(cutFilesHashSet_ && !cutFilesHashSet_->empty()) {
+        lastCutFilesDirPath_ = cutFilesDirPath_;
+    }
+    cutFilesDirPath_ = dirPath_;
+    cutFilesHashSet_ = cutFilesHashSet;
 }
 
 void Folder::onDirListFinished() {
@@ -661,7 +689,8 @@ void Folder::reload() {
     /* run a new dir listing job */
     // FIXME:
     // defer_content_test = fm_config->defer_content_test;
-    dirlist_job = new DirListJob(dirPath_, defer_content_test ? DirListJob::FAST : DirListJob::DETAILED);
+    dirlist_job = new DirListJob(dirPath_, defer_content_test ? DirListJob::FAST : DirListJob::DETAILED,
+                                 hasCutFiles() ? cutFilesHashSet_ : nullptr);
     dirlist_job->setAutoDelete(true);
     connect(dirlist_job, &DirListJob::error, this, &Folder::error, Qt::BlockingQueuedConnection);
     connect(dirlist_job, &DirListJob::finished, this, &Folder::onDirListFinished, Qt::BlockingQueuedConnection);
--- a/src/core/iconinfo.cpp
+++ b/src/core/iconinfo.cpp
@@ -57,16 +57,28 @@ void IconInfo::updateQIcons() {
     }
 }
 
-QIcon IconInfo::qicon() const {
-    if(Q_UNLIKELY(qicon_.isNull() && gicon_)) {
-        if(!G_IS_FILE_ICON(gicon_.get())) {
-            qicon_ = QIcon(new IconEngine{shared_from_this()});
-        }
-        else {
-            qicon_ = internalQicon_;
+QIcon IconInfo::qicon(const bool& transparent) const {
+    if(Q_LIKELY(!transparent)) {
+        if(Q_UNLIKELY(qicon_.isNull() && gicon_)) {
+            if(!G_IS_FILE_ICON(gicon_.get())) {
+                qicon_ = QIcon(new IconEngine{shared_from_this()});
+            }
+            else {
+                qicon_ = internalQicon_;
+            }
         }
     }
-    return qicon_;
+    else { // transparent == true
+        if(Q_UNLIKELY(qiconTransparent_.isNull() && gicon_)) {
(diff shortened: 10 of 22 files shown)


tsujan:
It's great that you work on it! I'll read your code as soos as I find some free time. Thanks!

vc-01:
I use pcmanfm-qt, I do it for myself and everyone else. It's hobby. But also please be patient when waiting for *my* responses and changes, could take some time too.

I'm not *that* content with this work but I couldn't find better solution actually.
Maybe I summarize it..
1. Cutting files, pasting files (pasting not implemented yet but easy) - very fast, good
2. Making another cut selection when one was already present - quite ok, fast (invalidating weak pointers)
3. Reloading folder with cut files - more problematic, looking cut files through std::set of file path hashes for every file (and restricted to max. 450)

More detailed discussion when ready..
(and open to any hints and critics)

src/foldermodel.cpp
```diff
@@ -217,7 +226,14 @@ QVariant FolderModel::data(const QModelIndex& index, int role/* = Qt::DisplayRol
     }
     case FileInfoRole:
         return QVariant::fromValue(info);
+    case FileIsCutRole:
+        bool isCut = false;
+        if (folder_ && folder_->hasCutFiles()) { // most of the time false
+            isCut = item->isCut() || item->info->isCut(); // then dynamic evaluation
```
vc-01: Why is spread twice (in FileInfo and in FolderModelItem)? Because of const-ness of FileInfo. Could be casted away but should be?

vc-01:
Anyway, I finished it the way it was started.

I forced pushed it as a full new patch but nothing has changed from before (if you started to review already), things were only added.

If settled and approved, I'll make another small PR in *pcmanfm-qt* repo to finalize it and make actual CTRL-X to work.

(Surely as mentioned before, this was not that straightforward as it may have seemed)

tsujan:
> If settled and approved, I'll make another small PR in pcmanfm-qt repo to finalize it

You could do so now for people to be able to test your idea. In your pcmanfm-qt PR's comment, you could add "follows https://github.com/lxde/libfm-qt/pull/88" or "based on ...".

tsujan:
BTW, sorry that I still haven't had time to read this! It needs to be read carefully because you've changed sensitive parts of the code. I'll find some time soon.

tsujan:
Before going to bed (it's past midnight here), I took a look at the code and hoped that it could be made simpler. However,  graying out cut files isn't as easy as it might seem and you've done a great job here.

Including `folderview.h` in `filemenu.h` may not be a good option but I don't know of a better way either. in my (updated) patch for inline renaming (not PR'ed yet), I needed to include `QAbstractItemView` in `filemenu.cpp` to make it aware of views. So, perhaps this is inevitable -- I can't say yet.

Anyhow, there's a serious issue, IMHO. In your code, I couldn't find a mechanism for removing the grayed-out state when another file is copied. So, I compiled your branch and saw the problem. Since the grayed-out state is for the user to know which file will be cut *and pasted* the next time, if that state isn't removed after copying another file, it'll be misleading. I don't mean that the implementation of this property is easy but it's necessary.

Please keep up the good work! Thanks!

tsujan:
Oh! More importantly, we need a mechanism to watch the clipboard! Once something is copied/cut *anywhere*, the grayed-out state should be removed. Whether we need another approach or this one could be modified, I don't know yet.

tsujan:
Your last commit isn't enough. The grayed-out state should be removed on copying anywhere -- copying a text inside a text editor, for example.

vc-01:
Yes, now I understand the problem. I can only hope I find a solution which wouldn't touch the core of the lib too much. Until now it was mostly about adding things. But this needs to be figured out, UI cannot mislead about its states, that would be a bug anyway.

(PR in *pcmanfm-qt* has to wait, this is not test ready yet again)

tsujan:
> PR in pcmanfm-qt has to wait,

I agree. I hope you'll find a solution for copy/cut in other apps.

palinek:
My poor 2 cents.... I think we should listen to to `QClipboard` changes and do something like in the `Fm::pasteFilesFromClipboard' & co. https://github.com/lxde/libfm-qt/blob/master/src/utilities.cpp#L69 . (maybe keep a global list of `Fm::FolderModel`s to set the `isCut` flag for items..)

vc-01:
@palinek 

> I think we should listen to QClipboard changes ...

Yes, this is the way I think, same approach is implemented in *Dolphin* file manager. I think I can come up with something but need some time.

> maybe keep a global list of Fm::FolderModels to set the isCut flag for items

To avoid adding *isCut* state to `FileInfo` and then also to `DirListJob`? But `FolderModelItem`s are re-created every chdir (https://github.com/lxde/libfm-qt/blob/master/src/foldermodel.cpp#L80). Maybe I didn't get your point, then need more explanation here.

tsujan:
> To avoid adding isCut state to FileInfo and then also to DirListJob?

I don't think @palinek meant that. Keeping track of files that are cut is at the core of your approach and since the directory may change, it seems necessary.

We all agree that QClipboard should be watched.

tsujan:
@vc-01 I'm still suspicious about graying out files on such a deep level as that of `FileInfo` because we can't cut files from different directories but please complete your method! Even if it's too generalized for graying out, its structure might be used for other purposes in future.

vc-01:
Two small bits to review first (bug-fixes).

Then the main patch, it's done with the least possible changes to the lib. In the next iteration, maybe I'd like a new global Clipboard (singleton) class which will wrap utilities' clipboard functions and save parsed data. But since this touches the core of the lib a lot, I left it to some subsequent PR.

Anyway, cut files are processed differently (aka more locally) for in-app made clipboard changes - same approach as already presented was kept.

I kept it, because I somehow believe in how was it done, because dealing directly with selections is fast. More general approach of listening to clipboard, signalling changes, and processing its data by saving and refreshing found folder items, may be cleaner design but slower (speed optimized code is always less readable and understandable).

Some slow method `FolderModel::findItemByPath` is used (but only for externally set cut files), I can provide refinement of this method by adding unordered map of `FilePath`s -> `FolderModelItem`s, if you want. But this part of code won't be run very often.

Last thing, when *pcmanfm-qt* is started it doesn't process clipboard data (so doesn't possibly show cut files) because according to QClipboard documentation `it is recommended that the contents of the clipboard are stored or retrieved in direct response to user-input events`. *Dolphin* does this is on startup, but here I'm not sure what is correct.

(tested with *Dolphin* file manager)

(I can do commits squashing for cleaner history at the end before merge)

tsujan:
Requested @PCMan's review.

src/core/fileinfojob.cpp
```diff
@@ -23,9 +31,15 @@ void FileInfoJob::exec() {
...
             auto dirPath = commonDirPath_.isValid() ? commonDirPath_ : path.parent();
-            std::shared_ptr<const FileInfo> fileInfo = std::make_shared<FileInfo>(inf, dirPath);
-            results_.push_back(fileInfo);
-            Q_EMIT gotInfo(path, fileInfo);
+            auto fileInfo = std::make_shared<FileInfo>(inf, dirPath);
+
+            if(hasCutFiles && cutFilesHashSet_->count(fileInfo->path().hash()))
+                fileInfo->bindCutFiles(cutFilesHashSet_);
+
+            auto fileInfoConst = std::const_pointer_cast<const FileInfo>(fileInfo);
```
vc-01: I see now, maybe I should have created `shared_ptr` only here, then `const_pointer_cast` can be ommited - _Can fix.._

src/core/fileinfojob.cpp
```diff
@@ -23,9 +31,15 @@ void FileInfoJob::exec() {
...
+
+            results_.push_back(fileInfoConst);
+            Q_EMIT gotInfo(path, fileInfoConst);
```
vc-01: Just to point out. Signal `gotInfo` is never used in the project.

vc-01:
I'm thinking about what is best for this pull request now, after this new merge conflict appeared.

History of this branch is getting harder to follow and it's not merged yet, so now I think it's good time to tidy up this - do rebase and squash, which will though invalidate comment history.

I will still keep two bugfixes made as separate commits because they are separate things, only they were found during the development of this feature.

Stop me if not to do this or if it's better to close this pull request, do rebase + squash, and re-open it as a new one (still learning the github and opensource way).

tsujan:
@vc-01 
I've had PRs which should have been rebased a few times, were marginalized by more important changes *many* times, or were removed and remade by me after several months. Such things often happen because of jumps in development, priorities or lack of manpower but they usually aren't about the PRs in question. So, it's up to the contributor to keep his/her idea up-to-date. The branch history isn't important -- if the code is clean, there will no need to a specific branch history.

tsujan:
I'll try to focus on this starting from tomorrow if no important bug is found.

tsujan:
@vc-01 I was impressed by seeing that if I cut some file in Dolphin, it's grayed out in pcmanfm-qt too and conversely but there's a (small) problem:

When a cut file is grayed out in pcmanfm-qt, if something is copied elsewhere, *the grayed-out state won't be removed until the cursor goes over the view*. This doesn't happen with Dolphin and is an issue. Could you fix it, please?

tsujan:
The problem only happens when desktop is shown (in the desktop mode).

src/core/folder.h
```diff
@@ -179,6 +183,8 @@ private Q_SLOTS:
     bool defer_content_test : 1;
 
     static std::unordered_map<FilePath, std::weak_ptr<Folder>, FilePathHash> cache_;
+    static FilePath cutFilesDirPath_;
+    static std::shared_ptr<const HashSet> cutFilesHashSet_;
```
tsujan: I think `cutFilesHashSet_` shouldn't be static, otherwise it'll cause the problem mentioned at https://github.com/lxde/libfm-qt/pull/88#issuecomment-319429259.

Moreover, as far as I see. `cutFilesDirPath_` is redundant and `setCutFiles()` doesn't need a `dirPath` argument because it should always be `dirPath_`.

I made the above changes (offline) and https://github.com/lxde/libfm-qt/pull/88#issuecomment-319429259 was fixed. Am I missing something here?

tsujan: Actually, I'm also suspicious about the static-ness of `cache_`, which may explain https://github.com/lxde/pcmanfm-qt/issues/508. I should look into that... but it isn't related to this PR.

vc-01: When cut files are set in some external application, it can be a random folder - not what is current folder in _pcmanfm-qt_.

Hence cut files are static (and "shared" for all folder instances). When this cut files' folder can be random, also `cutFilesDirPath_` is needed.

Originally it went static because I thought of possibility when cut files could be selected across more than one folder, but this was misunderstood so this design reasoning  went _old_ in some way.

Better solution might exists but I couldn't find a way how to setup locally cut files for a folder which was not loaded (and cached) yet.

tsujan: OK, I found what I was missing. `cutFilesHashSet_` should be static, of course, and `cutFilesDirPath_` is needed. The problem is that if another directory (like desktop) is loaded, `FolderModel::setCutFiles(const QItemSelection& selection)` won't work correctly because if that other directory calls it first, `refreshAll` will be made `false` for the folder in question.

Please tell me when you have time for a discussion! And sorry for my misunderstanding, which was caused by trying to find an easy solution to a problem!

tsujan: Oh, I didn't see your comment when I was typing my last comment! Yes, you're right about the static-ness; that was my mistake.

Today, I got familiar with your code . Tomorrow, I'll think about a solution and will share my findings. Get ready for a long discussion ;)

vc-01: For the _desktop-icons-not-showing-cut-status_ bug, I confirm it. I don't use desktop icons, so it didn't come up to me to test this use case.

tsujan: With the help of `FolderModel::setCutFiles` (both of them), I succeeded in reproducing it without desktop too. I think I've found a solution but let's continue our discussion tomorrow -- being too busy, I didn't have enough focus today.

vc-01: > I think I've found a solution...

It's always enough to just report a bug. I'm here to support (especially) my code.

tsujan: > I'm here to support (especially) my code.

My solution will be just a suggestion that you could disagree about; I won't touch your code. And I also support it.

tsujan:
My "solution" (I'll tell you why I put it inside quotes):

(1) These methods are removed (they're reduntant in what follows):
```cpp
isCurrentPidClipboardData(const QMimeData& data);
FolderView::onCutTriggered();
FolderView::onCopyTriggered();
FolderView::onPasteTriggered();
ProxyFolderModel::setCutFiles(const QItemSelection& selection)
FolderModel::setCutFiles(const QItemSelection& selection)
```

(2) The static variable `FilePath oldCutFilesDirPath_` is added to `folder.h` and is set by `Folder::setCutFiles` as:
```cpp
void Folder::setCutFiles(const std::shared_ptr<const HashSet>& cutFilesHashSet) {
    oldCutFilesDirPath_ = cutFilesDirPath_;
    cutFilesDirPath_ = dirPath_;
    cutFilesHashSet_ = cutFilesHashSet;
}
```
Of course, the argument `dirPath` is removed from `Folder::setCutFiles`. This simple method checks if the folder *had* cut files recently:
```cpp
bool Folder::hadCutFiles() const {
    return oldCutFilesDirPath_ == dirPath_;
}
```

(3) `FolderView::onClipboardDataChange` is changed as:
```cpp
void FolderView::onClipboardDataChange() {
    if(model_) {
        const QClipboard* clipboard = QApplication::clipboard();
        const QMimeData* data = clipboard->mimeData();
        Fm::FilePathList paths;
        bool isCutSelection;
        std::tie(paths, isCutSelection) = Fm::parseClipboardData(*data);
        if(isCutSelection) {
            model_->setCutFiles(paths);
        }
        else if(!isCutSelection && folder()->hasCutFiles()) {
            folder()->setCutFiles(std::make_shared<HashSet>());
            model_->setCutFiles(paths);
        }
    }
}
```

(4) At last, `FolderModel::setCutFiles` is changed as:
```cpp
void FolderModel::setCutFiles(const Fm::FilePathList& cutFilePaths) {
    if(folder_) {
        bool refreshAll(folder_->hasCutFiles() || folder_->hadCutFiles());
        auto cutDirPath = cutFilePaths.size() > 0 ? cutFilePaths[0].parent(): FilePath();
        if(cutDirPath == folder_->path()) {
            auto cutFilesHashSet = std::make_shared<HashSet>();
            folder_->setCutFiles(cutFilesHashSet);
            int row;
            for(const auto& path : cutFilePaths) {
                QList<FolderModelItem>::iterator it = findItemByPath(path, &row);
                if(it != items.end()) {
                    FolderModelItem& item = *it;

                    item.bindCutFiles(cutFilesHashSet);
                    if(!refreshAll) {
                        Q_EMIT dataChanged(index(row, 0), index(row, 0));
                    }
                }
                cutFilesHashSet->insert(path.hash());
            }
        }

        if(refreshAll) {
            Q_EMIT dataChanged(index(0, 0), index(rowCount() - 1, 0));
        }
    }
}
```

In this way, apparently all issues are fixed. However:

What worries me is the high CPU usage and also the long delay when the number of files is more than 1000. It makes no difference whether the above "solution" is used or not. That's while Dolphin grays out 3000 files almost instantly.

I still don't know in which function the CPU time is taken mostly but it's definitely not `Q_EMIT dataChanged`. This should be fixed, IMHO.

tsujan:
Oh, I forgot to tell that with the above solution, there's no need to a separate PR for pcmanfm-qt. Everything is done in libfm-qt. The reason is the new `FolderView::onClipboardDataChange()`.

vc-01:
### (1) Remove methods
They were created historically when processing cut files selection was not yet done from clipboard event. I left them alive also after that because they use _Selections_ which are more "prepared" for the needed processing.

Using
```c++
auto item = itemFromIndex(index);
```
instead of
```c++
QList<FolderModelItem>::iterator it = findItemByPath(path, &row);
```

and
```c++
Q_EMIT dataChanged(index, index);
```
instead of
```c++
Q_EMIT dataChanged(index(row, 0), index(row, 0));
```

Maybe creating a new index with `index()` is not an issue, but iterating through a `FilePathList` for every cut file it **is**.

But now, after re-thought, it is more clear to me, that better place to look after, is how to optimize `findItemByPath()` and then the whole bunch of code (which is also less understandable) can be removed.

I **will** remove those methods (and also close the related PR in _pcmanfm-qt_), finding more optimized way of `findItemByPath()` will be some next step but also change ready for a separate PR.

### (2-1) Remove `dirPath` argument from `Folder::setCutFiles()` 

This cannot be removed, because when you select cut files in external application in some random directory, you have to set exactly to which directory the selection belongs. Actually, with this approach of file paths hash-sets you don't, but then you'll check cut files' hashes in every directory change. This is some kind of optimization.

### (2-2, 3) Add `Folder::hadCutFiles()`

Works for the reported bug, but introduces another one: Try to set cut files in Dolphin in directory which is not opened in pcmanfm-qt, then go to that directory - files won't be grayed out.

But that's ok, I will fix this bug some way.

### (4) Modify `FolderModel::setCutFiles`
Setting _hashSet_ cannot be inside `if(cutDirPath == folder_->path()) { .. }` because of reasons mentioned before (externally made random directory cut files selection).

### (5) Slowness
Has to be analyzed (and fixed optionally :) ). Marking as TODO. What I'm little afraid of, this would require touches to some core behavior (we'll see).

tsujan:
> (2-1) Remove dirPath argument from Folder::setCutFiles()

> This cannot be removed, because when you select cut files in external application in some random directory, you have to set exactly to which directory the selection belongs.

I'll think about it. That removal wasn't important.

> (2-2, 3) Add Folder::hadCutFiles()

> Works for the reported bug, but introduces another one: Try to set cut files in Dolphin 

Will test and think about this too.

> (4) Modify FolderModel::setCutFiles

> Setting hashSet cannot be inside if(cutDirPath == folder_->path()) { .. } because of reasons mentioned before (externally made random directory cut files selection).

Didn't understand this one but it seems to be about setting cut files in another file manager again.

To simplify my share of work, I remove the "another-file-manager" parameter from my calculations for now. Apart from that, the above method works. Af for slowness with > 1000 files, the culprit should be `findItemByPath()`, as you mentioned.

BTW, this is the standard method of setting painter opacity, as far as I know:
```cpp
void IconEngine::paint(QPainter* painter, const QRect& rect, QIcon::Mode mode, QIcon::State state) {
    auto info = info_.lock();
    if(info) {
        if(transparent_) {
            painter->save();
            painter->setOpacity(0.45);
        }
        info->internalQicon().paint(painter, rect, Qt::AlignCenter, mode, state);
        if(transparent_) {
            painter->restore();
        }
    }
}
```
Please tell me if I'm missing something here!

tsujan:
@vc-01 In another try, I only used selections and avoided `findItemByPath()` completely. The gray-out process for files > 2000 became tangibly faster but not fast enough. Dolphin is still the winner.

tsujan:
Interesting finding: this time, only the first cut took the CPU time. I chose 3000 files from the mozilla cache folder and cut them. Then I cut a file on desktop and cut those 3000 files again. The second (third,...) time, the files were grayed out *immediately*.

For the record, the new functions are:
```cpp
void FolderView::onClipboardDataChange() {
    if(model_) {
        const QClipboard* clipboard = QApplication::clipboard();
        const QMimeData* data = clipboard->mimeData();
        Fm::FilePathList paths;
        bool isCutSelection;
        std::tie(paths, isCutSelection) = Fm::parseClipboardData(*data);
        if(isCutSelection) {
            auto cutDirPath = paths.size() > 0 ? paths[0].parent(): FilePath();
            model_->setCutFiles(folder()->path() == cutDirPath ? selectionModel()->selection()
                                                               : QItemSelection());
        }
        else if(folder()->hasCutFiles()) {
            folder()->setCutFiles(std::make_shared<HashSet>());
            model_->setCutFiles(QItemSelection());
        }
    }
}
```
and
```cpp
void FolderModel::setCutFiles(const QItemSelection& selection) {
    if(folder_) {
        bool refreshAll(folder_->hasCutFiles() || folder_->hadCutFiles());

        if(!selection.isEmpty()) {
            auto cutFilesHashSet = std::make_shared<HashSet>();
            folder_->setCutFiles(cutFilesHashSet);
            for(const auto& index : selection.indexes()) {
                auto item = itemFromIndex(index);
                item->bindCutFiles(cutFilesHashSet);
                if(!refreshAll) Q_EMIT dataChanged(index, index);
                cutFilesHashSet->insert(item->info->path().hash());
            }
        }

        if(refreshAll) {
            Q_EMIT dataChanged(index(0, 0), index(rowCount() - 1, 0));
        }
    }
}
```
They are unpolished, of course.

tsujan:
CPU usage problem solved :)

(1) Use selection and avoid `findItemByPath()` completely (as explained above);
(2) Don't use `Q_EMIT dataChanged(index, index);`. Instead, use
```cpp
Q_EMIT dataChanged(index(0, 0), index(rowCount() - 1, 0));
```
in all cases.

This time, I dared to cut 6500 files and they were grayed out immediately! Now, pcmafm-qt is the winner ;)

The function is simplified as:
```cpp
void FolderModel::setCutFiles(const QItemSelection& selection) {
    if(folder_) {
        if(!selection.isEmpty()) {
            auto cutFilesHashSet = std::make_shared<HashSet>();
            folder_->setCutFiles(cutFilesHashSet);
            for(const auto& index : selection.indexes()) {
                auto item = itemFromIndex(index);
                item->bindCutFiles(cutFilesHashSet);
                cutFilesHashSet->insert(item->info->path().hash());
            }
        }
        Q_EMIT dataChanged(index(0, 0), index(rowCount() - 1, 0));
    }
}
```
I knew your approach would work :)

tsujan:
And the same with ~7000 files :) Sorry, I don't have a folder with more files ;)

tsujan:
As for cutting files in another file manager, I think it's fair enough to use your function `Fm::isCurrentPidClipboardData()` to remove gray-out from pcmanfm-qt altogether; otherwise, `findItemByPath()` will be inevitable and pcmanfm-qt will take a lot of CPU time when 6000 files are cut in Dolphin, for example. IMHO, there may be no other solution with the current structure of the code and it wouldn't do any harm.

vc-01:
wow, this has some pace.

If you solved the most problematic (speed) part, I think we should start from here, and solve bugs on top of these changes - if drawing is slow or not functions when cut made in another file manager, this is not the first issue to look at rather corner case.

Sorry that I have time only for discussion not for patching, maybe during the weekend I will.

For the painter opacity question - it just look reasonably better, I'm a Qt and C++ newbie, if you see something misused or questionable only point to that and don't spend time thinking why it's done that way. What I can better answer are architectural questions and reasoning.

I think the easiest way not to overlook something from you work is to send me a full patch. Github even supports PRs for any branches for any repos, so you can PR to my branch and we can be sure not to omit anything.

Other questions open I will integrate easily myself (methods removal etc.).

vc-01:
I think we can agree on removing UI graying response for external file managers. The top most important thing is to have it fast and bug-free in-house.

tsujan:
> wow, this has some pace.

I'd promised to look into it today and I did. I don't think I'll have time tomorrow. I patch my libfm-qt accordingly and use it for a while to see if there's any other problem.

> if drawing is slow or not functions when cut made in another file manager, this is not the first issue to look at rather corner case.

I completely agree.

> Sorry that I have time only for discussion not for patching, maybe during the weekend I will.

Take your time.

> For the painter opacity question...

It might have a little of positive impact on speed but I'm not sure. Anyhow, I always save painters before doing anything with them and afterward, restore them.

> I think the easiest way not to overlook something from you work is to send me a full patch

I have it offline. This is your work and, IMO, there's no need to 2 branches. But OK, if our ways diverge too much, I'll make a branch -- however, I don't think that'll happen.

vc-01:
> I have it offline..

Or just `git format-patch <revision range>` and send the file by any means and you have you work authored under your name at least.

tsujan:
This isn't about the PR:

@vc-01's work made me realize that `findItemByPath()` and similar functions may be the main reason why pcmanfm-qt works slowly where Dolphin works fast (https://github.com/lxde/pcmanfm-qt/issues/462). Such cases aren't many but, IMO, the structure of the libfm-qt code should be improved in this regard at some point.

tsujan:
A screencast that shows how fast cut files are grayed out: https://www.dropbox.com/s/lnqx27nqyb9jeer/cut_files.mkv?dl=0

Also corrected the latest functions in the above comments.

tsujan:
Encountered a bug: when files are cut among search results, not only they aren't grayed out (unimportant) but also their state isn't reflected by the usual view (important).

palinek:
> Encountered a bug: when files are cut among search results,

Everything done here assumes, that the "cut files" are always in the very same directory....not? But they don't need to be... this is an example.

tsujan:
> Everything done here assumes, that the "cut files" are always in the very same directory

Yes. Could you give a counterexample? (You forgot it in the above commit).

tsujan:
So far, so good.

I thought about cutting files inside a symlink folder and its target simultaneously but saw that Dolphin didn't do it either. So, I don't think it's a problem -- paths are different after all.

I started to test this since yesterday and will keep testing and will report (and probably fix) bugs that might be found. Small aesthetic changes might be needed to be made, especially in `utilities.cpp` → `copyFilesToClipboard()` and cutFilesToClipboard() because the `QByteArray` for libfm-qt is very different from those of gnome and kde and is only used for gray-out -- that's a little weird to me.

vc-01:
@palinek I'm rethinking variables naming. I think it's not an issue if user in some less common way cut files - like in search results and then files aren't grayed out when browsing back directories. Especially when they aren't gray out in search result.

Confusing may be variable naming, because it looks like it's some low level cut files stuff - but it's not, it's used only for visual purposes, sometimes even skipped (hashes search for more than 450 cut files).

But if agreed upon, multi directory support is doable (will pollute the code a bit though).

tsujan:
> But if agreed upon, multi directory support is doable

Could you explain how it might be possible to do multi-dir cut *in pcmanfm-qt* (since we don't gray out files when the cut is done by another app)?

vc-01:
> Could you explain how it might be possible to do multi-dir cut in pcmanfm-qt

e.g. if you have files `testdir1/a1.test` and `testdir2/a2.test` and you search with _"Search in sub directories"_ option from the upper dir for `*a*` pattern  and cut the result.

tsujan:
> e.g. if you have files testdir1/a1.test and testdir2/a2.test and you search...

I know (and had edited my last comment) but we don't gray out files when they're cut n the search view.

tsujan:
Actually, the search view has several problems because searching is somehow "enforced". If, at some point, that's fixed, we could think about multi-dir gray-out. And as you said, it's possible,

tsujan:
OK. BTW, an explanation about `hadCutFiles()`:
It wasn't needed to correctly set cut files but without it, all indexes in all open views would be refreshed on copying anything anywhere. At first, that didn't seem to be a problem but when I opened a few windows with folders containing many files, using `Ctrl+C` multiple times and consecutively in a text editor caused a tangible CPU usage. So, I saw the need for finding the folder that contained the previous cut file(s) to updates only its indexes and only once.

src/core/dirlistjob.cpp
```diff
@@ -62,6 +63,12 @@ void DirListJob::exec() {
             false
     };
     if(enu) {
+        bool bindCutFiles = false;
+        if(cutFilesHashSet_
+                && cutFilesHashSet_->size() < 450 /* give up if more than 450 cut files */) {
```
tsujan: @vc-01 Could you please remove the condition about 450 files here and at FileInfoJob::exec()? It isn't needed anymore (there's no problem with thousands of files); moreover, it's confusing to the user.

vc-01:
> explanation about hadCutFiles()

`hadCutFiles()` is harder to understand in looking in the source, maybe some source comment will be helpful (for the future).

tsujan:
> hadCutFiles() is harder to understand in looking in the source, maybe some source comment will be helpful (for the future).

I agree. It's simple and, at the same time, tricky, and especially this part in `setCutFiles` is really needed but the reason may not be clear:
```cpp
    if(cutFilesHashSet_ && !cutFilesHashSet_->empty()) {
        oldCutFilesDirPath_ = cutFilesDirPath_;
    }
```

src/core/dirlistjob.cpp
vc-01: It looks like it's OK to remove it, it's fast for >1000 files, but be aware it's very different code path when selecting & cutting files inside a directory and go to a directory with already selected & cut files - hash set search is done only in the latter case.

tsujan: > but be aware it's very different code path when selecting & cutting files inside a directory and go to a directory with already selected & cut files

I know but don't see any extra CPU usage in the latter case with ~4000 cut files. Do you see any?

tsujan: By the way, what about adding `!cutFilesHashSet_.isEmpty()`? Is it needed?

tsujan: Sorry, corrected it ;)

vc-01: > I don't see any extra CPU usage in the latter case

I searched for what's the best implementation for this particular case and `std::set` was a winner, probably it's very capable with even much bigger numbers - but not tested. I will remove the condition.

> what about adding !cutFilesHashSet_->empty()? Is it needed?

It's tested already in `folder.cpp:693` and `folder.cpp:258` with `Folder::hasCutFiles()`

tsujan:
Yes, `lastCutFilesDirPath_` is a better name for that variable.

Unimportant suggestions:

`hadCutFilesInvalidate` and `hadCutFilesUnset` sound a little weird to me.

This comment may also need to be changed:
````
// Add current pid to skip processing own data
````
To something like this:
```
// Add current pid to trace cut/paste operations to current app
```

tsujan:
I haven't encountered any problem in 2 days. Tomorrow, I'll test with remote files.

vc-01:
> hadCutFilesInvalidate and hadCutFilesUnset sound a little weird to me

I've encountered with such naming when functions do something more which should be mentioned in their names. As you can see I've had problem with this name too, cause I changed it twice.

`hadCutFiles` is not very good name IMO, because it does something more - invalidates / unsets the flag. I think functions like `isSet` etc. shouldn't change states unless mentioned - do you have a suggestion for the name? I don't have a problem to rename it.

> // Add current pid to trace cut/copy operations to current app

_Will change_.

> GTM

Please let me do the squashing first, this branch history is unneeded to keep IMO.

tsujan:
@vc-01 With it comes to names, tastes differ. That was the reason why I said those suggestions were unimportant.

> Please let me do the squashing first, this branch history is unneeded to keep IMO.

Do so if you see it better in that way but I don't think it's important either. To me, the important thing is the nice job you've done here and it works without problem as far as I tested/used it.

vc-01:
@tsujan I change function names a lot during development and I'm not "sure" with its names often, that's why I'm very open to change it to whatever name. Only its first name confused me a bit when tried to understand its behavior.

Squashing: First commit message went a bit obsolete and it's unnecessary for anyone to follow development path of this branch in the future IMO.

tsujan:
> that's why I'm very open to change it to whatever name.

Frankly, no good name came to my mind because it checks something and invalidates it if found. If I hear of a better name in my dreams, I'll tell you ;) Fortunately, a name can't be an obstacle.

tsujan:
Just for the record:

This or https://github.com/lxde/libfm-qt/pull/112, whichever is merged first, the other one might need a rebase.

tsujan:
@vc-01 `proxyfoldermodel.cpp` has a conflict because of the file dialog commits; please rebase!

agaida:
@tsujan: God to go?

tsujan:
@agaida Yes, it's good. Please merge it!

tsujan:
Did it myself.

@vc-01 Thanks again for the good work!

PR #90, opened 2017-06-21
Update size column info

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/514 just by removing a condition.


tsujan:
Tested and simple. Merged.

PR #91, opened 2017-06-23
folderview: Don't allow D&D by Back or Forward

palinek:
By default Qt allows D&D by any mouse button.

fixes lxde/pcmanfm-qt#486


palinek:
@tsujan What do you think?

fulalas:
It's working flawlessly now! Thanks! :)

tsujan:
> @tsujan What do you think?

Since my mouse doesn't have side buttons, I can't test this but, on the one hand, your code seems correct and, on the other hand, @fulalas says it works fine. So, merge it!

PR #92, opened 2017-06-29
Fix detailed list crash on double clicking folders

tsujan:
Tries to fix https://github.com/lxde/pcmanfm-qt/issues/458

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -225,6 +225,10 @@ FolderViewTreeView::FolderViewTreeView(QWidget* parent):
 
     header()->setStretchLastSection(true);
     setIndentation(0);
+    /* the default true value may cause a crash on entering a folder
+       by double clicking because of the viewport update done by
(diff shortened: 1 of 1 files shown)


agaida:
GTM

tsujan:
@agaida You mean that, like me, you saw the crash but after this it doesn't occur?

agaida:
you shouldn't haved ask - tried it another ten times, was about to give it up - and it crashed again

tsujan:
Ha ha. I used it more than 30 times and didn't see a crash. But logically, it should have no effect.

I don't close it; it might be changed to a real fix later.

agaida:
my fault - don't restarted the second test machine after installing the fixed packages - so the old pcmanfm-qt was running ,,,

tsujan:
> i would tend to ship it

It's completely harmless as we don't have tree expanding in the folder view. If it has effect, that might be because of a Qt bug, which is worked around in this way.

agaida:
ok, works without any negative impact and does what it should to. GTM

PR #94, opened 2017-07-01
Fix Detailed List view DND

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/520


tsujan:
Tested and simple. Merged.

PR #95, opened 2017-07-03
MountOperationQuestionDialog: Fix handling responses

vc-01:
Remove check for 'if(r != QDialog::Accepted)' because
all buttons have 'QMessageBox::AcceptRole' set.

Implement close event to process click on title bar's X button.

----
ref. issue [#519](https://github.com/lxde/pcmanfm-qt/issues/519)

### DIFF
--- a/src/mountoperationquestiondialog.cpp
+++ b/src/mountoperationquestiondialog.cpp
@@ -35,14 +35,12 @@ MountOperationQuestionDialog::MountOperationQuestionDialog(MountOperation* op, g
     choiceButtons = new QAbstractButton*[choiceCount];
     for(int i = 0; i < choiceCount; ++i) {
         // It's not allowed to add custom buttons without standard roles
-        // to QMessageBox. So we set role of all buttons to AcceptRole and
-        // handle their clicked() signals in our own slots.
-        // When anyone of the buttons is clicked, exec() always returns "accept".
+        // to QMessageBox. So we set role of all buttons to AcceptRole.
+        // When any of the set buttons is clicked, exec() always returns "accept".
         QPushButton* button = new QPushButton(QString::fromUtf8(choices[i]));
         addButton(button, QMessageBox::AcceptRole);
         choiceButtons[i] = button;
     }
-    connect(this, &MountOperationQuestionDialog::buttonClicked, this, &MountOperationQuestionDialog::onButtonClicked);
 }
 
 MountOperationQuestionDialog::~MountOperationQuestionDialog() {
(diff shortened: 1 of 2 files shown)


agaida:
@tsujan - beside the issue - what is the state of this PR?

fulalas:
I'm using it right now, and so far so good. I just would like to add some usability improvement: https://github.com/lxde/pcmanfm-qt/issues/519#issuecomment-312790321

Besides that, GTM. :+1:

tsujan:
@agaida Sorry for not seeing your question! Since the report was confusing to me, I didn't follow the discussion anymore. However, I trust @vc-01's work and found no problem by a glance at his patch. So, if it does what it's expected to do and has no side effect, it'll be good to merge.

agaida:
@tsujan - thanks for the review

agaida:
@tsujan - i'm out of package building and testing for maybe one or two days - perl transition in unstable :D

tsujan:
@agaida Take your time! I might make a PR for wallpaper slide-show tomorrow. Since I've used it for half a year, it could wait for 2 or more days.

Perl version is 5.26.0 in Arch and Manjaro. I hope that problem won't happen here.

agaida:
No, no problem, but - 
`2017/07/21 21:23:20 <agaida> ok - no bad things would happen, but the desktop would look a little bit different, the good thing is that the console still should work fine` http://paste.debian.net/977558/ https://release.debian.org/transitions/html/perl5.26.html

tsujan:
> the good thing is that the console still should work fin

... laughed for a whole minute... :)

tsujan:
Couldn't you not say such a thing? I should turn off the laptop and clean the laughter spit off its monitor...

tsujan:
@agaida In last few days I've realized that when I compile an app source (LXQt or not), the executable file has the mime `application-x-sharedlib` instead of `application-x-executable`.  For example, `pcmanfm-qt` is an `application-x-sharedlib`, while it was an `application-x-executable` in a package I made 3-4 days ago. Can this have any relation to Perl update? Is it intentional or a bug?

agaida:
i don't have a clue - but we can investigate it if the transition is done in a different system

tsujan:
Good! I also saw that a recently updated package like kwin includes `/usr/bin/kwin_x11` as `application-x-sharedlib` while that file was `application-x-executable` last week!! Therefore, that isn't our problem but it's good to know the cause.

tsujan:
OK, found it. This is about GCC's PIE (position-independent executable). At least in Arch, the default gcc option is `-pie` now (instead of `-no-pie`). As a result, the executables compiled recently are `application-x-sharedlib` -- even Firefox's is so.

They say PIE improves security with a kind of randomization. In the long time I've used Linux, I've never had security issues. I just hope such changes won't result in poor performance -- I mean something like wayland because security was one of the reasons behind making it.

vc-01:
> and found no problem by a glance at his patch

I don't rate this patch as the most understandable one. I wrote for it an explanation [comment](https://github.com/lxde/pcmanfm-qt/issues/519#issuecomment-312607209) but on different thread.

Anyway I was too busy that I overlooked this discussion these last two days.

tsujan:
> I don't rate this patch as the most understandable one

By that glance, I saw what you mean ;)

vc-01:
> By that glance, I saw what you mean ;)

ok, now I have to laugh, haha .. :)

(but the patch has some foundation and i'm quite sure with it, we can be calm about it and continue on other work)

PR #96, opened 2017-07-03
Drops Qt5Core_VERSION_STRING

luis-pereira:
Use Qt5Core_VERSION. Qt5Core_VERSION_STRING is a compatibility variable and
it was removed in Qt 5.9 release.


tsujan:
GTM.

Oh, is it compatible with Qt≤5.8 but not too old?

agaida:
Works fine with 5.7.1 - GTM

luis-pereira:
@tsujan Yes, It's compatible. Actually `Qt5XXX_VERSION_STRING` is deprecated since Qt 5.2, AFAIR.

PR #97, opened 2017-07-07
FolderView: Optimize selectAll()

palinek:
As mentioned in the comment, by using of QItemSelectionModel::select(const
QModelIndex &index, QItemSelectionModel::SelectionFlags command) we were
emitting the selectionChanged() signal for each row. By constructing
the QItemSelection and calling the right QItemSelectionModel::select()
overload, we emit just one signal.

revealed while reviewing the lxde/pcmanfm-qt#526


tsujan:
GTM

PR #98, opened 2017-07-07
Inline Renaming

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/468 by implementing inline renaming for the icon, thumbnail and compact views (the detailed list view needs a separate handling).

Inline renaming of the current item is started by the context menu-item `Rename` or `F2` when only one file/folder is selected. It is finished by Enter/Return or by removing focus from the item, and canceled by Esc. If several items are selected, renaming will be done as before, i.e. by calling the rename dialog.

For user convenience, wheel scrolling is disabled for the folder view during inline renaming.

Another PR for desktop is not needed after recent changes in libfm-qt but a pcmanfm-qt PR for `F2` will follow this.

A screenshot for icon and compact views:
inline_renaming

### DIFF
--- a/src/filemenu.cpp
+++ b/src/filemenu.cpp
@@ -31,6 +31,7 @@
 #include "customaction_p.h"
 
 #include <QMessageBox>
+#include <QAbstractItemView>
 #include <QDebug>
 #include "filemenu_p.h"
 
@@ -328,6 +329,13 @@ void FileMenu::onPasteTriggered() {
 }
 
 void FileMenu::onRenameTriggered() {
+    if (QAbstractItemView* view = qobject_cast<QAbstractItemView*>(parentWidget())) {
+        // if there is a view and this is a single file, just edit the current index
+        if (view->currentIndex().isValid() && files_.size() == 1) {
+            view->edit(view->currentIndex());
+            return;
+        }
+    }
     for(auto& info: files_) {
         Fm::renameFile(info, nullptr);
     }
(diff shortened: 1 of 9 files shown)


tsujan:
Oh, I should rebase this again!

tsujan:
Rebased.

tsujan:
Please test/review this before another rebase would be needed -- I'm not comfortable with rebasing after rebasing ;)

Please also note that , because of recent changes in libfm-qt, NO separate PR is needed for desktop anymore; everything is done here.

agaida:
erm - only one question - how to enable inline renaming?

tsujan:
@agaida Inline renaming automatically takes over when only one file is going to be renamed. Dialog renaming is kept for multiple files. I saw no reason to make inline renaming optional -- IMO, it's and overkill.

tsujan:
@agaida I forgot something (that wasn't a problem in the previous version of this patch): `f2` still uses the rename dialog, even with single files. This should be fixed in pcmanfm-qt -- a fact that I overlooked when I said there was no need to a separate PR for pcmanfm-qt. In other words, a separate pcmanfm-qt PR *is* needed for `f2`.

So, the test should be done with context menu.

tsujan:
The `F2` PR is here: https://github.com/lxde/pcmanfm-qt/pull/544

agaida:
Tested and works like charm.

PR #100, opened 2017-07-09
Avoid using grid size of QListView so we can have margins in the folder views.

PCMan:
Because we use "grid size" of QListView to enforce uniform icon size and better icon layout, we lost proper margins since the QListView "icon spacing" settings cannot be used along with "grid size". So the folder view of pcmanfm-qt looks pretty ugly for years.

Before the PR:
folder_view_no_margins

Finally I figured out how to do it properly with QListView.
After the PR:
folder_view_margins

There are still some glitch and the spacing settings still requires some tuning, I think it looks much nicer now.

Also, this PR moves the text shadow painting code from PCManFM::DesktopItemDelegate to Fm::FolderItemDelegate. So pcmanfm-qt can reuse the class directly without adding its own item delegate.
https://github.com/lxde/pcmanfm-qt/pull/529

### DIFF
--- a/src/folderitemdelegate.cpp
+++ b/src/folderitemdelegate.cpp
@@ -22,6 +22,7 @@
 #include "foldermodel.h"
 #include <QPainter>
 #include <QModelIndex>
+#include <QAbstractItemView>
 #include <QStyleOptionViewItem>
 #include <QApplication>
 #include <QIcon>
@@ -34,7 +35,6 @@ namespace Fm {
 
 FolderItemDelegate::FolderItemDelegate(QAbstractItemView* view, QObject* parent):
     QStyledItemDelegate(parent ? parent : view),
-    view_(view),
     symlinkIcon_(QIcon::fromTheme("emblem-symbolic-link")),
     fileInfoRole_(Fm::FolderModel::FileInfoRole),
     iconInfoRole_(-1) {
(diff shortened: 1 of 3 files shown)


PCMan:
@tsujan I think the PR still requires a little bit fine tuning for the optimal marings/spacing. Also the mouse click hit-test for the items might need some fine-tune as well. But I don't have enough to finish them. Would you mind help since you are quite familiar with this part?
With this PR also we can eradicate the code duplication in PCManFM::DesktopItemDelegate, which is great. :)

tsujan:
@PCMan 
Sure! As soon as I find some free time, I'll look into it.

And good to see you back :)

tsujan:
OK, I succeeded in adjusting margins according to pcmanfm-qt's Preferences. I'll fix the mouse click region later and add a commit for both. The PR may need other fixes too.

I hope there's no haste in this because fixes might take some time (due to my limited free time) -- but not too long.

tsujan:
Solved the click area problem too. Some cleanup + test + a commit → ready to merge. Maybe tomorrow :)

tsujan:
Added a commit. Everything seems OK now. This is with a 22x32 margin setting in pcmanfm-qt:

correct_spacing


Please test it, tell me if there's a problem and merge it if not. Please also note that https://github.com/lxde/pcmanfm-qt/pull/529 is necessary with this -- desktop will crash without it.

fulalas:
Is this issue, by any chance, has a connection to https://github.com/lxde/pcmanfm-qt/issues/516?

tsujan:
> Is this issue, by any chance, has a connection to...

Please be patient! This isn't an issue but a PR. I'll fix detailed list view spacing after this is merged.

fulalas:
Hahaha! Sorry, I was just curious. Thanks for the feedback :)

src/folderitemdelegate.cpp
```diff
@@ -45,52 +45,30 @@ FolderItemDelegate::~FolderItemDelegate() {
...
-
-        // "opt.decorationSize" may be smaller than the requested size because
-        // "QStyledItemDelegate::initStyleOption()" uses "QIcon::actualSize()" to set it
-        // (see Qt -> qstyleditemdelegate.cpp). So, we always get decorationSize from "option".
-        Q_ASSERT(itemSize_ != QSize());
-        QRectF textRect(0, 0, itemSize_.width(), itemSize_.height() - decorationSize.height());
-        drawText(nullptr, opt, textRect); // passing nullptr for painter will calculate the bounding rect only.
-
-        // to draw text shadow, we need one more pixel for it.
-        if(shadowColor_.isValid()) {
```
PCMan: @tsujan Why is this removed?
The shadow of text, if drawn, requires a one pixel shift of the original text area.
So the bounding box needs to be 1 pixel larger than that when no shadow is drawn.

tsujan: I removed the whole calculation because we set the item size now -- but I forgot about the shadows.

Please let me first take a closer look because the calculations are correct without shadow.

tsujan: OK, I took a closer look. We have enough space for shadow because we add 2 px at `FolderView::updateGridSize()`.

On the other hand:

(1) Shadow wasn't considered in `FolderItemDelegate::drawText` → `selRect`; so adding 1px for it would have no effect on the space between text and its selection rectangle borders.

(2) Adding 1px for shadow and adjusting `selRect` would *really* have no visible effect other than decreasing the number of desktop items per column, which wouldn't be good at all (we would lose something and gain really nothing).

agaida:
GTM

PR #102, opened 2017-07-16
Fix compact view regression

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/537.

`uniformItemSizes` had side effects not only for compact view but also when set for icon view because compact view would be broken after switching to it from icon view. On the other hand, `uniformItemSizes` had no use for us because we set item sizes in icon view.


tsujan:
This is very simple and needed by forthcoming fixes. Merging.

PR #103, opened 2017-07-17
Fix all height issues in horizontal layouts

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/516 and https://github.com/lxde/pcmanfm-qt/issues/524.

It also fixes other unreported problems regarding item heights in detaled list view, compact view and side-pane. For example, previously, when a size of 24px was requested for side-pane icons, if the icon theme had only 22px icons (the standard in KDE), the height would be 22px. With this patch, all heights are precisely as they should be.


agaida:
works fine - GTM

PR #105, opened 2017-07-25
Cut files fixes

vc-01:
These two bugfixes were written during the work on graying out cut files and are a separate thing from the mentioned work.

How to test patches is written it their respective comment message.

Dolphin uses mime type of `application/x-kde-cutselection` for cut files, assuming this is the KDE way, as Dolphin is the main file managing tool.

### DIFF
--- a/src/utilities.cpp
+++ b/src/utilities.cpp
@@ -88,8 +88,8 @@ void pasteFilesFromClipboard(const Fm::FilePath& destPath, QWidget* parent) {
     if(paths.empty() && data->hasUrls()) {
         // The KDE way
         paths = Fm::pathListFromQUrls(data->urls());
-        QByteArray cut = data->data("x-kde-cut-selection");
-        if(!cut.isEmpty() && cut.at(0) == '1') {
+        QByteArray cut = data->data(QStringLiteral("application/x-kde-cutselection"));
+        if(!cut.isEmpty() && QChar::fromLatin1(cut.at(0)) == QLatin1Char('1')) {
             isCut = true;
         }
     }
@@ -113,7 +113,7 @@ void copyFilesToClipboard(const Fm::FilePathList& files) {
     data->setData("x-special/gnome-copied-files", QByteArray("copy\n") + urilist.replace("\r\n", "\n"));
     // The KDE way
     data->setData("text/uri-list", urilist);
-    // data.setData("x-kde-cut-selection", "0");
+    // data->setData(QStringLiteral("application/x-kde-cutselection"), QByteArrayLiteral("0"));
     clipboard->setMimeData(data);
 }
 
@@ -126,7 +126,7 @@ void cutFilesToClipboard(const Fm::FilePathList& files) {
     data->setData("x-special/gnome-copied-files", QByteArray("cut\n") + urilist.replace("\r\n", "\n"));
     // The KDE way
     data->setData("text/uri-list", urilist);
-    data->setData("x-kde-cut-selection", "1");
+    data->setData(QStringLiteral("application/x-kde-cutselection"), QByteArrayLiteral("1"));
     clipboard->setMimeData(data);
 }
 
--- a/src/utilities.cpp
+++ b/src/utilities.cpp
@@ -97,6 +97,7 @@ void pasteFilesFromClipboard(const Fm::FilePath& destPath, QWidget* parent) {
     if(!paths.empty()) {
         if(isCut) {
             FileOperation::moveFiles(paths, destPath, parent);
+            clipboard->clear(QClipboard::Clipboard);
         }
         else {
             FileOperation::copyFiles(paths, destPath, parent);
--- a/src/utilities.h
+++ b/src/utilities.h
@@ -21,7 +21,6 @@
 #define FM_UTILITIES_H
 
 #include "libfmqtglobals.h"
-#include <QClipboard>
 #include <QUrl>
 #include <QList>
 #include <libfm/fm.h>


src/utilities.cpp
```diff
@@ -88,15 +88,16 @@ void pasteFilesFromClipboard(const Fm::FilePath& destPath, QWidget* parent) {
     if(paths.empty() && data->hasUrls()) {
         // The KDE way
         paths = Fm::pathListFromQUrls(data->urls());
-        QByteArray cut = data->data("x-kde-cut-selection");
-        if(!cut.isEmpty() && cut.at(0) == '1') {
+        QByteArray cut = data->data(QStringLiteral("application/x-kde-cutselection"));
```
tsujan: Good catch!

tsujan:
Now files are correctly cut from Dolphin and pasted into pcmanfm-qt.

GTM.

vc-01:
I'm once again **impressed** with these fast responses. One may expect longer response times in commercial projects (this is very encouraging). @tsujan I hope you'll keep up. Still, I believe in this project too.

tsujan:
@vc-01 I was testing an idea and compiling libfm-qt again and again when I saw your patch. So, it was a good opportunity to take look at it ;)

> @tsujan I hope you'll keep up

I really hope you'll continue to work on libfm-qt and pcmanfm-qt. I was impressed by your gray-out-on-cutting PR but, unfortunately, there were so many jobs that I didn't have time to take a deeper look into it :(

The active members of LXQt aren't many and they're usually busy. So, please don't get disappointed if the responses come late.

palinek:
GTM

src/utilities.cpp
```diff
@@ -88,8 +88,8 @@ void pasteFilesFromClipboard(const Fm::FilePath& destPath, QWidget* parent) {
...
+        if(!cut.isEmpty() && cut.at(0) == QLatin1Char('1')) {
```
luis-pereira: Consider this:
``` c++
 if(!cut.isEmpty() && QChar(cut.at(0)) == QLatin1Char('1')) {
```

src/utilities.cpp
```diff
@@ -126,7 +126,7 @@ void cutFilesToClipboard(const Fm::FilePathList& files) {
     data->setData("x-special/gnome-copied-files", QByteArray("cut\n") + urilist.replace("\r\n", "\n"));
     // The KDE way
     data->setData("text/uri-list", urilist);
-    data->setData("x-kde-cut-selection", "1");
+    data->setData(QStringLiteral("application/x-kde-cutselection"), "1");
```
vc-01: _Will change_.

src/utilities.cpp
vc-01: This looks like even necessary, `QChar` is not used in the related Dolphin code, but _will change_, if tests pass.

luis-pereira: @vc-01 
 All new code should compile with:
``` cmake
add_definitions(
    -DQT_USE_QSTRINGBUILDER
    -DQT_NO_CAST_FROM_ASCII
    -DQT_NO_CAST_TO_ASCII
    -DQT_NO_URL_CAST_FROM_STRING
    -DQT_NO_CAST_FROM_BYTEARRAY
)
```
From the doc:

> If you defined QT_NO_CAST_FROM_ASCII and QT_NO_CAST_TO_ASCII, as explained in the QString documentation, you will need to explicitly call fromLatin1(), or use QLatin1Char, to construct a QChar from an 8-bit char, and you will need to call toLatin1() to get the 8-bit value back.

But you are right, the code I proposed doesn't work. `QChar::QChar(char ch)` is private.
``` c++
QChar::fromLatin1(c.at(0))
```
does the job

tsujan:
@luis-pereira If a developer tried to fix unrelated things in a single PR, we would recommend him make separate PRs. You're right about such details but, IMHO, they place isn't this PR. The current PR has a clear goal, which it achieves. What you rightly see as a problem (although very small to me) was there before and if such things should be fixed, they should be included in a separate PR with a different title.

src/utilities.cpp
vc-01: @luis-pereira Ok, why I don't have mentioned build definitions present in my build files, I will study later. I will update the commit now, to make it ready first. (again force pushed, it's small to see it whole).

vc-01:
@tsujan Hmm, true is I was, who started with these conversions. I took them without further thinking from Dolphin code - they worked so I _felt_ it's a more correct solution, without understanding details (it will be something with X protocol is my _guess_).

You agreed to @luis-pereira's comments and I committed those changes already. I will change it to whatever final decision is.

tsujan:
@vc-01 It' OK. What I said was meant generally. @luis-pereira and I have had our disagreements on relative importance of things ;)

luis-pereira:
@tsujan My comments were only about changed code. Nothing more.

tsujan:
> ...Nothing more.

@luis-pereira It's really OK and it would be OK if it was more -- I didn't mean "shoulds" (although I used the word) but only preferences. Sorry for any misunderstanding I may have caused!

luis-pereira:
@tsujan @vc-01 
Just to be clear:
I do prefer that new/changed code does not rely on implicit strings conversions. KDE is almost free of implicit string conversions.
We don't have a policy stating that, so I gave a GTM anyway giving a suggestion: _Pls consider the mentionated_ changes.

tsujan:
@luis-pereira I'd understood your intention. I just told my opinion about why this wasn't a good place to do such things (because the purpose of the PR was quite different) and why it's better to do such jobs in a separate PR. I also understood your [reply](https://github.com/lxde/libfm-qt/pull/105#issuecomment-318156084) .

PR #106, opened 2017-07-26
Support custom folder icons

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/547.

This patch does two interrelated jobs:

(1) It reads the file ".directory" inside a (native) folder if it exists and sets the icon of that folder accordingly;

(2) It also activates the icon button of the properties dialog for setting a custom icon when the chosen files are all (native) directories. The parent folder will be automatically reloaded on clicking `OK` if a (different) custom icon is selected.

NOTES:

● To test it, you could click on the icon button of the properties dialog of a folder or multiple folders; then, a dialog will pop up, showing the active icon theme folder for selecting a custom icon.

● I couldn't find any tangible difference, with and without applying this patch, in the time needed to load a directory with ~340 folders after a cold boot.

● If approved, another patch will follow this to set bookmark icons accordingly, so that their icons be identical to their corresponding folders.

### DIFF
--- a/src/core/fileinfo.cpp
+++ b/src/core/fileinfo.cpp
@@ -169,10 +169,31 @@ _file_is_symlink:
         }
     }
 
-    /* try file-specific icon first */
-    gicon = g_file_info_get_icon(inf.get());
-    if(gicon) {
-        icon_ = IconInfo::fromGIcon(gicon);
+    /* if there is a custom folder icon, use it */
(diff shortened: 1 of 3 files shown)


agaida:
GTM

tsujan:
Wow, so fast! I thought a long discussion would follow about how it might make pcmanfm-qt heavy ;) But trust me, it doesn't do so; I tested it with a 13-year-old Desktop computer.

PR #107, opened 2017-07-26
Check nullity of FileInfo before calling FolderMenu

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/540.

(I guess the probability of a crash was higher with sshd.)


tsujan:
To reproduce the crash that's fixed by this patch, try right click into a heavy folder immediately on changing directory to it (before it's loaded) *after a cold boot*.

agaida:
Tried hard to break it - not successfull - GTM

tsujan:
> Tried hard to break it

It's unbreakable with the patch ;) This change was necessary.

PR #108, opened 2017-07-27
Arrange Custom Actions

tsujan:
Fixes from https://github.com/lxde/pcmanfm-qt/issues/438 → "Issue 2"

Arranges custom actions according to the standard explained at http://www.nautilus-actions.org/?q=node/377.

The items that are included in `~/.local/share/file-manager/actions/level-zero.directory` come first and in the order specified in that file. Other items follow them in the alphabetical order.

NOTE: The patch is logically very simple: a comparison function does the job.


agaida:
GTM

PR #109, opened 2017-07-28
Check nullity of QMimeData

tsujan:
This is done in `QAbstractItemModel::dropMimeData()` and should be done here too.

NOTE: Without this, I encountered a very rare (and unreproducible) crash, whose backtrace started with:
```
0  0x00007fbd1da1464d in QMimeData::hasUrls() const () at /usr/lib/libQt5Core.so.5
1  0x00007fbd1f0ec4e2 in  () at /usr/lib/libfm-qt.so.3
```


tsujan:
> It doesn't hurt

I agree. Moreover, having looked into the code, I saw that the above crash can't be about anything else. So, merging.

PR #111, opened 2017-08-01
Ignore button for error dialog

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/512.

When the error is not serious, an ignore button is added to the error dialog. If the user presses it instead of `OK`, the operation will be continued without showing another dialog for a non-critical error.

To test (as @fulalas [proposed](https://github.com/lxde/pcmanfm-qt/issues/512#issuecomment-312261245)):

(1) Enable "Confirm before moving files into trash can" in Preferences;
(2) Create a folder and multiple files int it;
(3) Select all of those files and try to move them to trash but do NOT confirm yet;
(4) Delete those files within a terminal;
(5) Confirm moving to trash now.

### DIFF
--- a/src/fileoperationdialog.cpp
+++ b/src/fileoperationdialog.cpp
@@ -29,7 +29,8 @@ namespace Fm {
 FileOperationDialog::FileOperationDialog(FileOperation* _operation):
     QDialog(nullptr),
     operation(_operation),
-    defaultOption(-1) {
+    defaultOption(-1),
+    ignoreNonCriticalErrors_(false) {
 
     ui = new Ui::FileOperationDialog();
     ui->setupUi(this);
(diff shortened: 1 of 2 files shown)


fulalas:
Nice to see this. But do you think that 'If the user presses it instead of OK, the operation will continue' is a good solution? I mean, maybe we could have an option to just cancel the upcoming operations. What you think?

tsujan:
@fulalas This is not about canceling any operation but about hiding non-critical errors. In the example of deleting files, there may still be non-deleted files.

Anyhow, canceling operations is something that should be done in libfm and, since libfm functionalities are going to be ported to libfm-qt, we might not see it soon.

agaida:
GTM

PR #112, opened 2017-08-02
A quick and dirty drop-in replacement for QFileDialog that is more feature-rich.

PCMan:
As the first step to implement: https://github.com/lxde/pcmanfm-qt/issues/551#event-1187260629

fm_file_dialog

### DIFF
--- a/src/filedialog.cpp
+++ b/src/filedialog.cpp
@@ -48,6 +48,8 @@ FileDialog::FileDialog(QWidget* parent, FilePath path) :
         }
     });
     ui.folderView->setModel(proxyModel_);
+    // update selection mode for the view
+    updateSelectionMode();
 
     // selection changes
     connect(ui.folderView->selectionModel(), &QItemSelectionModel::currentRowChanged, this, &FileDialog::onCurrentRowChanged);
@@ -87,6 +89,11 @@ void FileDialog::accept() {
 
     // get full paths for the filenames and convert them to URLs
     for(auto& name: parsedNames) {
+        // add default filename extension as needed
+        if(!defaultSuffix_.isEmpty() && name.lastIndexOf('.') == -1) {
+            name += '.';
+            name += defaultSuffix_;
+        }
         auto fullPath = directoryPath_.child(name.toLocal8Bit().constData());
         selectedFiles_.append(QUrl::fromEncoded(fullPath.uri().get()));
     }
@@ -145,18 +152,41 @@ void FileDialog::onCurrentRowChanged(const QModelIndex &current, const QModelInd
 
 void FileDialog::onSelectionChanged(const QItemSelection &selected, const QItemSelection &deselected) {
     QString fileNames;
-    auto selFiles = ui.folderView->selectedFilePaths();
+    auto selFiles = ui.folderView->selectedFiles();
     for(auto& fileInfo: selFiles) {
-        if(!fileNames.isEmpty()) {
-            fileNames += ' ';
+        if(fileMode_ == QFileDialog::Directory && !fileInfo->isDir()) {
+            // if we want to select dir, ignore selected files
+            continue;
         }
-        // FIXME: use a more reliable way to quote file names.
-        // otherwise names with embedded " will break.
-        fileNames += '\"';
-        fileNames += fileInfo.baseName().get();
-        fileNames += '\"';
+        else if(fileInfo->isDir()) {
+            // if we want to select files, ignore selected dirs
+            continue;
+        }
+
+        auto baseName = fileInfo->path().baseName();
+        if(fileMode_ == QFileDialog::ExistingFiles) {
+            // support multiple selection
+            if(!fileNames.isEmpty()) {
+                fileNames += ' ';
+            }
+            // FIXME: use a more reliable way to quote file names.
+            // otherwise names with embedded " will break.
+            fileNames += '\"';
+            fileNames += baseName.get();
+            fileNames += '\"';
+        }
+        else {
+            // support single selection only
+            fileNames = baseName.get();
+            break;
+        }
     }
     ui.fileName->setText(fileNames);
+}
+
+void FileDialog::updateSelectionMode() {
+    // enable multiple selection?
+    ui.folderView->childView()->setSelectionMode(fileMode_ == QFileDialog::ExistingFiles ? QAbstractItemView::ExtendedSelection : QAbstractItemView::SingleSelection);
 }
 
 QUrl FileDialog::directory() const {
@@ -172,18 +202,17 @@ void FileDialog::selectFile(const QUrl& filename) {
     // FIXME: add a method to Fm::FolderView to select files
 
     // FIXME: need to add this for detailed list
-    //int flags = QItemSelectionModel::Rows;
-    ui.folderView->selectionModel()->select(idx, QItemSelectionModel::Select);
+    QItemSelectionModel::SelectionFlags flags = QItemSelectionModel::Select;
+    if(viewMode_ == QFileDialog::Detail) {
+        flags |= QItemSelectionModel::Rows;
+    }
+    ui.folderView->selectionModel()->select(idx, flags);
 }
 
 QList<QUrl> FileDialog::selectedFiles() {
     return selectedFiles_;
 }
 
-
-void FileDialog::setFilter() {
-    // FIXME: what's this?
-}
 
 void FileDialog::selectNameFilter(const QString& filter) {
     if(filter != currentNameFilter_) {
@@ -224,18 +253,17 @@ void FileDialog::setViewMode(QFileDialog::ViewMode mode) {
     default:
         break;
     }
+
+    // update selection mode for the view
+    updateSelectionMode();
 }
 
 
 void FileDialog::setFileMode(QFileDialog::FileMode mode) {
     fileMode_ = mode;
-    // TODO:
-    if(mode == QFileDialog::ExistingFiles) {
-        // TODO: enable multiple selection
-    }
-    else {
-        // TODO: limit to single selection
-    }
+
+    // enable multiple selection?
+    updateSelectionMode();
 }
 
 
@@ -325,7 +353,8 @@ QString FileDialog::labelText(QFileDialog::DialogLabel label) const {
 bool FileDialog::FileDialogFilter::filterAcceptsRow(const ProxyFolderModel *model, const std::shared_ptr<const FileInfo> &info) const {
     if(dlg_->fileMode_ & QFileDialog::Directory) {
         // we only want to select directories
-        if(!info->isDir() && dlg_->options_& QFileDialog::ShowDirsOnly) { // not a dir
+        if(!info->isDir()) { // not a dir
+            // NOTE: here we ignore dlg_->options_& QFileDialog::ShowDirsOnly option.
             return false;
         }
     }
--- a/src/filedialog.cpp
+++ b/src/filedialog.cpp
@@ -23,6 +23,7 @@ FileDialog::FileDialog(QWidget* parent, FilePath path) :
     ui{new Ui::FileDialog()},
     folderModel_{nullptr},
     proxyModel_{nullptr},
+    folder_{nullptr},
     options_{0},
     viewMode_{FolderView::DetailedListMode},
     fileMode_{QFileDialog::AnyFile},
@@ -306,7 +307,6 @@ void FileDialog::setDirectoryPath(FilePath directory, FilePath selectedPath) {
 
    if(folder_) {
         if(folderModel_) {
-            //stopWatchingNewFiles();
             proxyModel_->setSourceModel(nullptr);
             folderModel_->unref(); // unref the cached model
             folderModel_ = nullptr;
--- a/src/filedialog.cpp
+++ b/src/filedialog.cpp
@@ -69,16 +69,8 @@ FileDialog::FileDialog(QWidget* parent, FilePath path) :
         disconnect(ui->folderView->selectionModel(), &QItemSelectionModel::selectionChanged, this, &FileDialog::onSelectionChanged);
         ui->folderView->selectionModel()->clearSelection();
         QStringList parsedNames = parseNames();
-        if(!parsedNames.isEmpty()) {
-            QString curDir(directoryPath_.toString().get());
-            curDir += QDir::separator();
-            for(auto& name: parsedNames) {
-                // when there's no file, checking for file existence is faster than
-                // ProxyFolderModel::indexFromPath(), which is called by selectFilePath()
-                if(QFileInfo(curDir + name).exists()) {
-                    selectFilePath(directoryPath_.child(name.toLocal8Bit().constData()));
-                }
-            }
+        for(auto& name: parsedNames) {
+            selectFilePath(directoryPath_.child(name.toLocal8Bit().constData()));
         }
         updateAcceptButtonState();
         updateSaveButtonText(false);
@@ -165,7 +157,7 @@ FileDialog::FileDialog(QWidget* parent, FilePath path) :
     setSplitterPos(200);
 
     // browse to the directory
-    if(path.isValid() && QFileInfo(path.toString().get()).isDir()) {
+    if(path.isValid()) {
         setDirectoryPath(path);
     }
     else {
@@ -272,10 +264,10 @@ void FileDialog::accept() {
         if(fileMode_ != QFileDialog::Directory) {
             auto firstName = parsedNames.at(0);
             auto childPath = directoryPath_.child(firstName.toLocal8Bit().constData());
-            QFileInfo info = QFileInfo(childPath.toString().get());
-            if(info.exists()) {
+            auto info = proxyModel_->fileInfoFromPath(childPath);
+            if(info) {
                 // if the typed name belongs to a (nonselected) directory, chdir into it
-                if(info.isDir()) {
+                if(info->isDir()) {
                     setDirectoryPath(childPath);
                     return;
                 }
@@ -343,12 +335,6 @@ void FileDialog::goHome() {
 void FileDialog::setDirectoryPath(FilePath directory, FilePath selectedPath, bool addHistory) {
     if(!directory.isValid() || directoryPath_ == directory) {
         updateAcceptButtonState(); // FIXME: is this needed?
-        return;
-    }
-    if(!QFileInfo(directory.toString().get()).isDir()) {
-        QMessageBox::critical(this, tr("Error"), tr("No such directory:\n\"%1\"").arg(directory.toString().get()));
-        ui->location->setPath(directoryPath_); // toggle the previous path button
-        updateAcceptButtonState();
         return;
     }
 
@@ -842,9 +828,8 @@ void FileDialog::updateSaveButtonText(bool saveOnFolder) {
         if(!saveOnFolder) {
             QStringList parsedNames = parseNames();
             if(!parsedNames.isEmpty()) {
-                QString curDir(directoryPath_.toString().get());
-                curDir += QDir::separator();
-                if(QFileInfo(curDir + parsedNames.at(0)).isDir()) {
+                auto info = proxyModel_->fileInfoFromPath(directoryPath_.child(parsedNames.at(0).toLocal8Bit().constData()));
+                if(info && info->isDir()) {
                     saveOnFolder = true;
                 }
             }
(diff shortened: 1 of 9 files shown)


PCMan:
The next step is plug this into lxqt-qtplugin. However, if it's done in a naiive way, every Qt program will have libfm-qt loaded into their process since the plugin links to libfm-qt and the file dialog.
It's better to make it some loadable plugin or dbus service called via RPC.
I'll try to see if I can make it.

PCMan:
Note that this PR introduces a backward incompatible API change in Fm::FolderView's constructor. Without this it does not work with Qt designer UI files.

palinek:
> Note that this PR introduces a backward incompatible API change in Fm::FolderView's constructor. Without this it does not work with Qt designer UI files.

Wouldn't just adding a new constructor with the `parent` be better? (just a suggestion)

```diff
diff --git a/src/folderview.cpp b/src/folderview.cpp
index 10aaf02..5d3580f 100644
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -450,7 +450,7 @@ void FolderViewTreeView::onSortFilterChanged() {
 
 //-----------------------------------------------------------------------------
 
-FolderView::FolderView(QWidget *parent, FolderView::ViewMode _mode):
+FolderView::FolderView(ViewMode _mode, QWidget* parent):
     QWidget(parent),
     view(nullptr),
     model_(nullptr),
@@ -476,6 +476,10 @@ FolderView::FolderView(QWidget *parent, FolderView::ViewMode _mode):
     connect(this, &FolderView::clicked, this, &FolderView::onFileClicked);
 }
 
+FolderView::FolderView(QWidget* parent/* = 0*/)
+    : FolderView(IconMode, parent)
+{}
+
 FolderView::~FolderView() {
 }
 
diff --git a/src/folderview.h b/src/folderview.h
index 3c702a2..61299bd 100644
--- a/src/folderview.h
+++ b/src/folderview.h
@@ -66,7 +66,8 @@ public:
     friend class FolderViewTreeView;
     friend class FolderViewListView;
 
-    explicit FolderView(QWidget* parent = 0, ViewMode _mode = IconMode);
+    explicit FolderView(ViewMode _mode = IconMode, QWidget* parent = 0);
+    explicit FolderView(QWidget* parent = 0);
 
     virtual ~FolderView();
 
diff --git a/src/tests/test-folderview.cpp b/src/tests/test-folderview.cpp
index 427f1e5..da52c6b 100644
--- a/src/tests/test-folderview.cpp
+++ b/src/tests/test-folderview.cpp
@@ -16,7 +16,7 @@ int main(int argc, char** argv) {
     Fm::LibFmQt contex;
     QMainWindow win;
 
-    Fm::FolderView folder_view;
+    Fm::FolderView folder_view(nullptr);
     win.setCentralWidget(&folder_view);
 
     auto home = Fm::FilePath::homeDir();
```

tsujan:
@palinek, I agree with you and think your suggestion is a better way. A commit?

tsujan:
Please don't merge it until either @palinek's idea is implemented or a PR for pcmanfm-qt is added!

PCMan:
Agree, I'll fix the constructor instead.

PCMan:
@tsujan @palinek Fixed the constructor already.

tsujan:
Ready to be merged, IMO.

src/proxyfoldermodel.cpp
```diff
@@ -156,14 +156,34 @@ bool ProxyFolderModel::lessThan(const QModelIndex& left, const QModelIndex& righ
...
+        FolderModel* srcModel = static_cast<FolderModel*>(sourceModel());
+        if(srcModel) {
+            QModelIndex srcIndex = mapToSource(index);
+            return srcModel->fileInfoFromIndex(srcIndex);
+        }
     }
     return nullptr;
 }
 
+QModelIndex ProxyFolderModel::indexFromPath(const FilePath &path) const {
```
tsujan: @palinek Thank you for your attention! `indexFromPath` also checks items one by one, so I don't think it'll make any difference. That being said, I'll surely give it a try.

Our only problem in graying out cut files is about when another app cuts them -- the current functions might take a lot of CPU time in that case. For now, we don't gray out cut files in pcmanfm-qt when that happens (as a workaround).

luis-pereira:
With the Qt Dialog
screen-2017-08-04-00-44-00 png

With the new File Dialog
screen-2017-08-04-00-46-57 png

The file filter are not displayed correctly, although they are being used.

tsujan:
I saw that too -- among several other things -- but thought that all issues could be fixed after merging. Please tell if the plan is otherwise!

tsujan:
Another thing: typing inside "File name" line-edit doesn't show any list...

luis-pereira:
@tsujan I don't know of any plan. Just commenting my very little findings on this PR.

tsujan:
@luis-pereira The main part is done. Like you, I don't know if this is a WIP and commits could be added here or fixes should be in PRs after merging. @PCMan said this is "the first step", so I thought the latter is the case.

tsujan:
After finding a little time to play with the code, I admit my misunderstanding: It's a WIP and I'm sure @PCMan is aware of the gaps (options, filters, settings, etc). So, it'll be merged only when finished.

PCMan:
This is an initial attempt and it hasn't reach production quality.
I opened the PR for code review and feedbacks.
Polishing and bug fixes are needed. I'll see if I can fix the bugs when time permits.
@tsujan Feel free to push to this branch if you want to fix some obvious bugs.

tsujan:
@PCMan 
Understood. In order to not interfere with your plan and since I don't know which methods you might add to implement the missing functionalities, I prefer to wait until all functions are added.

PCMan:
Just fixed the mentioned problems in the PRs for libfm-qt and lxqt-qtplugin.
I think it's almost there. Just needs more testing.

PCMan:
@tsujan If you found any bugs, please feel free to fix them and git push. I finished the required features already and don't have much time for it now. So there is no plan for further changes from my side at the moment. Thanks!

tsujan:
> @PCMan 

First, thank you very much for this!

> @tsujan If you found any bugs, please feel free to fix them and git push. I finished the required features already and don't have much time for it now.

I would gladly if I could guess the structure you wanted to give to the code (as I explained elsewhere). But if you have no plan for more changes, OK, I might find a way to make this work. It just might take a while. And I hope others will join me in this.

tsujan:
This + https://github.com/lxde/lxqt-qtplugin/pull/23 are usable now, although there are still problems and my work isn't finished yet. Tests and issue reports are really needed.

tsujan:
BTW, @PCMan's libfm-qt branch was a little old. If you want to test it, you could preferably apply the patch to the the latest source.

tsujan:
TODO (might change):

● Focus the selected item.
● Preserve selection on view change.
● Fix the dialog name.
● Save sizes on closing.
● Add a refresh button.

tsujan:
Overwrite prompt aside, file opening and saving is wrong. For example, when it's a save dialog, clicking "Save" button when a folder is selected shouldn't try to save anything but should open that folder; etc... will make a commit to fix such issues.

tsujan:
**WARNING:** If you test this, please recompile https://github.com/lxde/lxqt-qtplugin/pull/23 after the last commit [here](https://github.com/lxde/libfm-qt/pull/112/commits/1f0a06df25c33b8d0eb1122c0f9674a46ecd00ff) (otherwise a crash will happpen because private variables are added to FileDialog)!

tsujan:
Since also `QFileDialog` has problem with parsing file names containing double-quote and since slash (/) is the only forbidden character for file names in Linux (and Unix?), if there's no objection, I wanted to use slash for separating selected file names from each other, like this:
```
/NAME1/ /NAME2/ /NAME3/
```

PCMan:
@tsujan About the setLabelText problem, it's still better to avoid overwriting the custom labels set by the user. It's a public API and can be called by the user anytime. Now since it's only used in lxqt-qtplugin the order of calling different methods is totally controlled by us, but that's not the right way to design public APIs. It's possible, though rarely, to use the class outside lxqt-qtplugin.
About the filename separator part, please don't (ab)use slashes. For now keeping it compatible with QFileDialog is OK since the behavior is well defined.
The proper way to handle names containing quotes, IMHO, is escaping them using \" properly as what we did in shell scripts.
If you don't want to implement this yourself, glib provides it already.
https://developer.gnome.org/glib/stable/glib-Shell-related-Utilities.html#g-shell-quote
That's yet another reason why I like glib a lot. It really provides what a general UNIX programmer will need nowadays.

tsujan:
@PCMan Thank you for the hint on quotes!

Labels are correctly handled by the private function `setLabelTextControl()` and those that are explicitly set are remembered by `explicitLabels_`. Nothing is overwritten; I tested with some custom labels.

Until now, the hardest part of my job was the correct setting of the temporary label (when a folder was selected with the file save dialog, for example) and the enabled/disabled state of the OK button in various situations, as `QFileDialog` did. Fortunately, it's done now and I'm testing it.

tsujan:
Oh, I forgot to push all changes I made to labels for them not to be overwritten when explicitly set! Will do it soon.

tsujan:
Escaping quotes inside names works without problem and solves QFileDialog's issue but, IMO, it's a little ugly visually -- although we can live with it.

What about prefixing inside quotes with "zero width non-joiner" , which is `QChar(0x200C)`? It looks normal and works with both single and multiple files:

quotes

As far as I know, no language on the planet uses it *before a quote*.

... [ BTW, just used the new file dialog to find and attach the screenshot with qupzilla ;) ]

tsujan:
My last question (in addition to the above one about inside quotes) to finish this:

Is `~/.config/libfm/filedialog.conf` a good place to save dialog size and its splitter size (and other settings in future)? `~/.config/libfm/` is used to save per-folder settings of pcmanfm-qt.

PCMan:
@tsujan the config file of lxqt-qtplugin seems to be a reasonable place to me. Just add a new section for it.
Since the config only applies when LXQt is in use, I think this is OK.

tsujan:
@PCMan Thanks for the reply! So, the size settings can go under `~/.config/lxqt/lxqt.conf` → [FileDialog] → `SplitterPos` and `WindowSize`?

What about prefixing filename quotes with "zero width non-joiner" in the name entry and grouping filenames with quotes as before?

PCMan:
@tsujan zero with non-joiner is a brilliant idea, but it brings some drawbacks.
The filename QLineEdit is freely editable by the user. Having invisible chars in the QLineEdit hence can cause the char be embedded into filenames accidentally if the user tries to edit the filenames by hand.
IMHO, it's acceptable to show the " embedded in a filename as \\".
It follows the same convention as UNIX shell so it's quite natural for a regular UNIX user.
for example: "filenames with \\"embedded\\" quote"
Though it looks ugly, as most of the shells do this so it won't surprise the users.

tsujan:
> The filename QLineEdit is freely editable by the user.

Yes, I realized that too but thought such an exception would be acceptable. But all right! I'll go with `\"`, as you recommended. We can live with it and its ugliness is quite tolerable; as you correctly mentioned, users won't be surprised.

tsujan:
@PCMan 

The is a serious issue with ‍‍`~/.config/lxqt/lxqt.conf‍‍‍`: every time that file is written to, all fonts in all apps will be updated, although the font isn't changed. Since the user might want to change the dialog size often, such an update isn't acceptable at all. There are 2 ways out of this:

(1) Choosing `~/.config/libfm/filedialog.conf`; or
(2) Preventing unnecessary updates when nothing changes.

Which solution do you recommend?

PCMan:
@tsujan You are right. I forgot the file change can induce UI updates.
Then maybe put it in its own config file makes sense. For example: ~/.config/lxqt/filedialog.conf
To make the components more flexible and reusable, libfm-qt Fm::FileDialog class itself does not manage config files. Instead the config can be written by lxqt-qtplugin for LXQt only.
So I'd say we can put in in ~/.config/lxqt instead of ~/.config/libfm-qt since it's not a libfm config file. It's an LXQt-only config file.

tsujan:
> libfm-qt Fm::FileDialog class itself does not manage config files.

Makes perfect sense. So, I'll add a commit only for quote fix here; reading and writing of settings will go into https://github.com/lxde/lxqt-qtplugin/pull/23.

tsujan:
Preserving selection on changing view and a refresh button could be added later. IMO, this is ready to be merged.

PCMan:
@tsujan just added the missing mime-type filter APIs.

tsujan:
>> @tsujan just added the missing mime-type filter APIs.

Then, `LXQtFileDialogHelper` needs to be corrected; doesn't it?

tsujan:
After using this for 2 days, I found 2 issues:

(1) My fault: the lambda folder connection can't be disconnected in the usual way. I'll fix that soon when adding a reload button.
(2) Up and down buttons could be ignored (because of the path-bar) but backward and forward buttons are really needed.

tsujan:
It seems we should wait a bit longer. Still 2 other issues:

(1) A shortcut for toggling hidden files is needed (easy to fix).
(2) Computer, Application, and Network can't be opened in the side-pane; Trash is usable for emptying trash but can't be opened

tsujan:
Now, it's so:

new-dialog


No other button is needed, IMO.

tsujan:
Just wanted to explain why I've been excited about this. It's already the best file dialog in Linux: It has advanced functionalities of a file manager because of libfm-qt. All my custom actions are available and usable in it. When the Trash issue and a few other problems are fixed, it'll be even better.

An hour ago, I wanted to do some file management as root. For whatever reason, `Open as root` doesn't work in pcmanfm-qt menu (I think because of dbus). So, I did an `lxsudo featherpad` and did everything from inside its open file dialog :) File management through a text editor! Actually, through LXQt's file dialog.

PCMan:
@tsujan That's just my two cents. Your code is GTM already. No need to remove the lambda if you still think it's better to use it. :-)

tsujan:
> Your code is GTM already

I've always enjoyed the structure of your codes. If you have a way to simplify it, please do it! The simpler, the better. I admit that I'm not good at that.

PCMan:
@tsujan Thanks for the compliment. :) You are quite good at debugging, which I failed to do sometimes. Well, that's how collaboration works.
Nice to have you guys in the team.

tsujan:
@PCMan I wanted to (mis-)use this opportunity and ask you about a complete C++11 port of libfm into libfm-qt. Do you have a schedule for it? It'll be really great if libfm-qt becomes self-sufficient.

I tried to do it in some places but, as I admitted above, I may not be good at creating simple structures at once. Moreover, I can't imitate your style and, IMO, the style should be kept self-consistent.

tsujan:
Oh, I encountered a problem again while trying to test QFileInfo for non-local fs:

Network, in general, and especially SFTP doesn't work. The "No such directory" messagebox is shown instead. This is the only issue from my TODO list that I didn't fix. Will try to.

tsujan:
I admit my mistake: *QFileInfo shouldn't be used at all* because we use GIO.

I think it shouldn't be used in other places of libfm-qt either.. will search for it.

tsujan:
I removed `QFileInfo` completely and used `proxyModel_->fileInfoFromPath()` instead. Data model API should be improved in the relevant files of libfm-qt instead.

The file dialog can't open files with SFTP (which isn't related to `QFileInfo`):
````
file monitor cannot be created: Operation not supported by backend
````

**Comment by tsujan** [comment]
The Computer, Application, Trash and Network problems were all caued by `QFileInfo`. Once I removed it in https://github.com/lxde/libfm-qt/pull/112/commits/b59bfc8d30977745b53ac5da34ba37afb3e2870b, all of them were fixed. `QFileInfo` was a big mistake!

**Comment by tsujan** [comment]
IMO, GTM.

**Comment by tsujan** [comment]
Although I think this can be merged, there's still a problem -- or a discrepancy -- that could be serious for some users. If it isn't fixed here, I might open an issue for it later.

I don't know about FUSE (don't have it) but have tested with Trash and SFTP. The problem is that this file dialog is for Qt apps and Qt apps may have no clue about urls like "sftp://" or "trash:///". I don't need to say anything about Kate because it can't open any trash or SFTP file through LXQt file dialog. However, this is what happens with FeatherPad:

When the trash folder or an SFTP folder is opened in pcmanfm-qt, its files can be opened in FeatherPad by clicking on them. The issue is about when the file is supposed to be opened from inside the file dialog. With trash and SFTP:

(1) If I click on a file in the file dialog, nothing will happen.

(2) However, if I right click the file (in the file dialog) and open it in FeatherPad, it'll be opened correctly.

The reason is that the QUrl isn't local in the case of (1), while it's local in the case of (2). And this the *discrepancy* in question.

**Comment by PCMan** [comment]
Any updates on this? Is it going to be merged?
@tsujan  Yes, the FUSE problem can be solved by generating local paths for the FUSE-mounted remote file systems. For example, instead of sftp://xxxx/some/path, generating file:///home/<user_name>/.gvfs/sftp.../some/path
GIO has built-in support for this. Just try to get a local path for a remote URL.

**Comment by tsujan** [comment]
>  Yes, the FUSE problem can be solved by generating local paths for the FUSE-mounted remote file systems....

What I mean is that, currently, only the remote URL (sftp://...) is emitted, which makes the file dialog useless in such cases. Shouldn't the local paths be emitted *if found*, as proposed [above](https://github.com/lxde/libfm-qt/pull/112#issuecomment-323859482)? In fact, I've done that for myself; should I do it here too?

> Any updates on this? Is it going to be merged?

I think this and https://github.com/lxde/lxqt-qtplugin/pull/23 can be merged but it'll be better if the above-mentioned problem is solved first.

**Comment by PCMan** [comment]
@tsujan Yes, if a local path for a remote file is available, we should emit that local file:/// URL by QUrl instead. In this way the file can be correctly loaded by Qt programs without knowing it's FUSE-mounted.

**Comment by tsujan** [comment]
@PCMan Good! I'll add a simple commit today and if it's acceptable to you, we could merge this and let everyone enjoy it :)

**Comment by tsujan** [comment]
Done. Please merge both PRs!

File: src/filedialog.cpp [hunk 2]
```diff
@@ -253,7 +253,13 @@ void FileDialog::accept() {
     if(parsedNames.isEmpty()) {
         // when selecting a dir and the name is not provided, just select current dir in the view
         if(fileMode_ == QFileDialog::Directory) {
-            selectedFiles_.append(directory());
+            auto localPath = directoryPath_.localPath();
```
PCMan: @tsujan The .get() check is redundant.
`if(localPath)` if enough.

src/filedialog.cpp
```diff
@@ -294,7 +300,14 @@ void FileDialog::accept() {
                 name += defaultSuffix_;
             }
             auto fullPath = directoryPath_.child(name.toLocal8Bit().constData());
-            selectedFiles_.append(QUrl::fromEncoded(fullPath.uri().get()));
+            auto localPath = fullPath.localPath();
+            /* add the local path if it exists; otherwise, add the uri */
+            if(localPath && localPath.get()) {
```
PCMan: same here

tsujan:
> The .get() check is redundant.

Removed them a few minutes after you merged it ;) had added them for the sake of certainty.

Trust2k:
First of all, thank you so much for doing this, it's amazing that only two people were puzzled by this, because every linux user faces this every day!!! 🧐

Secondly, I found a bug, and I also want to make a suggestion to improve the design of the dialog box.

1) Some of the thumbnails are not displayed. In folders with a large or small number of images, some of the thumbnails are not displayed, so you can't call the function complete. You can see that loading thumbnails starts at the end of the list and stops before reaching the middle. 





2) Is it possible to swap elements in the panel? Just like that:

tsujan:
> Some of the thumbnails are not displayed.

The image size limit for creating thumbnails is 4 MiB. A file dialog isn't a tool for seeing all thumbnails.

> Is it possible to swap elements in the panel?

No. The tool-bar is appropriate for a file dialog.

Trust2k:
> The image size limit for creating thumbnails is 4 MiB. A file dialog isn't a tool for seeing all thumbnails.

Clear, I confirm, only images less than 4 megabytes are displayed, it's strange that I didn't check it right away 🤔. Strange, but what to do if you need to upload an image to the site, but it is not possible to select it because of this stupid restriction... Maybe you can add resizing to the settings?

> No. The tool-bar is appropriate for a file dialog.

Ok.

Trust2k:
В kde миниатюры всех размеров отображаются, можно посмотреть у них как это реализовано

tsujan:
> you can see how this is implemented

We have our code and already know how to set limits. The 4-MiB limit is intentional.

Please stop this!

PR #113, opened 2017-08-06
Initialize FileInfo::isShortcut_

tsujan:
It was left uninitialized. I encountered a strange problem while fixing https://github.com/lxde/libfm-qt/pull/112 and found that this was the cause. Luckily, it wasn't able to create a serious problem in the only place it was used, namely, desktop.


palinek:
GTM

tsujan:
> Consider doing it in the constructor.

I agree that it would be better but didn't want to change the code style... maybe later

PR #115, opened 2017-09-15
Fix mistakes in listview column width calculation

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/461.

Also fixes column widths when headers are moved. In short, sizeFromContents() should have been considered.


tsujan:
An explanation:

Previously, the column width was calculated only based on its contents and without taking the header width into account. that worked in most cases but not when the header was wider than the contents.

Hear, the header width is also calculated without any false presupposition about style parameters. A space is also preserved for the sorting indicator (arrow).

PR #117, opened 2017-09-19
UI improvements for Fm::MountOperationPasswordDialog()

vc-01:
Check _Connect as user_ radio button automatically if anonymous
login is not available. Set focus to password field if username
was already entered (and field is disabled).

Setting auto focus to username field would be a dead code because
anonymous login is predefined and this field is then disabled at first anyway.

----

Before the patch we start with this window

before

After 

after

ref. to issue [#519](https://github.com/lxde/pcmanfm-qt/issues/519#issuecomment-312790321)

btw. if all two radio buttons (for anonymous and username login) are available, I vote _No_ for setting focus automatically to user name field when clicking on _Connect as user_ in this case. I tested it and didn't like it.

### DIFF
--- a/src/mountoperationpassworddialog.cpp
+++ b/src/mountoperationpassworddialog.cpp
@@ -46,11 +46,17 @@ MountOperationPasswordDialog::MountOperationPasswordDialog(MountOperation* op, G
     }
     else {
         ui->Anonymous->setEnabled(false);
+        ui->asUser->setChecked(true);
     }
     if(!needUserName) {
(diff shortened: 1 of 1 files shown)


tsujan:
This way is more logical to me too. GTM.

vc-01:
You spare two mouse clicks and it feels more natural. Also it's not a UI change imo, just improvement.

tsujan:
@vc-01 I agree.

PR #118, opened 2017-09-20
Fix partially visible toggled path buttons

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/561

Sometimes, the toggled path button wasn't scrolled to correctly (in file dialog as well as in pcmanfm-qt) because the layout of the scroll-area was being updated simultaneously. The issue is fixed by using `QTimer::singleShot()` in a proper way.

Also compensated for a small miscalculation in Qt.


tsujan:
So fast :)

PR #120, opened 2017-09-21
Enable XDS subfolder drop

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/557

With this patch, files from inside gtk archive managers can be dropped into subfolders of pcmanfm-qt. Of course, the probable prompt dialogs are provided by those apps, not by pcmanfm-qt.


agaida:
GTM

PR #121, opened 2017-09-22
Disable context-menu actions that cannot be used

tsujan:
Disable file actions that cannot be performed (ref. to https://github.com/lxde/pcmanfm-qt/issues/567).

IMO, all advanced file managers should have this feature. It prevents showing unnecessary error dialogs and serves as a kind of visual feedback about the nature of files/folders.

A pcmanfm-qt PR will follow this for disabling unperformable menubar Edit actions.


agaida:
GTM

PR #123, opened 2017-10-19
Add data transferred to file operation dialog.

nathan-osman:
When performing long-running file operations, it is helpful to be able to see the amount of data being transferred in addition to the progress and estimated time remaining. This pull request adds such a feature:

screenshot_20171018_145027


palinek:
GTM

PR #124, opened 2017-10-25
Smooth scrolling for icon and thumbnail views

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/579.

The smooth wheel scrolling is implemented when the cursor is inside the icon/thumbnail view. The scrolling is as before (jumpy) when the cursor is on the scrollbar, so that both options are available to the user.

The CPU usage is optimized by disabling the animation when the wheel speed is too high. Unlike in Dolphin, the animation respects the number of scroll lines per wheel turn, as set in the LXQt configuration.

NOTES:
(1) **pcmanfm-qt should be recompiled after this; otherwise, it will most probably crash.**
(2) A partial workaround could be added for Qt5's scroll jump bug but I did not do so here because of the subject.

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -51,6 +51,11 @@
 
 Q_DECLARE_OPAQUE_POINTER(FmFileInfo*)
 
+#define SCROLL_FRAMES_PER_SEC 50
+#define SCROLL_DURATION 300 // in ms
+
+static const int scrollAnimFrames = SCROLL_FRAMES_PER_SEC * SCROLL_DURATION / 1000;
+
 using namespace Fm;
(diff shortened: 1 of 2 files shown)


tsujan:
Although I don't see a difference in CPU usage, please don't merge this if your tests show otherwise because pcmanfm-qt should feel lightweight with all computers.

agaida:
Tested it on my 10 year old laptop and my little lenovo  (modern pentium duo) - nothing negative.

tsujan:
> Tested it on my 10 year old laptop

Thanks a lot! Such a test was needed.

PR #126, opened 2017-11-06
Prompt dialog specifically for desktop files

tsujan:
Follows https://github.com/lxde/libfm/pull/21.

The "Execute in Terminal" button is removed because running in terminal is decided by the desktop file.


agaida:
GTM

PR #127, opened 2017-11-08
Merge side-pane with its surroundings

tsujan:
Closes https://github.com/lxde/pcmanfm-qt/issues/583.

This makes sidepanes transparent -- whether in pcmanfm-qt or in LXQt file dialog -- creating a more professional look, as in KDE.

The tree-view side-pane isn't touched because otherwise, tree-lines would be invisible in themes that have them.


agaida:
Applied, works, look great. GTM

PR #128, opened 2017-11-11
Fix wrong gray-out of cut files

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/596.

The selected files of a folder are cut only if their paths are equal to the cut paths. Otherwise, the selection shouldn't be used to gray out items. Previously, the selection was always used to gray out items and that resulted in a wrong gray-out when a folder was opened in more than one tab/window.

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -1218,8 +1218,10 @@ void FolderView::onClipboardDataChange() {
         if(!folder()->path().hasUriScheme("search") // skip for search results
            && isCutSelection
            && Fm::isCurrentPidClipboardData(*data)) { // set cut files only with this app
-            auto cutDirPath = paths.size() > 0 ? paths[0].parent(): FilePath();
-            if(folder()->path() == cutDirPath) {
+            auto cutDirPath = paths.size() > 0 ? paths[0].parent() : FilePath();
+            // set the cut file(s) only if the cutting is done here
+            if(folder()->path() == cutDirPath
+               && selectedFilePaths() == paths) {
                 model_->setCutFiles(selectionModel()->selection());
             }
             else if(folder()->hadCutFilesUnset() || folder()->hasCutFiles()) {


src/folderview.cpp
```diff
@@ -1218,8 +1218,10 @@ void FolderView::onClipboardDataChange() {
         if(!folder()->path().hasUriScheme("search") // skip for search results
            && isCutSelection
            && Fm::isCurrentPidClipboardData(*data)) { // set cut files only with this app
-            auto cutDirPath = paths.size() > 0 ? paths[0].parent(): FilePath();
-            if(folder()->path() == cutDirPath) {
+            auto cutDirPath = paths.size() > 0 ? paths[0].parent() : FilePath();
+            // set the cut file(s) only if the cutting is done here
+            if(folder()->path() == cutDirPath
+               && selectedFilePaths() == paths) {
```
vc-01: Using `&& hasFocus()` instead works too and it's faster. Do you see any problems with this solutions?

All three ways to cut files require focus - by keyboard shortcut, with mouse, using menu. I can't think of any problems using this.

We're not supporting graying out cut files if done by external application (at least so far).

tsujan: > Using && hasFocus() instead works too

No it doesn't; it took me a whole hour to realize that ;) Under some WM's, a window can have a right click menu but not only it may not be the active window but also it's view may not have focus on clicking "Cut". Enlightenment is an example (elsewhere, I mentioned why it's good to have Enlightenment for finding issues).

I tested with ~5000 files and saw no difference in speed/CPU usage.

vc-01: Ok, this was wrong assumption. It has focus now but doesn't have to have in future after some ui changes or other ways of cutting files added. Your solution is more stable.

tsujan: > Ok, this was wrong assumption.

An assumption I had too. Who could know about the above-mentioned observation before using `qDebug` ;) On the other hand, on such a deep level, it's better not to rely on what WM's may do.

vc-01: (I see you answered already only after hitting F5 to refresh, this breaks discussion a little bit)

Your other WM example (Enlightenment) is another one which I didn't think of. Also my other assumption about speed might be (and maybe is) wrong. It was more about letting things be done by other parts of the system. Which may be slower or faster depending of the underlying architecture, libraries used etc. I take that back too.

Anyway, at least I tested it and it works.

tsujan: Thanks for the discussion! Actually, I wanted to call you after finding the bug but, later, I realized that the bug was introduced by *me* -- I'd suggested using of item selection for speed (in your PR).

I think we should really think about a solution to graying out items when they're cut in other apps. `QHash` has always been handy for me but I'm not sure it could be used in this case; might try something like it. Anyhow, it could be the subject of another discussion.

vc-01: // off-topic answer //

> Thanks for the discussion

Discussion of one could be even a little bit ashamed of, but anyway.

> bug was introduced by me

I don't investigate details, it's not important IMO, I remember that I pushed _selections_ too for some speed optimizations.

> solution to graying out items when they're cut in other apps

I would have already but I changed my job and after that went silent. I'm quite out of time now. I want to work too on some real issues on `lxqt-runner`, hopefully I'll find some time soon.

> `QHash` has always been handy

Maybe it should have been `QHash` from the beginning, I discovered this class only now. But important is, it's half prepared already by using hash set (of any implementation) of full paths.

agaida:
GTM

PR #129, opened 2017-11-11
Prevent a potential crash in xdndworkaround

tsujan:
The pointer `lastDrag_`, which is used by `xdndworkaround.cpp` to keep track of the last drag, may become dangling in the meanwhile (when used in `dndWorkaround::selectionRequest`) and cause a crash. To prevent that, `QPointer` is used.

NOTES:
(1) Without tab DND, there was no chance for `lastDrag_` to become a dangling pointer. However, `QPointer` is always safer.
(2) Before this patch, I could reproduce a crash only under Enlightenment (it's a good environment for finding hidden issues because of its unusual ways of handling X).

### DIFF
--- a/src/xdndworkaround.h
+++ b/src/xdndworkaround.h
@@ -52,6 +52,7 @@
 #include <QAbstractNativeEventFilter>
 #include <xcb/xcb.h>
 #include <QByteArray>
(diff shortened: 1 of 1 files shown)


agaida:
GTM

PR #130, opened 2017-11-12
Really cancel multiple renaming on cancelling

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/597

Previously, there was no easy way to cancel renaming of a great number of files because pressing Cancel/Esc just canceled renaming of the current file while that could also be achieved by pressing OK/Enter without changing the name.

A PR will follow this for renaming through pcmanfm-qt's Edit menu and also with shortcut.


agaida:
GTM, works fine

PR #131, opened 2017-11-13
Always wait for the folder to load before selecting

tsujan:
Fixes https://github.com/lxde/lxqt/issues/1407

Previously, it was supposed that if the directory path had already been set, the folder should be loaded but, in real life, that might not happen, so we should always wait for the folder to be completely loaded before selecting any file in it.


agaida:
GTM

PR #133, opened 2017-11-14
Fix two devices for one mount

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/501

Previously, items that were already added with "mount-added" would be readded with "volume-added".

### DIFF
--- a/src/placesmodel.cpp
+++ b/src/placesmodel.cpp
@@ -429,6 +429,18 @@ void PlacesModel::onMountRemoved(GVolumeMonitor* monitor, GMount* mount, PlacesM
 }
 
 void PlacesModel::onVolumeAdded(GVolumeMonitor* /*monitor*/, GVolume* volume, PlacesModel* pThis) {
+    // the item may have been added with "mount-added" (as in loopback mounting)
+    bool itemExists = false;
+    GMount* mount = g_volume_get_mount(volume);
+    if(mount) {
+        if(pThis->itemFromMount(mount)) {
+            itemExists = true;
+        }
+        g_object_unref(mount);
+    }
+    if(itemExists) {
+        return;
+    }
     // for some unknown reasons, sometimes we get repeated volume-added
     // signals and added a device more than one. So, make a sanity check here.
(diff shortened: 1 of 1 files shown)


src/placesmodel.cpp
```diff
@@ -429,6 +429,18 @@ void PlacesModel::onMountRemoved(GVolumeMonitor* monitor, GMount* mount, PlacesM
 }
 
 void PlacesModel::onVolumeAdded(GVolumeMonitor* /*monitor*/, GVolume* volume, PlacesModel* pThis) {
+    // the item may have been added with "mount-added" (as in loopback mounting)
+    bool itemExists = false;
+    GMount* mount = g_volume_get_mount(volume);
```
palinek: How about using the `GObjectPtr<GMount>`...?

tsujan: Equivalent and no real gain.

tsujan:
@agaida This PR as well as the following PR's fix serious issues: https://github.com/lxde/libfm-qt/pull/134, https://github.com/lxde/libfm-qt/pull/131, https://github.com/lxde/libfm-qt/pull/129, https://github.com/lxde/libfm-qt/pull/128. I'll make yet another PR as an important fix. Is a minor release possible in the near future?

tsujan:
By a minor release, I mean 0.12.1.

src/placesmodel.cpp
palinek: > Equivalent and no real gain.

Try to rewrite the all the libfm-qt w/o using `GObjectPtr` ...

tsujan: ?!
`GObjectPtr` wasn't used in that file, so I kept the style. At some point, all such cases should be found and changed but in a coherent way.

tsujan: A simple way of doing that could be searching for `g_object_unref`. Something for another PR....

agaida:
Why not, i'm all for.

tsujan:
> Why not, i'm all for.

That would be great. Some of these fixes were because of new reports after the last release. The more feedback are given, there better the app will become.

PR #135, opened 2017-11-15
No change queue of files in the deletion queue

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/508

Apparently, `GFileMonitor` may issue `G_FILE_MONITOR_EVENT_ATTRIBUTE_CHANGED/G_FILE_MONITOR_EVENT_CHANGED` after `G_FILE_MONITOR_EVENT_DELETED`. However, if a file is queued for change after being queued for deletion, its item might not be removed. The current patch prevents such a scenario.


agaida:
GTM

PR #136, opened 2017-11-15
Prepare libfm-qt for bulk rename

tsujan:
This will be used and followed by a PR for bulk rename in pcmanfm-qt. `changeFileName()` returns `true` only if renaming is successful and shows an error message only if needed. The changes are backward compatible.

NOTE: Bulk rename is not included in libfm-qt itself because only a file manager may need it.

### DIFF
--- a/src/utilities.cpp
+++ b/src/utilities.cpp
@@ -156,7 +156,7 @@ bool isCurrentPidClipboardData(const QMimeData& data) {
     return !clip_pid.isEmpty() && clip_pid == curr_pid;
 }
 
-void changeFileName(const Fm::FilePath& filePath, const QString& newName, QWidget* parent) {
+bool changeFileName(const Fm::FilePath& filePath, const QString& newName, QWidget* parent, bool showMessage) {
     auto dest = filePath.parent().child(newName.toLocal8Bit().constData());
     Fm::GErrorPtr err;
     if(!g_file_move(filePath.gfile().get(), dest.gfile().get(),
@@ -165,8 +165,12 @@ void changeFileName(const Fm::FilePath& filePath, const QString& newName, QWidge
                                    G_FILE_COPY_NOFOLLOW_SYMLINKS),
                     nullptr, /* make this cancellable later. */
                     nullptr, nullptr, &err)) {
-        QMessageBox::critical(parent, QObject::tr("Error"), err.message());
+        if (showMessage){
+            QMessageBox::critical(parent, QObject::tr("Error"), err.message());
+        }
+        return false;
     }
+    return true;
 }
 
 void renameFile(std::shared_ptr<const Fm::FileInfo> file, QWidget* parent) {
--- a/src/utilities.h
+++ b/src/utilities.h
@@ -53,7 +53,7 @@ LIBFM_QT_API void cutFilesToClipboard(const Fm::FilePathList& files);
 
 LIBFM_QT_API bool isCurrentPidClipboardData(const QMimeData& data);
 
-LIBFM_QT_API void changeFileName(const Fm::FilePath& path, const QString& newName, QWidget* parent);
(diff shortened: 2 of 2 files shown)


agaida:
hmm, i have no problem merging this - but it introduce a symbol change - so we should find a way to advise our users that a rebuild of pcmanfm-qt and lximage-qt migth be useful:
```diff

--- new_symbol_file (libfm-qt3_0.12.1~1-gc667692-11_amd64)
+++ dpkg-gensymbolstSyr2E       2017-11-25 22:22:31.700751206 +0100
@@ -943,7 +943,7 @@
  (c++)"Fm::VolumeManager::volumeRemoved(Fm::Volume const&)@Base" 0.12.0
  (c++)"Fm::VolumeManager::~VolumeManager()@Base" 0.12.0
  (c++)"Fm::allKnownTerminals()@Base" 0.12.0
- (c++)"Fm::changeFileName(Fm::FilePath const&, QString const&, QWidget*)@Base" 0.12.0
+#MISSING: 0.12.1~1-gc667692-11# (c++)"Fm::changeFileName(Fm::FilePath const&, QString const&, QWidget*)@Base" 0.12.0
  (c++)"Fm::copyFilesToClipboard(std::vector<Fm::FilePath, std::allocator<Fm::FilePath> > const&)@Base" 0.12.0
  (c++)"Fm::createFileOrFolder(Fm::CreateFileType, Fm::FilePath, _FmTemplate*, QWidget*)@Base" 0.12.0
  (c++)"Fm::cutFilesToClipboard(std::vector<Fm::FilePath, std::allocator<Fm::FilePath> > const&)@Base" 0.12.0
@@ -964,6 +964,7 @@
  (c++)"Fm::uidFromName(QString)@Base" 0.10.0
  (c++)"Fm::uidToName(unsigned int)@Base" 0.10.0
  (c++)"Fm::uriExists(char const*)@Base" 0.10.0
+ _ZN2Fm14changeFileNameERKNS_8FilePathERK7QStringP7QWidgetb@Base 0.12.1~1-gc667692-11
  fm_search_add_dir@Base 0.10.0
  fm_search_add_mime_type@Base 0.10.0
  fm_search_dup_path@Base 0.10.0
```

tsujan:
@agaida For me, there was no need to recompile pcmanfm-qt after this and, actually, I didn't touch lximage-qt -- it works all right. Yes, there is a symbol change but it's backward compatible, as I mentioned in the PR comment.

Anyhow, pcmanfm-qt should be updated after [bulk renaming](https://github.com/lxde/pcmanfm-qt/pull/600), which requires this.

tsujan:
Oh, that's because I don't use the debug mode in compilation -- unless a bad bug with an unclear backtrace is detected -- like the one fixed here: https://github.com/lxde/libfm-qt/pull/141 -- then, I recompile with debugging.

agaida:
```diff
- (c++)"Fm::changeFileName(Fm::FilePath const&, QString const&, QWidget*)@Base" 0.12.0
+ (c++)"Fm::changeFileName(Fm::FilePath const&, QString const&, QWidget*, bool)@Base" 0.12.1~
```
isn't exactly backward compatible i guess

agaida:
GTM

tsujan:
> isn't exactly backward compatible i guess

Right but that `bool` is `true` by default, which is the old behavior -- but I guess you already know that.

tsujan:
Long story short, I needed a way to suppress the error dialogs because, in bulk renaming, only one error dialog is needed. Hence the `bool`, which makes the error dialog optional.

agaida:
Perfectly fine with, i don't see any problems right now  - i "only" searching for a nice and working way to advise users about ... :D

tsujan:
> i "only" searching for a nice and working way to advise users about...

I think most users should already know that because of the libfm-qt history: it has undergone heavy backward-incompatible changes.

agaida:
also true

PR #140, opened 2017-11-24
No cut path with search directory

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/603

For the search directory, setting the (static) cut path has no meaning because that directory doesn't support file monitoring. On the other hand, if the cut path is set, a crash will happen on quitting because the GFile that's stored in the path is already unreferenced (how?) but `__do_global_dtors_aux()` tries to unref it again.


tsujan:
This fixes a part of a bigger issue just by chance and isn't a real solution. Closing.

PR #141, opened 2017-11-25
Track folders containing cut files only with QString

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/603

`QString` is used, instead of `Fm:FilePath`, for tracking folders with cut files.

The reason is that a static `Fm:FilePath` can cause a crash on quitting because the `GFile` that's stored in it may already have a zero reference count when `__do_global_dtors_aux()` tries to unref it. QString is completely safe in this regard.


tsimonq2:
Looks sane to me.

agaida:
GTM

PR #142, opened 2017-11-27
Copy selected pathbar text to selection clipboard

tsujan:
Closes https://github.com/lxde/pcmanfm-qt/issues/604

Also use `QTimer::singleSho()` to focus the line-edit because otherwise, it might not have time to get focused if the main window didn't have focus (this can happen with some window managers like Enlightenment's).


agaida:
GTM

PR #143, opened 2017-12-04
Fixed the logic of queued deletion

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/508.

Queueing files for deletion was done inaccurately before.

First, as was the case with the update queue (see https://github.com/lxde/libfm-qt/commit/90531712fe05e96488ccd8fcfd29a618b49eb65c and https://github.com/lxde/pcmanfm-qt/issues/506), the deletion queue should not rely on the files that are listed *at the moment* because the order of some function calls isn't fixed -- the function `core/folder.cpp → Folder::processPendingChanges()` may be called multiple times without a single call to `Folder::onFileInfoFinished()` (although that's rare).

Second, and for the same reason, the deletion should be done only *after* the file info job is finished and processed by `Folder::onFileInfoFinished()`; otherwise, the order of addition, update and deletion might be wrong and, as a side effect, nonexistent files might be shown. Therefore, `FileInfoJob` should carry the deletion paths as an additional variable.

Also removed a redundant call (not important).

### DIFF
--- a/src/core/fileinfojob.cpp
+++ b/src/core/fileinfojob.cpp
@@ -3,9 +3,10 @@
 
 namespace Fm {
 
-FileInfoJob::FileInfoJob(FilePathList paths, FilePath commonDirPath, const std::shared_ptr<const HashSet>& cutFilesHashSet):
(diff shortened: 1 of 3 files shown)


agaida:
GTM

PR #146, opened 2017-12-17
Add Group column and don't use owner's full name

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/617.

In the detailed list mode, the owner's full name doesn't seem practical, while a Group column makes the details more complete.

NOTE: pcmanfm-qt should be recompiled after this.


agaida:
GTM

PR #147, opened 2017-12-19
Fix the "Folders" key in custom actions

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/438.

Because of a small mistake, the files that were immediately inside the "Folders" list were ignored (but the files inside subfolders weren't).


agaida:
GTM

PR #150, opened 2017-12-23
Fix memory leak in thumbnail loading

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/621.

Putting a statically cached shared pointer (`std::shared_ptr<const MimeType>`) inside a vector (`Thumbnailer::mimeTypes_`) increased its use count and caused a memory leak, which was detected by valgrind. The leak is fixed by removing the vector in question -- it wasn't used anywhere.


agaida:
GTM

PR #151, opened 2017-12-24
Support hiding items in Places side-pane

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/580.

The items can be hidden by using their context menus. If at least one item is hidden, all rows -- including the root ones (Places, Devices, Bookmarks) -- will have the checkbox "Show All Entries" in their context menus, so that the user could un-hide each hidden item again.

All items can be hidden, except for bookmarks. IMO, hiding bookmarks doesn't make much sense because if they aren't needed, they could be removed.

If all children of a root item are hidden, it'll be hidden too.

Hidden items are remembered during the session. For remembering them between sessions, a pcmanfm-qt PR will follow this one.

Also, the "Empty Trash" menu-item is moved to the top for easier access (as in Dolphin) and is disabled when the trash is empty. Change your habits ;)

NOTE: pcmanfm-qt should be recompiled after this.

### DIFF
--- a/src/placesmodel.cpp
+++ b/src/placesmodel.cpp
@@ -55,34 +55,25 @@ PlacesModel::PlacesModel(QObject* parent):
 
     createTrashItem();
 
-    // FIXME: add an option to hide network:///
-    if(true) {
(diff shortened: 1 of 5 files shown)


agaida:
GTM

PR #152, opened 2017-12-25
Fixed sorting by type/owner

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/459.

Previously, when files were sorted by their types/owners, they were shown irregularly within each group. With this patch, they're sorted by their display names within groups.


agaida:
GTM

PR #154, opened 2017-12-26
Ensure that rename editor has opaque background

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/630.

Depending on how line-edits are drawn by different widget styles, their background may be transparent with some styles (Fusion and Breeze) on the side-pane because the base color of the side-pane is transparent. This patch guarantees an opaque bg with all styles.


agaida:
GTM

PR #156, opened 2017-12-27
Drop Q_FOREACH

luis-pereira:
Replaced by the ranged for.
Q_FOREACH will be deprecated soon. Enforced with the Q_NO_FOREACH
definition.


tsujan:
I add that all cases *without* `static_cast` are correct.

luis-pereira:
@tsujan Ignore the `static_cast`. It's `const_cast` obviously. Thanks for catching it.

tsujan:
> Thanks for catching it.

Actually, I didn't ;) --  just checked the lines without it.

tsujan:
And am I wrong, thinking that `static_cast` is correct?

luis-pereira:
@tsujan Correct, yes.  The `const_cast` is more appropriated in this use case. We don't need a `static_cast` and a `const_cast` allows the compiler to catch more errors.

luis-pereira:
@tsujan What I mean is: if a `const_cast` is enough (in this case it is) there's no need to use a `static_cast`.

tsujan:
@luis-pereira I think `static_cast` is a little safer in general -- but, perhaps, not here,. However, if `const_cast` catches more errors, I'll have no objection to it either.

luis-pereira:
@tsujan `const_cast` catches more errors, no doubt.
PR updated.

palinek:
How about requiring Qt v5.7 and using the `qAsConst()`?

tsujan:
> How about requiring Qt v5.7 and using the qAsConst()?

It came to my mind too -- when the container can't be declared as const -- but later thought, "they deprecated `Q_FOREACH`, so they might do the same thing with `qAsConst`.."

luis-pereira:
If requiring Qt 5.7 is fine from a packagers/distro perspective, I'm fine with it.
On the technical side `qAsConst()` is syntactic sugar. It performs a `const_cast<const T&>`. `qAsConst()` is the Qt implementation of the new C++17 `std::as_const()`. When `std::as_const()` becomes available in Qt's targets compilers, it will most likely be deprecated. But that may take a while. When it becomes deprecated the transition will be simple because `qAsConst()` implementation follows `std::as_const()` on purpose.

agaida:
> https://packages.debian.org/stretch/qt5-default
Even debian stable has 5.7.1 - so i would give it a go

luis-pereira:
So... let's go with `qAsConst()` and requiring Qt 5.7. Comments wellcome.

tsujan:
@luis-pereira As you mentioned, when it's going to be deprecated, it'll be easy to replace it. On the other hand, `std::as_const()` may not be available everywhere yet. So, I agree.

palinek:
> So... let's go with qAsConst() and requiring Qt 5.7. Comments wellcome.

Full ack.

agaida:
Full ack.

src/filesearchdialog.cpp
```diff
@@ -144,7 +144,8 @@ void FileSearchDialog::onAddPath() {
 
 void FileSearchDialog::onRemovePath() {
     // remove selected items
-    Q_FOREACH(QListWidgetItem* item, ui->listView->selectedItems()) {
+    const auto itemList = ui->listView->selectedItems();
+    for(QListWidgetItem* const item : qAsConst(itemList)) {
```
tsujan: `qAsConst()` doesn't seem needed here.

src/dirtreemodelitem.cpp
```diff
@@ -343,7 +343,7 @@ DirTreeModelItem* DirTreeModelItem::childFromName(const char* utf8_name, int* po
 DirTreeModelItem* DirTreeModelItem::childFromPath(Fm::FilePath path, bool recursive) const {
     Q_ASSERT(path != nullptr);
 
-    Q_FOREACH(DirTreeModelItem* item, children_) {
+    for(DirTreeModelItem* const item : const_cast<const std::vector<DirTreeModelItem*>&>(children_)) {
```
tsujan: Shouldn't this be like `hiddenChildren_` above?

luis-pereira:
@tsujan Updated. Thanks.

tsujan:
Very GTM.

Issue #158, opened 2018-01-04
lambda in selectFilePathWithDelay maybe called even after object destructed


comicfans:
This bug is reported [here](https://bugs.kde.org/show_bug.cgi?id=387334),

when Qt application call 
QFileDialog::setOptions, it may trigger following code path:

LXQtFileDialogHelper::~LXQtFileDialogHelper() (this=0x6040002b1d10)
    at /srv/nfs4/working/lxqt-qtplugin/src/lxqt-qtplugin-0.12.0/src/lxqtfiledialoghelper.cpp:39
#1  0x00007ffff54a7dc6 in QDialogPrivate::deletePlatformHelper() () at /usr/lib/libQt5Widgets.so.5
#2  0x00007ffff54b9b19 in  () at /usr/lib/libQt5Widgets.so.5
#3  0x00007ffff54b9d69 in QFileDialog::setOptions(QFlags<QFileDialog::Option>) () at /usr/lib/libQt5Widgets.so.5

this will delete its _dlg member (which is a FileDialog instance)

after this call, if client call filedialog->exec before control flow returned to main loop,
lambda connected inside selectFilePathWithDelay will still be called , and it leads crash.

manually add processEvents can workaround this, so it seems that when dialog exec is called before previous main loop returned, QTimer connection didn't got auto disconnected when receiver destructed.

FileDialog construct backtrace:
0x00007fffdcef400c in LXQtFileDialogHelper::LXQtFileDialogHelper() (this=0x6040002b1d10)
    at /srv/nfs4/working/lxqt-qtplugin/src/lxqt-qtplugin-0.12.0/src/lxqtfiledialoghelper.cpp:22
#1  0x00007fffdcef5365 in non-virtual thunk to LXQtPlatformTheme::createPlatformDialogHelper(QPlatformTheme::DialogType) const ()
    at /usr/include/qt/QtCore/qcoreapplication.h:116
#2  0x00007ffff54a7b20 in QDialogPrivate::platformHelper() const () at /usr/lib/libQt5Widgets.so.5
#3  0x00007ffff54ba3fd in QFileDialogPrivate::init(QUrl const&, QString const&, QString const&) () at /usr/lib/libQt5Widgets.so.5
#4  0x00007ffff54ba822 in QFileDialog::QFileDialog(QWidget*, QString const&, QString const&, QString const&) () at /usr/lib/libQt5Widgets.so.5
[...]

PR #159, opened 2018-01-04
Fix lambda connections in filedialog

tsujan:
Fixes https://github.com/lxde/libfm-qt/issues/158 by adding "this" to `QTimer::singleShot()` lambdas.


palinek:
GTM

PR #160, opened 2018-01-06
Don't drop on files

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/639.

Instead of trying to drop on a file and showing an error message afterward, drop into its containing folder.

Also commented out `qDebug` messages.


agaida:
GTM

PR #162, opened 2018-01-17
Fixed custom action execution mode

tsujan:
The variable `exec_mode` wasn't initialized when the action lacked "ExecutionMode" (which is the case most of the time).

Strange that, previously, it always worked, although it shouldn't! Only today I saw an effect of this problem (which wasn't a crash).


agaida:
GTM

PR #165, opened 2018-01-30
 iconinfo: Properly handle multiple names

palinek:


palinek:
Warning: this breaks ABI & API

tsujan:
This seems interesting but should be checked and tested under various circumstances. I hope there's no haste here.

palinek:
> should be checked and tested under various circumstances

Definitely...

tsujan:
This and, especially, the removal of `IconTheme` looks more logical than before.

GTM.

agaida:
GTM

PR #168, opened 2018-03-22
Elided labels for file operation dialog

tsujan:
Fixes https://github.com/lxde/pcmanfm-qt/issues/578.

This prevents the dialog from widening when a file name/path is too long; the user could resize it at will.


agaida:
GTM

PR #171, opened 2018-03-24
Italic font for hidden items

tsujan:
Closes https://github.com/lxde/pcmanfm-qt/issues/632.

To distinguish hidden items visually, just make their texts italic.

NOTE: Since we follow `GLib`, hidden files aren't limited to those whose names start with a dot. For example, `libfm-qt` rightly hides `lost+found` too.


agaida:
Looks fine, GTM

PR #177, opened 2018-04-28
Fully port `deleting files' and `moving files to trash' to C++.

PCMan:
Previously, all file I/O jobs uses the C code from libfm.
I started to port the jobs to the new C++ API one by one.
This PR contains ports for file deletion and moving to trash.


tsujan:
I'm very glad :) Thanks a lot!

PCMan:
@tsujan Finally have some time to do the remaining parts. There are still many things to do. I will submit the PRs in small change sets for ease of review.
Thanks.

PR #181, opened 2018-05-01
Fully port all file operations to C++ 11

PCMan:
Previously, all file I/O jobs used the C code from libfm.
Now all of the file operations are fully ported to C++.
This PR contains C++ implementations for these jobs:

1. delete files
2. move to trash.
3. restore from trash
4. copy files
5. move files
6. change file attributes

Also I change the "Data transferred:" in the file operation dialog to "Files processed:".
When data size cannot be calculated, we show number of files processed instead.


PCMan:
@tsujan I integrate all file jobs in one single PR so intermediate changes that were removed later do not need reviews.

tsujan:
Yes, I saw it. That's better.

PR #182, opened 2018-05-01
Port archiver integration to C++

PCMan:


src/filemenu.cpp
```diff
@@ -226,7 +226,7 @@ FileMenu::FileMenu(Fm::FileInfoList files, std::shared_ptr<const Fm::FileInfo> i
                         addAction(action);
                     }
                 }
-                else {
+                else if(archiver->canCreateArchive()){
```
PCMan: @tsujan the only thing changed is this line. I only show the "Compress" menu item if the archiver supports compression.
Would you like to help take a look at this to see if your archiver setting is correctly loaded?

PCMan: @tsujan  If the value of `archiver` pointer is null, then the setting is not correctly loaded.

tsujan:
However, this isn't a blocker, IMO. These nice changes can be merged as they are. Other works could be done later.

agaida:
yes, please push the button

tsujan:
Another step closer to independence from libfm, thanks to @PCMan's magic :)

PR #188, opened 2018-05-11
Pcman@port search uri to c++

PCMan:
FmSearch class is still implemented in C / GObject.
However this PR removes all references to libfm FmPath in this part.
Later I can port the whole FmSearch stuff to C++.


tsujan:
@agaida After an update, I get those cast-between-incompatible-function-types warnings too. Will look into them later.

agaida:
@tsujan - thanks

PR #192, opened 2018-05-20
Fixed shortcut detection

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/676

This old problem didn't show up until recent C++ ports. It was "handled" before but in a wrong way.


agaida:
Good catch :)

tsujan:
> Good catch :)

I should correct Mark Twain: All you need in this life is ignorance and confidence. Then just wait for luck and success will be sure.

PR #194, opened 2018-05-24
Optionally show full name

tsujan:
Instead of display name. To be followed by a pcmanfm-qt patch.

### DIFF
--- a/src/foldermodel.cpp
+++ b/src/foldermodel.cpp
@@ -35,7 +35,8 @@
 namespace Fm {
 
 FolderModel::FolderModel():
-    hasPendingThumbnailHandler_{false} {
+    hasPendingThumbnailHandler_{false},
+    showFullNames_{false} {
 }
 
 FolderModel::~FolderModel() {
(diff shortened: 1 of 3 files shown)


tsimonq2:
Looks good to me.

tsujan:
@agaida Please decide this and its pcmanfm-qt counterpart. I want to implement "shadowed hidden icons"  (PCMan's disabled option) but there are common files -- don't like rebasing ;)

PCMan:
@tsujan I'd suggest renaming the methods to imply what they actually do.
What does "full name" mean?

PCMan:
showRealFileName() or showRawFileName() might be more meaningful I guess.
However, this will cause problems as the real filename might not be in UTF-8 encoding. That's why we have displayName which is always valid unicode. So QString::fromStdString() will fail sometimes. (Although we already assumed UTF-8 in many places, which is not correct either)

tsujan:
@PCMan 

(1) I borrowed that name from *you*. You'd used it in pcmanfm-qt and I found it good enough.

(2) The display name is always used as fallback.

tsujan:
As for the possibility of failure in the case of `QString::fromStdString()`, you're right but, IMO, such exceptions are acceptable because this is optional and not the default.

PCMan:
Oops... The naming is from my code. Sorry for the bad naming. LOL
I should have used a better name. :-(

tsujan:
@PCMan I could change all instances of "fullName" later (to "rawName", for example). I don't think it's a blocker. I think you used "fullName" because of desktop entries and that's quite acceptable.

PCMan:
LGTM

tsujan:
Merging it because it has no conflict with https://github.com/lxqt/libfm-qt/pull/196

PR #195, opened 2018-05-30
port bookmarks and folder config to c++

PCMan:
1. Remove the use of libfm FmBookmarks.
2. Port Fm::Config to C++

PR #196, opened 2018-05-30
Fully drop dependency on libfm C library.

PCMan:
Finally, the last piece to the full C++ port.
C code that is hard to port is copied and put into src/core/legacy.
Gio VFS implementations for menu:// and search:// are sitll in C and put in src/core/vfs.
Cheers!

This definitely requires much testing, though.


PCMan:
@tsimonq2 With this change, libfm is no longer a dependency of libfm-qt. So you will need to update your deb package for the next libfm-qt release. :-)

tsujan:
@PCMan Thanks you very much!

PCMan:
@tsujan yes I know. Let's merge #195 first, and I'll rebase this one on the new master.

agaida:
BTW: this so-name thing is the reason why the new libfm-qt and dependend packages are not in sid yet - they are in experimental, but the builders for experimental run with far lower priority - and that means - hopefully the packages for the release arches are built within the next two days, then i can add/fix the changed symbols and migrate libfm-qt/lxqt-qtplugin/pcmanfm-qt/lximage-qt to sid

src/CMakeLists.txt
```diff
@@ -126,7 +137,10 @@ install(EXPORT
 target_link_libraries(${LIBFM_QT_LIBRARY_NAME}
     Qt5::Widgets
     Qt5::X11Extras
-    ${FM_LIBRARIES}
+    ${GLIB_LIBRARIES}
```
palinek: Why aren't we using the `*_LDFLAGS` ? This caused problems with libraries installed in non-standard paths problems in other components in the past...

PCMan: @palinek our FindGLIB.cmake does not define LDFLAGS:
https://github.com/lxqt/lxqt-build-tools/blob/master/cmake/find-modules/FindGLIB.cmake

palinek: Sorry,... forgot about that... I thought we're using just the pkg-config...

luis-pereira:
Building and testing it.

tsujan:
I started to use this and https://github.com/lxqt/pcmanfm-qt/pull/684. So far so good. Will report probable issues here.

Of course, it needs rebasing.

PCMan:
rebased on the current master.

PR #197, opened 2018-05-31
Disconnect old source model completely

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/685

This may be a workaround for a bug in Qt 5.11.


PCMan:
@tsujan  Thanks for the workaround. I, however, am wonder if this is a bug of Qt since it should not happen.
FYI: http://code.qt.io/cgit/qt/qtbase.git/tree/src/corelib/itemmodels/qabstractproxymodel.cpp#n132
When you set a new source model, Qt does disconnect from the old model's "destroyed" signal so I'm not sure why it's still invoked sometimes.
When the old model is destroyed, the slot is still called and the proxy model resets its content even when a new model is correctly set. This is the cause of the bug but in theory, this should not happen according to the above source code.
So while the workaround works, this is still pretty weird. If it's a Qt bug, we should make a bug report to them.

tsujan:
@PCMan I also think it's a Qt bug. I hope this workaround is backward compatible.

As for reporting it, I've reported several bugs, some of which were important, but I rarely received anything other than denial or procrastination. Moreover, I don't know of a simple code for reproducing the bug (they will want it).

tsujan:
@Chiitoo found that problems started to happen after this change: http://code.qt.io/cgit/qt/qtbase.git/patch/?id=1c0fcbc887459d8963088309e83303eb1a7d2db0

PCMan:
@tsujan Awesome! How did you find that. Now I'm sure it's a Qt bug.
See the diff please. They call "_q_sourceModelDestroyed()" manually before calling "QAbstractProxyModel::setSourceModel()". This causes the bug.
After the call, the model is set to "QAbstractItemModelPrivate::staticEmptyModel()" and when calling "QAbstractProxyModel::setSourceModel()" it tries to disconnect from old model:
`disconnect(d->model, SIGNAL(destroyed()), this, SLOT(_q_sourceModelDestroyed()));`
But d->model has been changed to staticEmptyModel() so the actual old model object is never disconncected.
It's definitely a Qt bug and we should report that.
For your PR, I'd suggest that you only disconnect the "destroyed()" signal to minimize the potential side effects.

tsujan:
> How did you find that.

@Chiitoo did. He sometimes uses unstable Qt and tracks its changes.

> It's definitely a Qt bug and we should report that.

Would you please do so?

> I'd suggest that you only disconnect the "destroyed()" signal

I'll experiment with that and will tell you about the result. Thanks!

PCMan:
It's easy to trigger the Qt bug with a simple test case.
```
#include <QSortFilterProxyModel>
#include <QStandardItemModel>
#include <assert.h>

int main(int argc, char** argv) {
    auto proxyModel = new QSortFilterProxyModel();
    auto model1 = new QStandardItemModel();
    auto model2 = new QStandardItemModel();
    proxyModel->setSourceModel(model1);
    proxyModel->setSourceModel(model2);
    delete model1;
    assert(proxyModel->sourceModel() == model2);
    return 0;
}
```
Let me create a bug report. :)

PCMan:
Bug reported to Qt devs: https://bugreports.qt.io/browse/QTBUG-68581
@tsujan @agaida Let's follow the issue to see if the Qt folks fixes it.

tsujan:
The only thing I could do was voting up.

Until it's fixed, we should have our workaround; otherwise, pcmanfm-qt won't be usable. I'll test @PCMan's suggestion soon.

tsujan:
@PCMan Disconnecting the `destroyed()` signal was no enough -- the problem persisted with it.

I think this patch is safe as it is because we don't need the old source model.

Waiting for @agaida's test with Qt 5.10 to see if it doesn't break anything there.

PCMan:
@tsujan The disconnect trick works for me. See this example please.
```
#include <QSortFilterProxyModel>
#include <QStandardItemModel>
#include <assert.h>

int main(int argc, char** argv) {
    auto proxyModel = new QSortFilterProxyModel();
    auto model1 = new QStandardItemModel();
    auto model2 = new QStandardItemModel();
    proxyModel->setSourceModel(model1);

    // workaround the Qt bug
    QObject::disconnect(model1, SIGNAL(destroyed()), proxyModel, SLOT(_q_sourceModelDestroyed()));

    proxyModel->setSourceModel(model2);
    delete model1;
    assert(proxyModel->sourceModel() == model2);
    return 0;
}

```

src/proxyfoldermodel.cpp
```diff
@@ -52,17 +52,19 @@ ProxyFolderModel::~ProxyFolderModel() {
 void ProxyFolderModel::setSourceModel(QAbstractItemModel* model) {
     if(model == sourceModel()) // avoid setting the same model twice
         return;
+    FolderModel* oldSrcModel = static_cast<FolderModel*>(sourceModel());
+    if(oldSrcModel) {
+        oldSrcModel->disconnect(this); // needed with Qt 5.11
```
PCMan: I tested this fix and it works.
```
disconnect(oldSrcModel, SIGNAL(destroyed()), this, SLOT(_q_sourceModelDestroyed()));
```

agaida:
@tsujan @PCMan - current patch work for me, waiting for the next one :P

tsujan:
@PCMan IMHO, it's better to avoid `_q_sourceModelDestroyed` in the code. `disconnect(oldSrcModel, SIGNAL(destroyed()));` doesn't work. Do you know of a scenario, where the current change could cause a problem?

PCMan:
@tsujan at the moment, disconnecting all slots works without problems, but it's much more dangerous than disconnecting the required.slot since you affect many unrelated code, including our own slots.
The order of disconnect matters. In side proxymodel, it might rely on some slots being triggered before disconnecting them. With your change, they can never be triggered.
Currently there is no such case but in the future it's possible. So don't change the behavior of unrelated parts.
Please only disconnect the q_sourceDestroyed slot and add qt version checks so the workaround only applies to Qt >= 5.11.

tsujan:
@PCMan OK, I'll try `_q_sourceModelDestroyed()` soon.

tsujan:
One word about Qt version check: With Qt 5.11, `#if QT_VERSION >= 0x051100` doesn't work ;) But `#if (QT_VERSION >= QT_VERSION_CHECK(5,11,0))` works.

tsujan:
Done.

@agaida Since Qt < 5.10 is excluded, I think this can be merged now.

tsujan:
@agaida I also think a minor release (for libfm-qt?) may be needed. Qt-5.11 users that don't have git LXQt couldn't use pcmanfm-qt without the workaround.

agaida:
yes, we should care about some unknown bleeding edge distributions - but that would take away their fun discovering new and shiny bugs :D Will prepare it and release it tomorrow.

tsujan:
Apparently, the bug is fixed in Qt-5.11.1. We should wait to know what problems 5.11.1 will cause.

agaida:
@tsujan - so - what do you suggest - wait for a few days or fix it and be done with this special bug?

tsujan:
@agaida I have no problem with waiting. If it's really fixed in Qt-5.11.1, the workaround won't be needed. If not or if Qt-5.11.1 creates more problems, the workaround could be changed. But that's just a thought. You know distros much better than I do.

agaida:
hmm - biased - right now arch and maybe derivtives are bitten - the far bigger problem i see in debian and ubuntu, right now they are in the process of packaging Qt 5.11 - and my wild guess is that it will go through experimental and than will hit sid and ubuntu cosmic nearly unchanged - and the next point will be leaved out or hit sid in july - if it will be packed at all - maybe they will wait until .2

So i guess it will be better to wait a day or two and bring a point at weekend. The changelog is short, only one fix - and done. And nobody can say that we don't care.

PCMan:
@tsujan I did not notice that my bug report is duplicated.
It has been reported and fixed.
https://bugreports.qt.io/browse/QTBUG-67948
So the workaround is only needed for version between 5.11.0 and 5.11.1.
Please add version check so the woarkound only applies to versions within this range.
Then it does not matter whether you disconnect from all slots or just "destroyed()" as long as it won't break 5.11.0 since it won't affect future versions.

tsujan:
@PCMan We know that (see [here](https://github.com/lxqt/libfm-qt/pull/197#issuecomment-393643452)). The merge is postponed. I'll change the version check later.

agaida:
Damn, you guys know how to have fun - and i know why i created a release branch, where i can apply this patch to :D

tsujan:
@agaida The patch is OK now. The bug is supposedly fixed in 5.11.1 and the change is limited to 5.11.0.

agaida:
fine, will merge it now and make a release today (but not now, it's 01:01 in the morning.

Chiitoo:
I can confirm that http://code.qt.io/cgit/qt/qtbase.git/patch/?id=e15fc26e9fdbff141890a3e2e8dc4ef935d022a0 seems to at least fix the issues I reported here.

Knew I should have opened a Qt Bug for it (just didn't get to it).  Fix might have even got in 5.11.0 already.

Oh well.  Next time.  :]

agaida:
rebased from release branch

PR #198, opened 2018-06-01
Disconnect old source model

agaida:
Recreated for the release branch 0.13.x


agaida:
@tsujan - i guess it should work that way

PR #204, opened 2018-06-05
Optionally, shadow hidden icons

tsujan:
Instead of a trasnlucent painter, the disabled pixmap is used for shadowing because the former is used for cut files. The text is left enabled for hidden items not to look totally disabled.

This will be followed and used by a pcmanfm-qt patch.


agaida:
GTM

PR #207, opened 2018-06-07
Check if the opening app exists before using it

tsujan:
Also, show an error message if there's no app.

### DIFF
--- a/src/core/basicfilelauncher.cpp
+++ b/src/core/basicfilelauncher.cpp
@@ -256,18 +256,25 @@ FilePath BasicFileLauncher::handleShortcut(const FileInfoPtr& fileInfo, GAppLaun
                 || strcmp(scheme.get(), "trash") == 0
                 || strcmp(scheme.get(), "network") == 0
                 || strcmp(scheme.get(), "computer") == 0) {
-            return FilePath::fromUri(fileInfo->target().c_str());
+            return FilePath::fromUri(target.c_str());
         }
         else {
             // ask gio to launch the default handler for the uri scheme
-            GAppInfoPtr app{g_app_info_get_default_for_uri_scheme(scheme.get()), false};
-            FilePathList uris{FilePath::fromUri(fileInfo->target().c_str())};
-            launchWithApp(app.get(), uris, ctx);
+            if(GAppInfoPtr app{g_app_info_get_default_for_uri_scheme(scheme.get()), false}) {
+                FilePathList uris{FilePath::fromUri(target.c_str())};
+                launchWithApp(app.get(), uris, ctx);
+            }
+            else {
+                GErrorPtr err{G_IO_ERROR, G_IO_ERROR_FAILED,
+                              QObject::tr("No default application is set to launch '%1'")
+                              .arg(target.c_str())};
+                showError(nullptr, err);
+            }
         }
     }
     else {
         // see it as a local path
-        return FilePath::fromLocalPath(fileInfo->target().c_str());
+        return FilePath::fromLocalPath(target.c_str());
     }
     return FilePath();
 }


agaida:
Expected:

double click on IDEFIX - get the shares

password dialog should appaer if there are no credentials

apply the credentials

open a folder

open the selected file

tsujan:
I'll check the differences between libfm and libfm-qt in this regard but I can't test smb.

tsujan:
And I love that dog ;)

tsujan:
This PR is correct but not related to the problem it was supposed to solve.

@PCMan, please help! `Windows Network` under `network:///` is seen as `inode/x-shortcut`, which is wrong. Previously, it was seen as `inode/directory` and was mounted. There is a difference between libfm and libfm-qt in this regard, partly in `fileinfo.cpp` and partly in `basicfilelauncher.cpp`.

tsujan:
@PCMan Thanks! If it helps, the first difference I saw was about these lines in `fileinfo.cpp`:
```c++
    tmp = g_file_info_get_content_type(inf.get());
    if(!tmp) {
        tmp = "application/octet-stream";
    }
    mimeType_ = MimeType::fromName(tmp);
```
According to libfm, they could have been:
```c++
    tmp = g_file_info_get_content_type(inf.get());
    if(tmp) {
        mimeType_ = MimeType::fromName(tmp);
    }
```
Then, this may need to be added at the very end, when all mime-types are checked:
```c++
    if(!mimeType_) {
        mimeType_ = MimeType::fromName("application/octet-stream");
    }
```

tsujan:
@agaida The current nullity check and error message are needed but unrelated to the problem you reported; libfm had them too.

tsujan:
> she passed away in spring this year

Sad! A 15-year life span seems good though.

agaida:
@PCMan - i 've no problems with any regression, nothing to be sorry for - indipendence from lbfm far outwight it in short- and long-term tbh. Ok, some users might see this different, with my user hat on me too :P.

tsujan:
@agaida I'm pretty sure that @PCMan will fix it soon. The solution is in his libfm work but I don't know how to transfer it.

Related to this: computer:/// → File System was `inode/directory` before; it's `inode/mount-point` now. With the above-mentioned change in `fileinfo.cpp`, it'll become `inode/directory` again but won't be opened with the current code.

agaida:
sorry, i can't help here - but i'm sure there will be a solution soon - not a big problem right now, only uploaded it to siduction and will apply your patch  - network related things will not work, but at least it don't crash anymore and there will be a nice message popping up. Until solved libfm-qt/pcmanfm-qt will stay in debian/experimental - there is enough work to do fixing the changed symbols, so nearly no harm is done. Ok, Arch and derivative users might see this different.

tsujan:
@agaida  Whenever you have time, please test https://github.com/lxqt/libfm-qt/commit/716e18ae3c5bf23a11e699ec5a5b2ac11fd4b175 ([experiment](https://github.com/lxqt/libfm-qt/tree/experiment) branch).

agaida:
ok - i would call it a step in the right direction:


So, the available machines are displayed correctly.

tsujan:
> But i can't open them

Because there was no mounting ;) What about now: https://github.com/lxqt/libfm-qt/commit/310ba1eef1513227aa452892cf817b39059a2fcd  ?

agaida:
ok, just a second - was on another thing: https://github.com/lxqt/lxqt/issues/1495#issuecomment-395240855

agaida:
ok, just a second - was on another thing: https://github.com/lxqt/lxqt/issues/1495#issuecomment-395240855

ok, there we are

tsujan:
But that's the first dialog above. Are you sure you tried https://github.com/lxqt/libfm-qt/commit/310ba1eef1513227aa452892cf817b39059a2fcd ?

agaida:
no - sorry, my mistake - the current "kind of works"  -  but it is hard to describe

tsujan:
@agaida There's no haste in this. If you're doing another job, I don't want to bother you with this.

agaida:
you don't bother me :) - it's only hard to describe, i dont want to be misleading. Ok, lets start.

1. kill all pcmanfm-qt processes and start it from terminal
2. Single klick on network

3. double click on IDEFIX

after a short meditation local root is opend
4. try open with pcmanfm-qt

voila, something is mounted
5. click ok, the next message appears

6. click on the message - the message disapper, a delete MountOperation is logged


little bit boring, but now to the interesting part - type in the address directly and hit enter

hmm - thats really IDEFIX :)

double click on photos bring up the password dialog
 
log in - we are back to local root :8 - but one difference - the selected folder is now mounted :)

it turns out, that the mounted folder "photos" is really an opened smb-share:

normal operations like open folders or display files is fine


Unmount the mounted share results in a closed pcmanfm-qt - somewhat call the quit routine within the unmount operation.


At all: nice start and a step forward.

tsujan:
So, in short, the result is a better mess :D I tried to do what was done in libfm but some computations should still be different.

agaida:
Yesterday we were one step ahead of the abyss. Today we have taken a decisive step forward. :sunglasses:

tsujan:
> after a short meditation local root is opend

That's the first problem.

> click on the message - the message disapper, a delete MountOperation is logged

Which is the second problem.

tsujan:
@agaida I made a small change by consulting libfm (replaced `FilePath::fromUri` with `FilePath::fromPathStr`); the patch is this now: https://github.com/lxqt/libfm-qt/commit/d2e08c38edcbd116c7cafdeb1c23e2d36a2b8d17

PCMan:
GTM. Failing to open smb:// is another issue. Let's not mixing them. I merge this one first.

PR #208, opened 2018-06-07
Update from Weblate.

lxdeadmin:


palinek:
@agaida Why lxdeadmin ?

agaida:
i need a proper git user - and havn't found the time to switch to a better one :P

PR #210, opened 2018-06-09
Fix failure to open smb:// caused by incorrect file info handling.

PCMan:
The issue is caused by multiple problems.

1. in libfm, shortcuts targeting smb:// is also detected as a dir but libfm-qt does not do this
2. when smb:// is first accessed, gvfs will raise the error: path is not mounted. In libfm it just treat all unmounted paths as directories, which is not correct but can workaround the smb:// case. With the hack, smb:// can be detected as dir correctly even when it's unmounted. Libfm-qt does not contain the hack.

Instead of adding the hack, I implement error handling for the "not mounted" error so smb:// can be mounted automatically upon the first access. This should IMO be the more correct behavior.

### DIFF
--- a/src/core/basicfilelauncher.cpp
+++ b/src/core/basicfilelauncher.cpp
@@ -29,11 +29,13 @@ bool BasicFileLauncher::launchFiles(const FileInfoList& fileInfos, GAppLaunchCon
     FilePathList pathsToLaunch;
     // classify files according to different mimetypes
     for(auto& fileInfo : fileInfos) {
-        // qDebug("path: %s, target: %s", fileInfo->path().toString().get(), fileInfo->target().c_str());
-        if(fileInfo->isDir()) {
-            folderInfos.emplace_back(fileInfo);
-        }
-        else if(fileInfo->isMountable()) {
+        /*
+        qDebug("path: %s, type: %s, target: %s, isDir: %i, isDesktopEntry: %i",
+               fileInfo->path().toString().get(), fileInfo->mimeType()->name(), fileInfo->target().c_str(),
+               fileInfo->isDir(), fileInfo->isDesktopEntry());
+        */
+
+        if(fileInfo->isMountable()) {
             if(fileInfo->target().empty()) {
                 // the mountable is not yet mounted so we have no target URI.
                 GErrorPtr err{G_IO_ERROR, G_IO_ERROR_NOT_MOUNTED,
@@ -66,6 +68,9 @@ bool BasicFileLauncher::launchFiles(const FileInfoList& fileInfos, GAppLaunchCon
             if(path.isValid()) {
                 pathsToLaunch.emplace_back(path);
             }
+        }
+        else if(fileInfo->isDir()) {
+            folderInfos.emplace_back(fileInfo);
         }
         else {
             auto& mimeType = fileInfo->mimeType();
@@ -103,16 +108,27 @@ bool BasicFileLauncher::launchFiles(const FileInfoList& fileInfos, GAppLaunchCon
 bool BasicFileLauncher::launchPaths(FilePathList paths, GAppLaunchContext* ctx) {
     // FIXME: blocking with an event loop is not a good design :-(
     QEventLoop eventLoop;
-
     auto job = new FileInfoJob{paths};
     job->setAutoDelete(false);  // do not automatically delete the job since we want its results later.
 
     GObjectPtr<GAppLaunchContext> ctxPtr{ctx};
+
+    // error handling (for example: handle path not mounted error)
+    QObject::connect(job, &FileInfoJob::error,
+            &eventLoop, [this, job, ctx](const GErrorPtr & err, Job::ErrorSeverity /* severity */ , Job::ErrorAction &act) {
+        auto path = job->currentPath();
+        if(showError(ctx, err, path, nullptr)) {
+            // the user handled the error and ask for retry
+            act = Job::ErrorAction::RETRY;
+        }
+    }, Qt::BlockingQueuedConnection);  // BlockingQueuedConnection is required here to pause the job and wait for user response
+
     QObject::connect(job, &FileInfoJob::finished,
             [&eventLoop]() {
         // exit the event loop when the job is done
         eventLoop.exit();
     });
+
     // run the job in another thread to not block the UI
     job->runAsync();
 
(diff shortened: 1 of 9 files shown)


tsujan:
@PCMan `Computer` → `File System` is correctly seen as `inode/directory` but it can't be opened now (`the specified directory 'computer:///root.link' is not valid` + `file monitor cannot be created: Not supported`).

It can be opened with https://github.com/lxqt/libfm-qt/commit/658642a724ba60ab81ae835a3a871a44b34be44a (if it helps).

PCMan:
@tsujan Ouch! Let me take a look at computer:///.

tsujan:
@PCMan The mount points under Computer are all OK. It's just that `File System` isn't opened; it isn't a mount point anymore -- and that's correct.

PCMan:
@tsujan fixed. I made root.link a mount point. It's also a dir, too.

PCMan:
@tsujan Thanks for the catch. Fixed. It's really bad to use goto :-(
For this part, previously I just copied the old C code from libfm and convert them to C++. Newly written C++ code should not contain goto anymore.

tsujan:
@PCMan Thanks!

IMHO, `goto` is a good thing when needed and, especially, I like the way you used it in libfm.

tsujan:
I confirm the the issue with `File System` is fixed while it's `inode/directory` :)

I can't test `smb`. The crucial test will be @agaida's.

@agaida You'll need to recompile pcmanfm-qt after this.

tsujan:
@HelixBot == @agaida ?

tsujan:
And I was happy that @HelixBot joined LXQt... I hope you don't have another username, like shapeshifter_at_github.

yan12125:
Based on my testing, this also fixes connecting to an SSH/SFTP server via Go -> Connect to remote server in pcmanfm-qt!

tsujan:
@yan12125 That's was very good news. Did/could you test `smb` too?

yan12125:
Nope as I rarely use SMB. Just tried it - clicking on a random SMB share and a dialog jumps out asking credentials. I assume that's the desired behavior? I don't own an SMB server so I can't test further.

tsujan:
@yan12125 That's a good sign.

agaida:
GTM

tsujan:
@agaida I also had to recompile `lxqt-qtplugin` after this (no icon, no widget style). I think you also need to do so with Debian.

tsujan:
... Or maybe after another merged PR; I'm not sure.

agaida:
don't forget about lximage-qt :P 
https://qa.debian.org/developer.php?login=pkg-lxqt-devel@lists.alioth.debian.org <-- the experimental ones

tsujan:
Oh, yes, lximage-qt definitely needs a recompilation too.

agaida:
don't forget about lximage-qt :P 
https://qa.debian.org/developer.php?login=pkg-lxqt-devel@lists.alioth.debian.org <-- the experimental ones

agaida:
ok, all rebuilds went fine - but please - where i find any changed behaviour?

tsujan:
@agaida Are your IDEFIX folders opened correctly now under Network? I'm also curious to know what you meant above by "more or less fine"?

agaida:
i will make a screencast :P

tsujan:
> i will make a screencast :P

Which means there's still a problem.

agaida:
yes - but i have to double check it in Arch.

agaida:
https://youtu.be/8RADzce-cMo - click on the network things just open /  - but connect to server is right - the ssh part without login

agaida:
https://www.youtube.com/watch?v=YxyaK81UDj8 - connect to server - ssh - ok, the mount is nearly right, target dir not. Fallback and display / isn't right too.

agaida:
now to smb:// - typing in smb://$server works fine, shares are displayed - when one connect to a share - / is shown again, but the mount itself is fine. Additional question - why there isn't smb/cifs available in connect to server?  https://www.youtube.com/watch?v=krgvrOYYfKA

tsujan:
I knew we should wait until you test it but you merged it yourself ;)

@PCMan ?

agaida:
why not? the now current state is clearly an improvement - and it build and didn't segfault.

tsujan:
> the now current state is clearly an improvement

It is, definitely. But the problem you've encountered is important. I remember that you saw a similar issue with my patch too. Both may be caused by a little bug hiding somewhere.

johndoe71rus:
hi. do you fix "click on the network things just open /" in master? 
i start use calculate linux (gentoo) and this is NEW for me. 
I just install 0.13.0-r1
https://packages.gentoo.org/packages/x11-libs/libfm-qt
https://packages.gentoo.org/packages/x11-misc/pcmanfm-qt
and i have this problem.
I'm not good at programming and patching. Therefore, the hope is only for developers of the distribution kit.

tsujan:
> do you fix "click on the network things just open /" in master?

It's fixed in the latest git libfm-qt+pcmanfm-qt.

dracwyrm:
@tsujan is there a specific patch that fixes @johndoe71rus problem? These are the three commits I added to the ebuild in Gentoo: https://github.com/gentoo/gentoo/tree/master/x11-libs/libfm-qt/files I searched the commit history and didn't see anything related to his issue.

Or will there be a new release or patch release soon and I don't need to worry about this?

Thanks

tsujan:
@dracwyrm Please always try the latest git versions of libfm-qt and pcmanfm-qt if you want to have the fixes. I assure you that he git versions are practically stable and if they have a problem, it'll be fixed soon after you report it. Gentoo is a good distro for that.

tsujan:
BTW, commit titles should be short and so, they may not reflect what they do.

jubalh:
>  Please always try the latest git versions of libfm-qt and pcmanfm-qt if you want to have the fixes. I assure you that he git versions are practically stable and if they have a problem, it'll be fixed soon after you report it.

But we cannot ask distro to always use latest git in their packages. Usually a distro takes the latest stable version and applies certain patches for things that they need fixed. They don't want all the new functionality but have known bugs fixed.

I think @dracwyrm is asking which commits we think might fix the bug so that he can take those patches in. If a commit contains fixes and new features it is their task to seperate the two from each other and apply only the fixes. But pointing to relevant commits helps a lot.

> BTW, commit titles should be short and so, they may not reflect what they do.

That's right. They should just give a short description. The explanation or context should be after the title (if needed)
For example:

> Fix buffer overflow in MyFunc()
>
> We had an issue that happens when.... (continue explanation)

I like [this one](https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53)

tsujan:
> But we cannot ask distro...

Suggested to the person; don't have a good verbal memory.

> I think @dracwyrm is asking which commits...

It belonged to a period when several interrelated issues were being attacked; I couldn't be sure commit-wise.

> That's right. They should just give a short description...

Didn't mean that. We thought this commit fixed it. It was necessary but what finally fixed it was beyond a special case.

Actually, I couldn't work on SMB -- that was a big problem -- but @agaida's tests and clear descriptions did the job.

agaida:
Addition: I mentioned it some times - depending on the knowledge of the backporter one will introduce follow-up bugs with wild porting back upstream changes. So: If one port back something - please check for abi-breaks. We didn't bumped the soname for nothing in the upcoming release. 

Rule of thumb: If porting back something to libfm-qt/release check for new provided symbols or at least build lximage, lxqt-qtplugin and pcmanfm-qt against the changed lib.

PR #216, opened 2018-06-16
Fix overwrite prompt with default suffix

tsujan:
Fixes https://github.com/lxqt/lxqt/issues/1504


tsujan:
This works in simple cases, like that of https://github.com/lxqt/lximage-qt/pull/141, but not in all cases (the non-native dialog works in all cases). Please don't merge it until I find a complete fix!

tsujan:
OK, never mind! The problem I've found isn't related to this PR. I'm not even sure how to describe it but it's about changing the default suffix *after* the dialog is shown. That's possible with the non-native file dialog but not with LXQt (or KDE) file dialog.

This PR can be merged if approved.

agaida:
GTM

PR #217, opened 2018-06-20
Prevent skipping of path deletions

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/508

Previously, deletions might be skipped because, although update signals are emitted in the proper order (deletion after creation, for example), info jobs are run asynchronously and a simple job of passing deleted paths is faster than that of adding file infos. For example, when a temporary file is added and deleted quickly, the deletion job might finish before the addition job and so, be skipped. Here, skipped deletions are remembered and completed after next info jobs.


agaida:
GTM

PR #218, opened 2018-06-20
Cancel previous list job on reloading folder

tsujan:
Because the second list job may be started before the first one is finished (I've seen such cases and there's no guarantee they don't happen), making `finishLoading()` be emitted twice after `startLoading()`, which can have side effects. Moreover, redundant computations are prevented.


agaida:
GTM

PR #242, opened 2018-07-24
Support adding metadata for trusting executables

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/737

### DIFF
--- a/src/core/fileinfo.cpp
+++ b/src/core/fileinfo.cpp
@@ -9,7 +9,8 @@ const char defaultGFileInfoQueryAttribs[] = "standard::*,"
                                             "time::*,"
                                             "access::*,"
                                             "id::filesystem,"
-                                            "metadata::emblems";
+                                            "metadata::emblems,"
+                                            "metadata::trust";
 
 FileInfo::FileInfo() {
     // FIXME: initialize numeric data members
@@ -23,6 +24,7 @@ FileInfo::~FileInfo() {
 }
 
 void FileInfo::setFromGFileInfo(const GObjectPtr<GFileInfo>& inf, const FilePath& parentDirPath) {
+    inf_ = inf;
     dirPath_ = parentDirPath;
     const char* tmp, *uri;
     GIcon* gicon;
@@ -378,6 +380,37 @@ bool FileInfo::isExecutableType() const {
         return false;
     }
     return mimeType_->canBeExecutable();
+}
+
+bool FileInfo::isTrustable() const {
+    if(isExecutableType()) {
+        /* to avoid GIO assertion warning: */
+        if(g_file_info_get_attribute_type(inf_.get(), "metadata::trust") == G_FILE_ATTRIBUTE_TYPE_STRING) {
+            if(const auto data = g_file_info_get_attribute_string(inf_.get(), "metadata::trust")) {
+                return (strcmp(data, "true") == 0);
+            }
+        }
+    }
+    return false;
+}
+
+void FileInfo::setTrustable(bool trust) const {
+    if(!isExecutableType()) {
+        return; // "metadata::trust" is only for executables
+    }
+    GObjectPtr<GFileInfo> info {g_file_info_new()}; // used to set only this attribute
+    if(trust) {
+        g_file_info_set_attribute_string(info.get(), "metadata::trust", "true");
+        g_file_info_set_attribute_string(inf_.get(), "metadata::trust", "true");
+    }
+    else {
+        g_file_info_set_attribute(info.get(), "metadata::trust", G_FILE_ATTRIBUTE_TYPE_INVALID, nullptr);
+        g_file_info_set_attribute(inf_.get(), "metadata::trust", G_FILE_ATTRIBUTE_TYPE_INVALID, nullptr);
+    }
+    g_file_set_attributes_from_info(path().gfile().get(),
+                                    info.get(),
+                                    G_FILE_QUERY_INFO_NONE,
+                                    nullptr, nullptr);
 }
 
 
--- a/src/core/basicfilelauncher.cpp
+++ b/src/core/basicfilelauncher.cpp
@@ -193,7 +193,7 @@ bool BasicFileLauncher::launchDesktopEntry(const FileInfoPtr &fileInfo, const Fi
     const char* desktopEntryName = nullptr;
     FilePathList shortcutTargetPaths;
     if(fileInfo->isExecutableType()) {
-        auto act = quickExec_ ? ExecAction::DIRECT_EXEC : askExecFile(fileInfo);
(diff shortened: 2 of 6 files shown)


tsujan:
WARNING: pcmanfm-qt and, probably, lxqt-qtplugin should be recompiled after this!

src/core/fileinfo.cpp
```diff
@@ -380,6 +382,37 @@ bool FileInfo::isExecutableType() const {
...
+}
+
+void FileInfo::setTrustable(bool trust) const {
+    if(!isExecutableType()) {
+        return; // "metadata::trust" is only for executables
+    }
+    GObjectPtr<GFileInfo> info {g_file_info_new()}; // used to set only this attribute
+    if(trust) {
+        g_file_info_set_attribute_string(info.get(), "metadata::trust", "true");
+        g_file_info_set_attribute_string(inf_.get(), "metadata::trust", "true");
```
palinek: These doubled lines here and below are intentional?

tsujan:
An explanantion may be needed (not in the code):

(1) We can set the attribute for the `GFile` asynchronously, use a callback function, and create a file info job to update the file info (theoretically).

(2) We can simply set the attribute for the `GFile` and, at the same time, add it to the *existing* file info because nothing else is changed. If a folder is opened multiple times in pcmanfm-qt, all of them will reflect the change made to the file.

I chose (2) because I found it more frugal, especially when attributes of multiple files are set/unset together.

yan12125:
> WARNING: pcmanfm-qt and, probably, lxqt-qtplugin should be recompiled after this!

Maybe time to introduce the Q_D approach to avoid ABI changes.

tsujan:
@yan12125 Please open an issue for that at https://github.com/lxqt/lxqt/issues. I'm curious to know what you mean.

PCMan:
@yan12125 I personally don't like pImpl pattern (Q_D). This makes C++ language features less useful since you can only have one private data member. Besides, to allocate one object, now you need two memory allocation calls, one of which only allocates 8 bytes for storing a pointer. So I like to avoid this deliberately.
Avoiding ABI changes in C++ is not always possible since you also need to make sure you don't add new virtual methods or change their orders. It's better to just use C if ABI is an important issue.
Q_D() makes code very ugly. If pimpl is really needed, I'd suggest just write plain C++ and avoid the ugly Qt macros. So the code can be used outside Qt projects as well.

About the Qt plugin, I think the best solution is to make libfm-qt an optional `runtime` dependency which can be loaded dynamically via calling a C function exported by libfm-qt.
In this way we fix the ABI problems and also avoid hard compile-time dependency on libfm-qt for lxqt-qtplugin. I can create a PR for this later.

yan12125:
@PCMan Thanks a lot for the detailed and easy-to-understand explanation! I appreciate it, and after reading that I changed my mind.

agaida:
@yan12125 - i wouldn't care about ABI changes right now - i would absolutly care about a clean implementation. To be honest, right now ABI changes are not a problem for any distribution as only a few packages use libfm-qt. This might change over time - and all things that we don't implement clean right now will then fall onto our feets at the time libfm-qt will be used by more packages.

yan12125:
@agaida - Right. Indeed Q_D or other PIMPL patterns bring more complications than benefits.

agaida:
@tsujan - maybe i overlooked a bit - but:
* on my Desktop there is no question about the execution anymore (check or uncheck the settiing in pcmanfm-qt has no effect)
* the trust setting is not shown on desktop files
* in other places it work fine - the trust thing is shown and works. Beside of that checking unchecking the main setting has no effect in other places too.

tsujan:
As for telling the user about an app restart, I've solved it in FeatherPad by showing a red label every time the user changes such an option. Perhaps a similar thing is needed with pcmanfm-qt.

agaida:
that would be cool - i didn't notice the tooltip

tsujan:
> Would you like to add the red label in a separate PR?

Oh, yes! The red label -- or anything that tells about app restart -- is very different from the subject of this PR.

As for adding an emblem to trusted desktop files, we'll discuss it elsewhere -- I think it needs a discussion. Moreover, its code will be different from that of the current PR.

agaida:
So let's just merge the PRs and hope the users find the hint for the time being.

tsujan:
Actually, once the folder is closed and reopened, it'll show the new state of that main option but, of course, Desktop is always open and needs app restart.

Adding a warning-bar is easy. Will open an issue for it soon.

agaida:
i see no problem merging it right now - but we should add this bar before a release ;)

tsujan:
Piece of cake.

wxl:
I recognize this is a little bit late in coming, but @tsujan is there a reason why you went with `metadata::trust` as opposed to `metadata::trusted` as it seems others like GNOME use?

tsujan:
I didn't know how Gnome set it and, IMO, there is no reason to change it now. There is no standard for it in GLib and Gnome's way can't be seen as a standard.

tsimonq2:
As an Ubuntu flavor, if we modify core infrastructure code (in this case we had to modify casper so this change would apply to our Calamares desktop file on our live ISO), we are expected to uphold consistency with the rest of the flavors. We are the only flavor who sets it to `metadata::trust` instead of `metadata::trusted`, and so instead of special-casing the infrastructure, we were told we should patch Ubuntu's copy of libfm-qt.

We'd like to send this back upstream so we don't have to carry a permanent patch, but as an Ubuntu flavor, we have expectations. I understand and I agree with not treating GNOME as a standard, but in this case we're aiming for consistency.

tl;dr we're patching this as a distributor anyway

tsujan:
> we're patching this as a distributor anyway

It's responsibility will be yours.

Frankly, if such patches become numerous, I'll have no choice other than closing reports made by Ubuntu users.

tsimonq2:
> Its responsibility will be yours.

We don't want it to be, but the Ubuntu Release Team expects consistency.

Is there an objection to making this `metadata::trusted` upstream __besides__ "there's no set standard for it"?

palinek:
For me it seems reasonable to use the same as gnome uses. But I'm not the one in charge here :)

tsujan:
> Is there an objection to making this metadata::trusted upstream besides "there's no set standard for it"?

The ardent discussion about the "inconsistency" of something SO unimportant on the part of a distro that heavily patched gtk3 for a long time makes no sense to me.

The pertinent question is that, apart from advertising for Ubuntu, do you have a really important use case to justify such a discussion? What's the "disaster" you want to prevent?

If you have an answer, feel free to make a PR; other devs might accept and merge it.

agaida:
btw - Gnome dropped desktop icons at all - so esp. this discussion about handling the trust for desktop icons and consistency is pointless.

PR #246, opened 2018-07-30
Prevent a too big file dialog

tsujan:
Fixes https://github.com/lxqt/libfm-qt/issues/244

### DIFF
--- a/src/filedialoghelper.cpp
+++ b/src/filedialoghelper.cpp
@@ -8,6 +8,8 @@
 #include <QTimer>
 #include <QSettings>
 #include <QtGlobal>
+#include <QApplication>
+#include <QDesktopWidget>
 
 #include <memory>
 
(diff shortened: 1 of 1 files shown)


PCMan:
@tsujan maybe we can use the facilities provided by Qt?
http://doc.qt.io/qt-5/restoring-geometry.html

tsujan:
@PCMan Yes but, here, the whole point is subtracting `QSize(2, 2)` because, apparently, some WMs make the window full-screen if its size is that of the desktop.

PCMan:
@tsujan I thought Qt already handled that. So it seems that we need to handle this ourselves?
Apart from size, we might also need to keep window state like "fullscreen" or "maxmized".
They are defined in NETWM protocols and I guess Qt more or less support them?
Standard conforming WMs should handle the hints properly.

PCMan:
@tsujan The PR looks good to me. Just curious if Qt handles the window size issue better so we can let Qt do the heavy lifting instead. If it doesn't, then please merge this.

tsujan:
@PCMan Qt handles the win size/state well; the problem is the WM, which may enforce a state.

On second thought, it's good to add the standard shortcut `F11` for (un-)full-screening and do as you suggested. Do you agree? Can I close this and make another PR later?

agaida:
F11 sounds like a good idea for this special case

tsujan:
OK. I think @PCMan agrees too. Closing this.

PR #249, opened 2018-08-03
Fullscreen shortcut for file dialog

tsujan:
Fixes https://github.com/lxqt/libfm-qt/issues/244

NOTE: saveGeometry/restoreGeometry isn't added to the code because we don't want to restore the position and, especially, we don't want the dialog to be opened on another screen. So, as before, only the size is saved but `F11` is added as the fullscreen shortcut.

### DIFF
--- a/src/filedialog.cpp
+++ b/src/filedialog.cpp
@@ -91,9 +91,15 @@ FileDialog::FileDialog(QWidget* parent, FilePath path) :
     ui->fileTypeCombo->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Fixed);
     ui->fileTypeCombo->setCurrentIndex(0);
 
+    // shortcut for showing/hiding hidden files
     QShortcut* shortcut = new QShortcut(QKeySequence(Qt::CTRL + Qt::Key_H), this);
     connect(shortcut, &QShortcut::activated, [this]() {
         proxyModel_->setShowHidden(!proxyModel_->showHidden());
+    });
+    // fullscreen shortcut
+    shortcut = new QShortcut(QKeySequence(Qt::Key_F11), this);
+    connect (shortcut, &QShortcut::activated, [this]() {
+        setWindowState(windowState() ^ Qt::WindowFullScreen);
     });
 
     // setup toolbar buttons


tsimonq2:
LGTM.

agaida:
Look good, but does nothing with xfwm4 - not sure what it should do with openbox or kwin.

agaida:
no

tsujan:
So, xfwm4 is an exception. Nothing can be done for it. The code is standard: http://doc.qt.io/qt-5/qwidget.html#setWindowState

tsujan:
@agaida Since the full-screen problem appeared only with xfwm4 and because, apparently, xfwm4 doesn't follow Qt's full-screen code, I think there's no need to this patch. Why a file dialog should have a full-screen shortcut anyway? Feel free to close this PR if you agree.

agaida:
i agree that the patch make more or less no sense - but i don't agree that there should be no solution :)

tsujan:
> but i don't agree that there should be no solution

In programming, there's always a solution but, in this case, it may be a X11 code, which isn't an option for us.

tsujan:
And the lesson learned: not all X11 WMs behave the same way when it comes to window states.

agaida:
i guess there is a more simple solution - somewhere the open dialog save the dimensions - and it is possible that these dimensions are bigger than the current screen size - only a guess, that can be  a starting point

tsujan:
BTW, I think you should see the same xfwm4 issue with pcmanfm-qt and other Qt apps. Make pcmanfm-qt's window bigger than the screen, exit and run it again. If not, xfwm4 should be the enemy of Qt dialogs ;)

agaida:
/me is in search for something big to throw it in @tsujan's directon. Hmm, maybe the ISS would fit.

In nicer words - i don't have that behaviour in a not installed live system. So everything is ok first hand. But i have this damn behaviour on my main workstation. Have to check it on a different machine, it is something really obscure.

tsujan:
> i don't have that behaviour in a not installed live system.

If I were you, I'd first check xfwm4 settings -- both as root and as user.

agaida:
found the reason for the funny behaviour - it's called ~/.config/lxqt/filedialog.conf

After deleting this file all things are normal - so my questions are: Why was the file created, why is the file still in use if it exists? And the last question: can it created again :sunglasses:

agaida:
Did we save the positions somewhen in the past? And might it be that the file isn't created anymore, but if there still reed - and far more important: if the file and the keys in it exist - is there any write access too? I was a bit dumb, i deleted it - should have only moved it.

tsujan:
`~/.config/lxqt/filedialog.conf` isn't the cause; xfwm4 is. `filedialog.conf` is like other LXQt/Qt config files; there's nothing special about it.

In short, a WM shouldn't make windows fullscreen without their apps' consent and if it does so (a bad behavior), it should be able to make them un-fullscreen itself.

agaida:
oha - unbelievable - thats so absurd that it is funny again :D - In case you noted the unusual geometry in the edge of my screenshot - it is a xfwm4 setting.

tsujan:
Haha... I new there should be a setting :)

That setting isn't good for the above-mentioned reason. See how caring KWin is, in comparison:

kwin

agaida:
the funny thing is that the window don't behave like a full screen window - it is movable etc - but because of the missed decorations part not sizable anymore :D

tsujan:
Happy to know your problem is solved :)

agaida:
kind of :D

PR #254, opened 2018-08-13
Emblem for untrusted desktop files inside home

tsujan:
Closes https://github.com/lxqt/pcmanfm-qt/issues/751

WARNING: pcmanfm-qt should be recompiled immediately after this.

For practical reasons, `emblem-important` is added to the file icon only if it is:
 * A desktop file;
 * Untrusted; and
 * Deletable (by user).

Also, an old issue is found and fixed: previously, icons of list views could have only one emblem, while 2 emblems were possible -- and now, 3 emblems are possible (symlink, untrusted and an emblem set by the user).


agaida:
All things fine, GTM

PR #256, opened 2018-08-16
Selection/deselection marks on mouse-over

tsujan:
Closes https://github.com/lxqt/pcmanfm-qt/issues/132

Selection(/deselection) marks are drawn on mouse-over, near top left corners of icons. Their size is 1/3 of the icon size. They are only for icon views and icon sizes >= 48 px (because otherwise, the user couldn't click on them easily).

Each selection mark has its own mouse-over state, being semi-transparent when the cursor is located inside the icon but not over the mark, and becoming opaque when under the cursor.

For now, they are icons (`list-add` and `list-remove`) but it might be possible to draw them as texts with (semi-transparent) backgrounds.


tsujan:
Oh, I forgot to add that *pcmanfm-qt should be recompiled after this*.

agaida:
Works fine, GTM

PR #261, opened 2018-08-25
Ensure visibility of toggled path button on resizing

tsujan:
Fixes https://github.com/lxqt/libfm-qt/issues/260

The reason is that the toggled path button shows the current path to the user.

The code keeps track of the toggled button to find it quickly when needed.

NOTE: Recompile pcmanfm-qt after this!


agaida:
GTM

PR #267, opened 2018-09-03
Fixed file info job with parentless paths

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/768

A path may not have a parent, like `trash:///`. Such cases are now covered in file info jobs.

### DIFF
--- a/src/core/fileinfojob.cpp
+++ b/src/core/fileinfojob.cpp
@@ -29,7 +29,9 @@ void FileInfoJob::exec() {
             };
             if(inf) {
                 // Reuse the same dirPath object when the path remains the same (optimize for files in the same dir)
-                auto dirPath = commonDirPath_.isValid() ? commonDirPath_ : path.parent();
+                auto dirPath = commonDirPath_.isValid() ? commonDirPath_
+                                                        : path.hasParent() ? path.parent()
+                                                                           : path; // e.g., trash:///
                 auto fileInfoPtr = std::make_shared<FileInfo>(inf, dirPath);
 
                 // FIXME: this is not elegant


tsujan:
@agaida Apart from the issue this patch fixes, please also see whether it has effect on the SMB problem.

tsujan:
This patch solves a problem and creates another one. Please don't merge it until I fix everything!

tsujan:
Done! The patch is ready for tests :)

The problem was hidden when we used libfm in libfm-qt.

agaida:
@tsujan: Do you want the good or the bad news first?

tsujan:
Bad news are more interesting in the end ;)

agaida:
I will make it short - nearly all filesystems that come to mind work more or less fine - the problem is the more or less. Even if i would be a native speaker i wouldn't be able to describe the few remaining problems correct. At least i have found three of them. Will make a screencast - guess that will take 10 min

tsujan:
Great that your tests are so comprehensive! Looking forward to your screencast.

agaida:
ok - a few things first:
```
admin:// - works fine, but try to open a file
pcmanfm-qt sftp://agaida@dev.siduction.org/home/agaida/pyfll - really impressive
pcmanfm-qt trash:/// - all fine, but try to open a file
                     - noteworthy trash:///\home...
pcmanfm-qt network:/// - opens fine, one machine and the smb-net are displayed
pcmanfm-qt smb:/// - but no items are displayed
pcmanfm-qt smb://idefix - shares are displayed
pcmanfm-qt smb://idefix/$share - the share will be mounted and displayed
pcmanfm-qt davs://agaida@nextcloud.xxx.org/remote.php/webdav - cool, davfs will be mounted after login
```

agaida:
pcmanfm show only idefix and the workgroup as network:/// - this should be fine because pcmanfm-qt should know idefix somehow. When i double-click on idefix the name resolution is smb://idefix.local - might be mdnss or how they call the service today - smbtree output:
```
WORKGROUP
        \\IDEFIX                        Samba 4.7.1
                \\IDEFIX\IPC$                   IPC Service (Samba 4.7.1)
                \\IDEFIX\photos                 Samba-Export
                \\IDEFIX\music                  Samba-Export
                \\IDEFIX\videos                 Samba-Export
                \\IDEFIX\sorter                 Samba-Export
        \\FRITZ-NAS                     FRITZ!Box
                \\FRITZ-NAS\IPC$                IPC Service (FRITZ!Box)
                \\FRITZ-NAS\FRITZ.NAS      
        \\ANNE_LAPTOP    
                \\ANNE_LAPTOP\Users          
                \\ANNE_LAPTOP\IPC$              Remote-IPC
                \\ANNE_LAPTOP\C$                Standardfreigabe
                \\ANNE_LAPTOP\C              
                \\ANNE_LAPTOP\ADMIN$            Remoteverwaltung
```

tsujan:
Thanks! I can test and work on everything except for SMB; I can just hope that fixing other issues might have a positive effect on SMB. Will start to test your list and wrestle with the problems tomorrow.

agaida:
there is one problem with reading remote directories - unfortunately i'm not able to reproduce the bug right now. In short - a message shows up that the current processing is incomplete or canceled - but the operation succeed and the directories and files are displayed right. hmm, will try it harder.

tsujan:
`admin:///` doesn't work either when put in the path entry, in the same way you described. I have a 3-4 months old pcmanfm-qt on another machine and also an ancient one on an ancient Debian; should see whether `admin:///` worked there but there's definitely a problem to fix.

I should also see if these problems are interrelated or unrelated to each other. If fixed, they might need separate PRs.

tsujan:
Oh, I'm so forgetful: https://github.com/lxqt/pcmanfm-qt/issues/567 . The problem with `admin:///` is old and, probably, not related to the problem with Trash.

agaida:
ok - [](https://youtu.be/m8HVsTdmksQ)

tsujan:
The good thing is that everything is inside libfm-qt. So, theoretically speaking, all issues should be fixed -- admin too.

tsujan:
@agaida As for opening files in Trash, Nautilus and Dolphin have the same behavior. The trashed files aren't for opening and neither Nautilus nor Dolphin has an "Open" item for them. Dolphin opens trashed text files in FeatherPad when they're clicked but not audio or video files. I think we could just remove "Open" and "open with" for trashed files later.

agaida:
Make sense

tsujan:
@agaida I forgot to ask this: Here we checked command-line but is it OK when you click on your `IDEFIX` item? In one of your previous screencasts, clicking on `IDEFIX` opened `/`.

agaida:
:) - it's a bit different
smbtree know //IDEFIX - so thats happend when i open it from command line:


if i open idefix with a double click directly in pcmanfm-qt it look like this

agaida:
So i guess i should investigate which service deliver the idefix.local name resolution :)

tsujan:
Strange, but not more than before. I'll keep it in mind but something like it should happen outside SMB for me to be able to investigate it.

tsujan:
@agaida After more digging into the code, I think this isn't a fix -- not even for `trash:///` -- but an ugly workaround for a deeper problem in `FileInfo`. Will try to deal with the real problem.

PR #269, opened 2018-09-05
Revert the fix for parentless paths

tsujan:
Reverts https://github.com/lxqt/libfm-qt/commit/b11935e510804e68b393ad7bc4eb5ca12297ccf6 and https://github.com/lxqt/libfm-qt/commit/26094801dad1aedcb95a90d2b1f77c39d563adb4

That method wasn't a good idea and had side effects. A real solution should be found.


tsujan:
Sorry for this; I should have been more cautious.

The previously "fixed" issue is returned, of course, but the "solution" had side effects (like restoring the trash contents to a mounted partition) and there were similar issues that weren't fixed by it. A more fundamental solution is needed.

PR #285, opened 2018-09-30
Fix selection on file creation in detailed list view

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/788

The whole row should be selected, not just its first column.


tsujan:
Also made the code style coherent (4 spaces instead of two for indentation).

PR #286, opened 2018-10-03
Set dir list job cancellation error to mild

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/789

A directory list job may be legally cancelled at `Folder::reload()`, in which case, pcmanfm-qt shouldn't show an error message (pcmanfm-qt only shows errors with severities greater than `ErrorSeverity::MILD`).


agaida:
GTM

PR #298, opened 2018-10-14
replace mempcpy with basic memcpy

BabylonAS:
in this context GNU-specific mempcpy offers no advantages over standard memcpy


palinek:
GTM

PR #303, opened 2018-10-29
faster invertSelection()

damianatorrpm:


tsujan:
Very nice!

Please change the comment to this for it to be understandable and for your PR to be merged:
```
// we don't use a "for" loop on rows because it would be slow
```

damianatorrpm:
@tsujan updated comment

tsujan:
Thank you very much! Merged.

PR #308, opened 2018-11-08
Don't update trash icon too fast

tsujan:
Instead, use a timer to limit the updates to 4 times per second at most.


PCMan:
LGTM.

PR #315, opened 2018-11-21
Don't hide template acions when template folder is loaded

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/828


tsujan:
Sorry, I close this to make a PR for all template issues soon because they are interrelated.

PR #324, opened 2018-12-06
Fix invisible view icons with HDPI scaling

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/834

To test this, stop pcmanfm-qt and run it with `QT_SCALE_FACTOR=2 pcmanfm-qt`: previously, the view icons were invisible.

Blurry side-pane icons with PNG icon sets and HDPI scaling are not specific to pcmanfm-qt/libfm-qt but are caused by an issue in libqtxdg, which will be fixed later.


agaida:
not really forgotten - different story.

PR #333, opened 2018-12-26
fileprogsdialog: handle file number in translation

Fat-Zer:
This sort of things like declenting a word against a number should be handled in translations rather in source.
See: http://doc.qt.io/qt-5/i18n-source-translation.html#handling-plurals

### DIFF
--- a/src/filepropsdialog.cpp
+++ b/src/filepropsdialog.cpp
@@ -376,10 +376,8 @@ void FilePropsDialog::onFileSizeTimerTimeout() {
             unsigned int fc =  totalSizeJob->fileCount(); // the directory is included
             if (fc <= 1)
                 str = tr("no file");
-            else if (fc == 2)
-                str = tr("one file");
             else
-                str = tr("%1 files").arg(fc - 1);
+                str = tr("%n file(s)", "", fc - 1);
             ui->fileNumber->setText(str);
         }
     }


tsujan:
> ... should be handled in translations rather in source.

I disagree. The current source is correct and elegantly handles numbers. See it as a style!

tsujan:
These things aren't written in stone.

Fat-Zer:
@tsujan, current approach suitable only for English, it doesn't allow to properly translate to languages with different declination rules for plural numbers. You should at least use `tr("%n files", "", fc - 1)` for the last case.

And if s-ending would be handled by translations, I see no point in such inconsistency as writing ’one’ in text and everything else in digits (but if you insist on this I'm ok with it).

Fat-Zer:
> The plural is handled separately in the code. So, it's the exact opposite of what you said above.

There is now no way to handle that in translations for non-english properly now. For most of languages rules are more complicated than just «1 => nothing, 2-… => 's'».

e.g. for Russian I need 3 variants:
* 2-4 файл**а**
* 5-20 файл**ов**
* 21 файл_

That's why you should use such plural handling system exists in first the first place. And this is an exactly good moment to actually *follow* the rule.

>The coder may want to write "no file" or "nothing" instead of (the ugly string) "0 files"; he/she may also want to have "one file", "only one file", "just a file", etc. for "1 file". That's called a "coding style", not an "inconsistency".

Ok, let it be a separate string...

tsujan:
`tr("%n file(s)", "", ...)` creates strings like this *without translation*: "37 file(s)"; it looks ridiculous. Enforcing a translation for the default language (us English) seems like an ugly overkill to me.

Please tell me if you find an elegant solution, without strings like "37 file(s)".

Fat-Zer:
I believe the universal approach would be to have an english translation which handles only plural forms (as e.g. Gnome does). Ofc, I won't advocate to rush it now right away =). Actually I haven't that lxqt doesn't have such before the original patch...

The «ok for now» as I mentioned above would be to use `tr("%n files", "", fc - 1)` for the plural case. I've created pr#334 (I couldn't manage to change this one), it also uses %Ln, so it should have nice localized number representation (e.g. for English locale it would be "10,234 files" instead of 10234).

PR #334, opened 2018-12-26
fileprogsdialog: (i18n) handle file number plural forms

Fat-Zer:
Use correct plural forms for file size for sake of laguages with multiple number of those.

### DIFF
--- a/src/filepropsdialog.cpp
+++ b/src/filepropsdialog.cpp
@@ -379,7 +379,7 @@ void FilePropsDialog::onFileSizeTimerTimeout() {
             else if (fc == 2)
                 str = tr("one file");
             else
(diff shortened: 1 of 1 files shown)


tsujan:
This does not result in strings like "40 file(s)" without translation and respects the singular and empty strings.

So, if it works with translations, GTM.

tsujan:
BTW, using `%Ln` is very nice; IMO, it's better to replace all cases of `%1` with `%Ln` in such strings.

Fat-Zer:
`%Ln` is for `tr()`. But for [`QString::arg()`](http://doc.qt.io/qt-5/qstring.html#arg-4) you may use `%L1`, `%L2` etc with the same effect.

PR #343, opened 2019-01-20
Fixed mimetype filters in LXQt file dialog

tsujan:
Fixes https://github.com/lxqt/libfm-qt/issues/342

(1) The problem was just a typo, which is corrected.

(2) I also took the liberty of replacing `QRegExp` with `QRegularExpression` because the former is deprecated -- I know it's a bad practice to have two changes in one PR but I might never do it if not here.

NOTE: The only way of testing this is using `QFileDialog::setMimeTypeFilters()` in a code. It passed my tests though.

### DIFF
--- a/src/filedialog.cpp
+++ b/src/filedialog.cpp
@@ -204,7 +204,7 @@ QStringList FileDialog::parseNames() const {
            && (firstQuote == 0 || fileNames.at(firstQuote - 1) != QLatin1Char('\\'))
            && fileNames.at(lastQuote - 1) != QLatin1Char('\\')) {
            // split the names
-            QRegExp sep{"\"\\s+\""};  // separated with " "
+            QRegularExpression sep{"\"\\s+\""};  // separated with " "
             parsedNames = fileNames.mid(firstQuote + 1, lastQuote - firstQuote - 1).split(sep);
             parsedNames.replaceInStrings(QLatin1String("\\\""), QLatin1String("\""));
         }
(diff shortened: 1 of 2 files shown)


src/filedialog.cpp
```diff
@@ -965,10 +965,13 @@ void FileDialog::FileDialogFilter::update() {
         }
         nameFilter = nameFilter.mid(left, right - left);
     }
-    // parse the "*.ext1 *.ext2 *.ext3 ..." list into QRegExp objects
+    // parse the "*.ext1 *.ext2 *.ext3 ..." list into QRegularExpression objects
     auto globs = nameFilter.simplified().split(' ');
     for(const auto& glob: globs) {
-        patterns_.emplace_back(QRegExp(glob, Qt::CaseInsensitive, QRegExp::Wildcard));
+        // NOTE: Use QRegularExpression::anchoredPattern() later, when Qt 5.12 is required.
```
PCMan: For Qt 5.12, it's OK to use conditional compilation here?
`#if QT_VERSION >= ....`

PCMan: LGTM.

tsujan: > For Qt 5.12, it's OK to use conditional compilation here?

IMHO, there was no need to that because it does exactly the same thing. However, at some time, Qt 5.12 will be the minimum dependency.

src/filedialog.cpp
```diff
@@ -965,10 +965,13 @@ void FileDialog::FileDialogFilter::update() {
...
+        patterns_.emplace_back(QRegularExpression("\\A(?:"
+                                                    + QRegularExpression::wildcardToRegularExpression(glob)
```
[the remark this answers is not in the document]
tsujan: @agaida Sorry! Will fix it today.

NOTE: I thought only `QRegularExpression::anchoredPattern()` was from 5.12, so I didn't use it and left a comment instead.

Issue #225, opened 2018-06-27
(Harmless?) error message with file dialog


tsujan:
##### Steps to Reproduce (for bugs)

1. Open a Qt app that uses native file dialogs from a terminal emulator.
2. Open a file dialog from that app.
3. Enter a valid path into its path bar (after making it editable).
4. Press ENTER.

Then, this message will appear in the terminal: `QCoreApplication::postEvent: Unexpected null receiver`. The file dialog will work fine though.

##### Possible Solution

Should see which event is sent to a null receiver, of course.

##### Context

I don't know if this existed before or happens with Qt 5.11.

##### System Information

Qt 5.11.1 + Latest git LXQt

Fat-Zer:
It's not only about file dialogs, I can reproduce the same thing with pcmanfm-qt.

Backtrace:
```
#0  0x00007ffff3cc2da0 in write () from /lib64/libc.so.6
#1  0x00007ffff3c4eddd in _IO_file_write () from /lib64/libc.so.6
#2  0x00007ffff3c4df92 in new_do_write () from /lib64/libc.so.6
#3  0x00007ffff3c4f63e in _IO_file_xsputn () from /lib64/libc.so.6
#4  0x00007ffff3c22c08 in buffered_vfprintf () from /lib64/libc.so.6
#5  0x00007ffff3c1fd1e in vfprintf () from /lib64/libc.so.6
#6  0x00007ffff3ce4195 in __fprintf_chk () from /lib64/libc.so.6
#7  0x00007ffff47c532d in ?? () from /usr/lib64/libQt5Core.so.5
#8  0x00007ffff47c0377 in ?? () from /usr/lib64/libQt5Core.so.5
#9  0x00007ffff47c0488 in ?? () from /usr/lib64/libQt5Core.so.5
#10 0x00007ffff47ad07d in QMessageLogger::warning(char const*, ...) const () from /usr/lib64/libQt5Core.so.5
#11 0x00007ffff4965b63 in QCoreApplication::postEvent(QObject*, QEvent*, int) () from /usr/lib64/libQt5Core.so.5
#12 0x00007ffff7b72b94 in Fm::PathBar::closeEditor() () from /usr/lib64/libfm-qt.so.6
#13 0x00007ffff498c46e in QMetaObject::activate(QObject*, int, int, void**) () from /usr/lib64/libQt5Core.so.5
#14 0x00007ffff57436a5 in ?? () from /usr/lib64/libQt5Widgets.so.5
[...]

PR #351, opened 2019-01-26
pathbar: fix an issue with a recursive call slot

Fat-Zer:
Closes issue #225

### DIFF
--- a/src/pathbar.cpp
+++ b/src/pathbar.cpp
@@ -314,6 +314,9 @@ void PathBar::closeEditor() {
     }
     // If a menu has popped up synchronously (with QMenu::exec), the path buttons may be drawn
     // but the path-edit may not disappear until the menu is closed. So, we hide it here.
+    // However, since hiding the path-edit makes it lose focus and emit editingFinished(),
+    // we should first disconnect from it to avoid recursive calling of the current function.
+    tempPathEdit_->disconnect();
     tempPathEdit_->setVisible(false);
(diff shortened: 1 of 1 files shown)


src/pathbar.cpp
```diff
@@ -312,6 +312,9 @@ void PathBar::closeEditor() {
     if(tempPathEdit_ == nullptr) {
         return;
     }
+    // Avoid recursive calling of this slot from nested signals
+    // see https://github.com/lxqt/libfm-qt/issues/225
+    tempPathEdit_->disconnect();
     // If a menu has popped up synchronously (with QMenu::exec), the path buttons may be drawn
     // but the path-edit may not disappear until the menu is closed. So, we hide it here.
```
tsujan: Nice catch! Would you please move `tempPathEdit_->disconnect();` here, below the hiding comment but above the hiding line and add this to the hiding comment instead:
```c++
    // But since hiding the path-edit makes it lose focus and emit editingFinished(),
    // we should disconnect from it to avoid recursive calling of the current function.
```

tsujan:
Or a little better:
```c++
    // However, since hiding the path-edit makes it lose focus and emit editingFinished(),
    // we should first disconnect from it to avoid recursive calling of the current function.
```

src/pathbar.cpp
Fat-Zer: done

tsujan:
GTM.

PR #352, opened 2019-01-27
Added missing sort-by-group action

tsujan:
This is related to https://github.com/lxqt/pcmanfm-qt/issues/886 and will be followed by a pcmanfm-qt patch, which should be applied immediately after it (otherwise, a crash will happen).

It adds the missing context menu-item "By File Group".


tsujan:
The above-mentioned pcmanfm-qt PR is https://github.com/lxqt/pcmanfm-qt/pull/888

PR #355, opened 2019-01-29
Added custom widths and visibility for detailed columns

tsujan:
The options are in the header right-click menu. They already work for the current view but, of course, saving/restoring should not be done by libfm-qt. Hiding the file name column is not allowed.

A small pcmanfm-qt patch will follow this one soon to save and restore custom settings. LXQt file dialog could also be easily patched later.

NOTE 1: pcmanfm-qt should be recompiled after this, preferably with the above-mentioned patch.

NOTE 2: The previous patches about fixing file group should be applied before this.

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -42,6 +42,8 @@
 #include <QMessageBox>
 #include <QLineEdit>
 #include <QTextEdit>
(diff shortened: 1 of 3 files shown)


tsujan:
The pacmanfm-qt PR is https://github.com/lxqt/pcmanfm-qt/pull/889

PR #358, opened 2019-02-01
Add KDevelop project files to gitignore

apt-ghetto:
The project files of KDevelop are not part of the source, so ignore
them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ build*/
 Session.vim
 .netrwhist
 *~
+.kdev4
+*.kdev4


agaida:
afaik  it was only .kdev4 - but it will not hurt - we don't export them, so it seems to be forgotten, (.gitattributes)

tsujan:
How can something like KDevelop have any relation to LXQt?

apt-ghetto:
The KDevelop files are not related to LXQT, but if you import the project (repo) with KDevelop, it creates the project files. And to avoid, that they are added to a commit, I thought to create a pull request.

tsujan:
OK. KDevelop was a big surprise; it's worse than Qt Creator, which itself is an obstacle to development.

agaida:
tststs - try idea :D

agaida:
Nice question to start a war (see also emacs vs vim)
* the Qt guys like Qt Creator
* the KDE guys like KDevelop
* People who what to get work done take a MS IDE
* Some people use Eclipse or a derivative - only to complain about needed ressources and speed
* Jetbrains CLion might be a good choice

tsujan:
This question is asked before and LXQt devs have told about their preferences. Short answer: a good terminal + a good text editor.

However, you might be more comfortable with KDevelop; there's no problem with that. You might also want to try Qt Creator or other IDEs.

I just told about my personal choice above, knowing that some other LXQt devs would have similar choices. I didn't mean it to enforce anything.

PR #398, opened 2019-04-16
Replace char raw strings literals

luis-pereira:
Avoid memory allocations.
Done by CLAZY_FIXIT="fix-fromCharPtrAllocations".

### DIFF
--- a/src/colorbutton.cpp
+++ b/src/colorbutton.cpp
@@ -43,7 +43,7 @@ void ColorButton::setColor(const QColor& color) {
         color_ = color;
         // use qss instead of QPalette to set the background color
         // otherwise, this won't work when using the gtk style.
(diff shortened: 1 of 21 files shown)


tsujan:
It seems good. I'll read it line by line soon. Thanks!

luis-pereira:
I've checked, but another pair of eyes is always a good thing. And you now that source much better than I do.

tsujan:
> I've checked, but another pair of eyes is always a good thing.

Your PRs have always been most accurate. I guess I've become too cautious when it comes to libfm-qt.

luis-pereira:
> Your PRs have always been most accurate. I guess I've become too cautious when it comes to libfm-qt.

That's the power of open source.

PR #399, opened 2019-04-17
Port towards C++11 nullptr, part 2

luis-pereira:
These ones somehow escaped clang-tidy... and me.


luis-pereira:
The ones that got away!

tsujan:
GTM

PR #401, opened 2019-04-18
Fix editbookmarksdialog

damianatorrpm:
The path variable in new bookmarkitem was previously unitialized (path can be empty when clicking "OK") which creates a non-visible bookmark (for libfm-qt)  which crashes other file managers (which will show a bookmark and a null path).


damianatorrpm:
there is still an additional issue with editing a bookmarkitem, will push again after that is fixed as well

PR #402, opened 2019-04-18
Update editbookmarksdialog.cpp

damianatorrpm:
Fixes creating bookmark items with NULL path, which:
exists, but are invisible for fm-qt
crashes other filemanagers (tested with nautilus) which see bookmarks with NULL path

### DIFF
--- a/src/editbookmarksdialog.cpp
+++ b/src/editbookmarksdialog.cpp
@@ -77,7 +77,7 @@ void EditBookmarksDialog::accept() {
             QString name = item->data(0, Qt::DisplayRole).toString();
             QUrl url = QUrl::fromUserInput(item->data(1, Qt::DisplayRole).toString());
             if (!url.isValid())
-                url = QUrl("file:///");
+                url = QUrl::fromUserInput(QString::fromUtf8(Fm::FilePath::homeDir().toString().get()));
             file.write(url.toEncoded());
             file.write(" ");
             file.write(name.toUtf8());
@@ -92,7 +92,6 @@ void EditBookmarksDialog::accept() {
 void EditBookmarksDialog::onAddItem() {
     QTreeWidgetItem* item = new QTreeWidgetItem();
     item->setData(0, Qt::DisplayRole, tr("New bookmark"));
-    item->setData(1, Qt::DisplayRole, "/");
     item->setFlags(Qt::ItemIsSelectable | Qt::ItemIsEditable | Qt::ItemIsDragEnabled | Qt::ItemIsEnabled);
     ui->treeWidget->addTopLevelItem(item);
     ui->treeWidget->editItem(item);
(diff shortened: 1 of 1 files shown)


src/editbookmarksdialog.cpp
```diff
@@ -90,6 +92,7 @@ void EditBookmarksDialog::accept() {
 void EditBookmarksDialog::onAddItem() {
     QTreeWidgetItem* item = new QTreeWidgetItem();
     item->setData(0, Qt::DisplayRole, tr("New bookmark"));
+    item->setData(1, Qt::DisplayRole, "/");
```
tsujan: This is redundant.

src/editbookmarksdialog.cpp
```diff
@@ -76,6 +76,8 @@ void EditBookmarksDialog::accept() {
             }
             QString name = item->data(0, Qt::DisplayRole).toString();
             QUrl url = QUrl::fromUserInput(item->data(1, Qt::DisplayRole).toString());
+            if (!url.isValid())
+                url = QUrl("file:///");
```
tsujan: IMO, user's home is a better option:
```c++
            if (!url.isValid()) {
                url = QUrl::fromUserInput(QString::fromUtf8(Fm::FilePath::homeDir().toString().get()));
            }
```

damianatorrpm: agreed

src/editbookmarksdialog.cpp
damianatorrpm: This should provide the inital value of the new bookmark item, on clicking "Add" without that it would n't it be empty until 
```
            if (!url.isValid()) {
                url = QUrl::fromUserInput(QString::fromUtf8(Fm::FilePath::homeDir().toString().get()));
            }
```
or set by the user?

src/editbookmarksdialog.cpp
tsujan: Please also keep the code style. I didn't above (it should have been `if(...)`).

src/editbookmarksdialog.cpp
tsujan: An initial value isn't needed because, with you code, an empty bookmark isn't possible on clicking OK.

BTW, this PR a good idea.

src/editbookmarksdialog.cpp
damianatorrpm: I have updated the PR

tsujan:
OK, GTM.

src/editbookmarksdialog.cpp
damianatorrpm: I have updated the PR. Thanks! I have some more issues that I have found previously and am eager to fix ( I haven't worked a while on the stuff linking to fm-qt but on other parts).
I will push theses changes also in due time.

tsujan: > I have some more issues that I have found previously and am eager to fix

Looking forward to your fixes.

Please also note that you usually don't need to close a PR and make another one if something's wrong; you could fix the first one.

PR #403, opened 2019-04-18
Clarify the license usage in libfm-qt

agaida:
Added the missed root licenses to the project

Refers to lxqt/libfm-qt/issues/370


damianatorrpm:
I guess the best time for such question is now:
What licence to the tests have?

agaida:
Thats easy - if not otherwise mentioned these files are under the main project license - LGPL. But you are right, best practice is that these files should have a explicit license header. Please open a bug about.

PR #405, opened 2019-04-18
Fix the license header of code we took from libfm

PCMan:
圖片

According to LStranger who is the original author of the vfs modules, the non-GTK+ part of libfm should be in LGPL. So I think the file headers for these two files should be fixed.

For the vfs-search.c, @tsujan fixed one of the bug so if @tsujan also has no objections it can be LGPL for sure.

### DIFF
--- a/src/core/vfs/vfs-menu.c
+++ b/src/core/vfs/vfs-menu.c
@@ -4,20 +4,19 @@
  *
  *      Copyright 2012-2014 Andriy Grytsenko (LStranger) <andrej@rep.kiev.ua>
  *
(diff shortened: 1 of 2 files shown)


tsujan:
> ... if @tsujan also has no objections

None at all.

agaida:
@PCMan - we (you) should add the date from the mail to the commit, just in case to mention the effective date of the license change - we will not be able to change the files in the wild, but at least we should mention that date in our source/commit

PCMan:
Updated the commit log to include the exact date of libfm license change.

PR #406, opened 2019-04-18
Warnings

luis-pereira:


luis-pereira:
It compiles with gcc and clang cleanly.

tsujan:
Very nice! Will test it soon.

tsujan:
This kind of warning isn't silenced:
```
warning: cast between incompatible function types from ‘void (*)(gpointer)’ {aka ‘void (*)(void*)’} to ‘GFunc’ {aka ‘void (*)(void*, void*)’} [-Wcast-function-type]
```

For example, in `basicfilelauncher.cpp`, we have:
```c++
g_list_foreach(uris, reinterpret_cast<GFunc>(g_free), nullptr);
```
`g_free` has one argument, while `GFunc` has two.

I think the warning can be suppressed by putting `g_free`(and other GLib functions that create similar warnings) inside a new function with the unused second  argument `gpointer user_data` but it doesn't seem to be worth the effort. Such warnings can be easily found in other GLib-based codes and if they get nasty, the aforementioned solution exists.

So, GTM.

luis-pereira:
@tsujan I didn't get those warnings. Probably different compiler versions. I used gcc 7.3.0 and clang 6.0... "old" stuff.

tsujan:
It's gcc 8.2.1 here but they're well-known and, I think, they can be ignored for now -- unless I find a concise way of silencing them (other than making a wrapper function).

Merging this...

luis-pereira:
With gcc I get this:
https://pastebin.com/raw/6vBU8Ldu
Odd

tsujan:
Should be related to the version. I even don't compile it with debugging symbols.

Googling for "cast between incompatible function types GFunc" shows a lot of results.

tsujan:
An example: https://gitlab.freedesktop.org/poppler/poppler/-/jobs/97144 (the compilation is successful).

PR #407, opened 2019-04-18
 added the licence headers

damianatorrpm:
and as year the one of the original commit


damianatorrpm:
@PCMan 
I hope the intention was to put the test cases to LGPL, if not I will adjust the PR.

PCMan:
Thanks! I forgot to add them.

PR #408, opened 2019-04-20
Don't ignore creation-deletion sequences

tsujan:
Fixes lxqt/pcmanfm-qt#944

Previously, if a file was in addition queue and then it came into the deletion queue, its addition and deletion were both ignored. That was wrong and could result in showing nonexistent files because addition can also happen in directory list job before being processed by file info job.

Also process accumulated changes only after finishing the current info job and don't clear all deletion paths after processing them (because, logically, only those paths that can be deleted should be removed).

### DIFF
--- a/src/core/folder.cpp
+++ b/src/core/folder.cpp
@@ -228,16 +228,6 @@ void Folder::onFileInfoFinished() {
         return;
     }
 
-    // process the changes accumulated during this info job
-    if(filesystem_info_pending // means a pending change; see "onFileSystemInfoFinished()"
-       || !paths_to_update.empty() || !paths_to_add.empty() || !paths_to_del.empty()) {
-        QTimer::singleShot(0, this, &Folder::processPendingChanges);
-    }
-    // there's no pending change at the moment; let the next one be processed
-    else {
-        has_idle_update_handler = false;
-    }
-
     FileInfoList files_to_add;
     FileInfoList files_to_delete;
     std::vector<FileInfoPair> files_to_update;
@@ -271,6 +261,16 @@ void Folder::onFileInfoFinished() {
         Q_EMIT filesChanged(files_to_update);
     }
     Q_EMIT contentChanged();
+
+    // process the changes accumulated during this info job
+    if(filesystem_info_pending // means a pending change; see "onFileSystemInfoFinished()"
+       || !paths_to_update.empty() || !paths_to_add.empty() || !paths_to_del.empty()) {
+        QTimer::singleShot(0, this, &Folder::processPendingChanges);
+    }
+    // there's no pending change at the moment; let the next one be processed
+    else {
+        has_idle_update_handler = false;
+    }
 }
 
 void Folder::processPendingChanges() {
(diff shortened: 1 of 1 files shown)


agaida:
Better, but still not ok ...

And i know that i'm a pain in the ass - why are these files created first hand - is there a reason? I could understand that they will be created if not there, but not every time when i click on apply or start the system.

tsujan:
> why are these files created first hand - is there a reason?

They aren't the problem; they show there is a problem. What you see with is just one example of what can happen (on VirtualBox, for example).

Until now, 2 flaws are fixed in the logic (I'm not sure how the second one can be practical). We need either patience or luck. And I'll need you to do still more tests if something comes to my mind. So, it'll be me who's a pain in the ass ;)

agaida:
if i'm right the desktop files will be created in desktopwindow.cpp like this:
```
void DesktopWindow::createHomeShortcut() {
    GKeyFile* kf = g_key_file_new();
    g_key_file_set_string(kf, "Desktop Entry", "Type", "Application");
    g_key_file_set_string(kf, "Desktop Entry", "Exec", Fm::CStrPtr(g_strconcat("pcmanfm-qt ", Fm::FilePath::homeDir().toString().get(), nullptr)).get());
    g_key_file_set_string(kf, "Desktop Entry", "Icon", "user-home");
    g_key_file_set_string(kf, "Desktop Entry", "Name", g_get_user_name());

    auto path = Fm::FilePath::fromLocalPath(XdgDir::readDesktopDir().toStdString().c_str()).localPath();
    auto trash_can = Fm::CStrPtr{g_build_filename(path.get(), "user-home.desktop", nullptr)};
    g_key_file_save_to_file(kf, trash_can.get(), nullptr);
    g_key_file_free(kf);
}
```
ok, fine - what i don't get is:
```
    // Home
    if(ds.contains(QLatin1String("Home"))) {
        createHomeShortcut();
    }
```
shouldn't it be
```
    // Home
if(ds.contains(QLatin1String("Home"))) {
    if not_aready_there(foo) {
        createHomeShortcut();
    }
}
```

tsujan:
> shouldn't it be...

No because someone may have created a file with the same name but different contents. Comparing contents isn't CPU friendly. That's intentional.

I say it again: it isn't the problem but a symptom.

agaida:
right - second question - why is the temp file created in the target dir instead of /tmp? Wouldn't solve the maybe underlying problem, but it would not be visible (and faster if /tmp is a tempfs)

tsujan:
> why is the temp file created in the target dir instead of /tmp? 

I discovered that with Qt years ago and found that GLib does the same. I also think /tmp is the best place because it's on RAM in most distros if not all.

Anyways, this is how I see this situation: an opportunity to find the cause of a problem. If no more ideas come to my mind, I'll have to install Manjaro on VitualBox, use qDebug() and see it first hand. It isn't something I can ignore.

agaida:
and i agree that finding the root cause is important - more import as the problem with the wrongly displayed files.

tsujan:
Now, it definitely works fine. Changed the title and comment.

agaida:
https://youtu.be/NQCBc_gFLkY - looks good, @wxl can you please test too

GTM

tsujan:
It definitely works everywhere. That report showed an important flaw in the logic, which is fixed here. The issue could happen anywhere but, because of timing and reloading, it was detectable on desktop with a weak CPU VirtualBox could provide.

PR #410, opened 2019-04-21
Workaround for GLib's recursive moving error, e.g. with bound mounts

tsujan:
`g_file_move()` may not work recursively on the same filesystem, especially with bound mounts (to `/mnt`, for example) and give the `G_IO_ERROR_WOULD_RECURSE` error. This patch ignores the error and tries our `FileTransferJob::copyFile()`.

Closes https://github.com/lxqt/pcmanfm-qt/issues/943

### DIFF
--- a/src/core/filetransferjob.cpp
+++ b/src/core/filetransferjob.cpp
@@ -83,6 +83,13 @@ bool FileTransferJob::moveFileSameFs(const FilePath& srcPath, const GFileInfoPtr
         // do the file operation
         if(!g_file_move(srcPath.gfile().get(), destPath.gfile().get(), GFileCopyFlags(flags), cancellable().get(),
                        nullptr, this, &err)) {
+            // Specially with mounts bound to /mnt, g_file_move() may give the recursive error
+            // and fail, in which case, we ignore the error and try copying and deleting.
+            if(err.code() == G_IO_ERROR_WOULD_RECURSE) {
+              if(auto parent = destPath.parent()) {
+                  return copyFile(srcPath, srcInfo, parent, destPath.baseName().get());
+              }
+            }
             retry = handleError(err, srcPath, srcInfo, destPath, flags);
         }
         else {


agaida:
You are a genius - will test it with some more cases before merge - anyways GTM

tsujan:
I'm not. PCMan's code is neatly organized.

agaida:
only nitpicking - it isn't limited to /mnt

just try 
```
cd ~
mkdir -p foo/bar foo/baz
mount -B ~/Downloads foo/bar
mount -B ~/Documents foo/baz
```
and have a bit fun.

tsujan:
> only nitpicking - it isn't limited to /mnt

I'd suspected that but the code is independent from /mnt.

agaida:
https://i.imgur.com/R5SzECR.png ;)

tsujan:
>  would you be so kind and change the comments and PR description that way?

Inside the code, I've said,  "Specially with mounts bound to /mnt....", which means it isn't limited to /mnt.

As for the PR comment, OK, I'll add "for example".

Is everything else OK? I mean, could I merge it after changing the PR comment?

tsujan:
Please also note that this isn't even limited to bound mounts; it's for any occurrence of G_IO_ERROR_WOULD_RECURSE, with or without bound mounts.

agaida:
i think so

tsujan:
> i think so

I'm sure ;)

Will do it tonight, after finishing a job about lximage-qt.

tsujan:
Changed the commit explanation as well as the title and explanation of this page (PR).

tsujan:
BTW, gnome users don't encounter this issue with Nautilus because Nautilus doesn't use `g_file_move()` but something like our `FileTransferJob::copyFile()` *everywhere*. Maybe, the code was written before `g_file_move()` was introduced; I don't know.

Anyhow, our code shouldn't fail anymore because `FileTransferJob::copyFile()` is the fallback now.

agaida:
unblocked for debian buster

tsujan:
> unblocked for debian buster

Any plan for a point release? I don't think it'll be needed if the next LXQt version is going to be released 4-5 months later.

PR #413, opened 2019-04-26
Reminder to change lxqt-qtplugin when bumping the soversion

agaida:
Can be removed when lxqt-build-tools will do this for us


agaida:
not needed anymore - we have now a clean solution

PR #416, opened 2019-04-28
cleanup trash handling

damianatorrpm:
refactors trash handling, simplifies a lot

### DIFF
--- a/src/placesmodel.cpp
+++ b/src/placesmodel.cpp
@@ -53,6 +53,40 @@ PlacesModel::PlacesModel(QObject* parent):
     placesRoot->appendRow(desktopItem);
 
     trashItem_ = new PlacesModelItem("user-trash", tr("Trash"), Fm::FilePath::fromUri("trash:///"));
+    QTimer* trashUpdateTimer_ = new QTimer(this);
+    trashUpdateTimer_->setSingleShot(true);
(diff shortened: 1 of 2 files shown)


tsujan:
Sorry but no!

The current code is there for NOT opening the trash directory. To monitor the trash, we're only interested in `G_FILE_ATTRIBUTE_TRASH_ITEM_COUNT`, not the whole attributes defined at the top of `fileinfo.cpp` and, definitely, not all the computations done in `folder.cpp`.

I'm afraid a simpler code doesn't mean a better code necessarily.

damianatorrpm:
> Sorry but no!
> 
> The current code is there for NOT opening the trash directory. To monitor the trash, we're only interested in `G_FILE_ATTRIBUTE_TRASH_ITEM_COUNT`, not the whole attributes defined at the top of `fileinfo.cpp` and, definitely, not all the computations done in `folder.cpp`.
> 
> I'm afraid a simpler code doesn't mean a better code necessarily.
I see, the benefits for each approach would be:
1) Not opening:
    not having the trash folder loaded, lower memory consumption
1) opening:
    faster trash loading as it is already preloaded (and folder is a std::shared_ptr<Folder>)

tsujan:
> 2. opening:
> faster trash loading as it is already preloaded (and folder is a std::shared_ptr)

Lots of more computations when the trash isn't opened (= most of the time) and no benefit when it is opened.

damianatorrpm:
Indeed the memory usage is noticeable.

PR #418, opened 2019-04-30
rework trash

damianatorrpm:
another shot at reworking (trash-part) of placesmodel
please have a look at the // comments in 3rd commit


tsujan:
Reviewing these kinds of PR and, especially, checking for regressions, is time-consuming and needs a good reason. Why do you insist on reworking something? What's the problem you want to fix?

damianatorrpm:
> Reviewing these kinds of PR and, especially, checking for regressions, is time-consuming and needs a good reason. Why do you insist on reworking something? What's the problem you want to fix?

Insist, no.

This does reduce the code from 70 lines to 40. 
The goal is to share the simplified code with libfm-qt

The logic in it is the same and it makes use of newer c++ functions (static lamda in this case).
I recompiled it and the test are running fine.

I have similar code in my own class which is a replacement of PlacesModel for my own purposes 
(Not devicesroot,placesroot, bookmarksroot, but a single rootitem and some other stuff)

tsujan:
> The logic in it is...

No one has doubt that you know the logic of what you write.

> This does reduce the code from 70 lines to 40

Reducing the number of code lines isn't always desirable, as you already know from a previous PR.

With every new code structure, not only we should make sure that it doesn't have problem but also we should be very careful about possible regressions and new bugs it introduces. As I said, there should be a good reason to do the tedious job. For example, PCMan's porting libfm code into C++11 not only had a good reason but also was necessary but it introduced a lot of regressions we were fixing in 208 and even in 2019.

Mere reduction of code lines isn't a good reason, IMO. You might need to wait a long time before this is reviewed by someone.

damianatorrpm:
> > The logic in it is...
> 

Meaning that it does only G_FILE_ATTRIBUTE_TRASH_ITEM_COUNT
like you mentioned instead of storing the whole folder.
https://github.com/lxqt/libfm-qt/pull/416#issuecomment-487425305

> No one has doubt that you know the logic of what you write.
> 
> > This does reduce the code from 70 lines to 40
> 
> Reducing the number of code lines isn't always desirable, as you already know from a previous PR.
> 

In the previous PR I definitely made a mistake as you pointed out which would have decreased the performance.

> With every new code structure, not only we should make sure that it doesn't have problem but also we should be very careful about possible regressions and new bugs it introduces. As I said, there should be a good reason to do the tedious job. For example, PCMan's porting libfm code into C++11 not only had a good reason but also was necessary but it introduced a lot of regressions we were fixing in 208 and even in 2019.
> 
> Mere reduction of code lines isn't a good reason, IMO. You might need to wait a long time before this is reviewed by someone.

Understood, adjusting stuff from my own PlacesModel implementation to work with libfm-qt vision of a PlacesModel means additional work (good intended) for me anyways.

PCMan:
I added some comments in the changes.
1. Using helper function can often help readability and maintainability. C++ 11 lambdas are great, but it need to be used with care. You can make as many things use lambda as you want, but I'd suggest that only do this when it improve the readability. There is nothing wrong in writing separate helper functions.

2. We used glib/gio/gobject stuff internally. So you will see GObject which is purely C code and requires manual memory management. They are not taken care by QObject.

3. Some of the code was ported from the gtk+ version but some features are not fully implemented yet. So they seem to be unused at the moment but they are useful.

damianatorrpm:
> I added some comments in the changes.
> 
> 1. Using helper function can often help readability and maintainability. C++ 11 lambdas are great, but it need to be used with care. You can make as many things use lambda as you want, but I'd suggest that only do this when it improve the readability. There is nothing wrong in writing separate helper functions.

IMHO it increases the readability a lot, but this of course depends very much on the viewers perspective.

> 2. We used glib/gio/gobject stuff internally. So you will see GObject which is purely C code and requires manual memory management. They are not taken care by QObject.

I understand, it is strange though, manually freeing objects in the destructor, especially from a class that should live as long as the runtime of the application (probably they are added to be consistent).

> 3. Some of the code was ported from the gtk+ version but some features are not fully implemented yet. So they seem to be unused at the moment but they are useful.

Thank you for the explanation @PCMan @tsujan

tsujan:
> IMHO it increases the readability a lot

Yes and, personally, I like it. But it has 3 drawbacks: (1) it could result in a crash if its context is missing. (2) It can create a hassle when a disconnection is needed. (3) When its code lines are needed elsewhere, a slot is more handy.

Otherwise, lambda is nice.

PR #419, opened 2019-04-30
Row-by-row scrolling inside view when Shift is pressed

tsujan:
Some users of Qt apps know that Shift may change wheel scrolling speed. This patch makes the view scroll row-by-row with the mouse wheel when Shift is pressed and the cursor is inside the view.

Of course, in the compact view, Shift wheel scrolling is done column-by-column.

When the cursor is on the scrollbar, Qt's default behavior is left intact, i.e. the scroll is faster with Shift. Although I don't think users of a file manager need that, I didn't add extra code to change it.

NOTE: The arguments of `QWheelEvent` have changed recently. I used the old constructor to not break the compilation with older Qt versions but we should really bump the minimum version.

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -1528,6 +1528,20 @@ bool FolderView::eventFilter(QObject* watched, QEvent* event) {
                     return true;
                 }
             }
+            // row-by-row scrolling when Shift is pressed
+            if (QApplication::keyboardModifiers() & Qt::ShiftModifier)
+            {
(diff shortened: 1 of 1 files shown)


wxl:
Just to put this on record, I tested this out and it worked well. Definitely an improvement to have another option. The idea that it's modifier-based rather than preference-based is actually something I like.

tsujan:
@wxl Thanks for the test and feedback!

agaida:
GTM

PR #422, opened 2019-05-07
Don't use implicit string conversions

luis-pereira:
Make us aware of the implicit "casts".

### DIFF
--- a/src/filedialog.cpp
+++ b/src/filedialog.cpp
@@ -72,8 +72,8 @@ FileDialog::FileDialog(QWidget* parent, FilePath path) :
         ui->folderView->selectionModel()->clearSelection();
         QStringList parsedNames = parseNames();
         for(auto& name: parsedNames) {
-            if(!defaultSuffix_.isEmpty() && name.lastIndexOf('.') == -1) {
-                name += '.';
+            if(!defaultSuffix_.isEmpty() && name.lastIndexOf(QLatin1Char('.')) == -1) {
+                name += QLatin1Char('.');
                 name += defaultSuffix_;
             }
             selectFilePath(directoryPath_.child(name.toLocal8Bit().constData()));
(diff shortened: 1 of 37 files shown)


tsujan:
Yes, of course, it should be done, like in other parts of LXQt. One of us will have to rebase later because I want to add a PR; I hope it will be you ;)

src/foldermodel.cpp
```diff
@@ -224,7 +224,7 @@ QString FolderModel::makeTooltip(FolderModelItem* item) const {
         tip += QStringLiteral("<p><i>") + tr("Location:") + QStringLiteral("</i> ") + QString::fromUtf8(parent_str.get()) + QStringLiteral("</p>");
     }
     // file type
-    tip += QStringLiteral("<i>") + tr("File type:") + QStringLiteral("</i> ") + info->mimeType()->desc();
+    tip += QStringLiteral("<i>") + tr("File type:") + QStringLiteral("</i> ") + QLatin1String(info->mimeType()->desc());
```
tsujan: Wouldn't `QString::fromUtf8(info->mimeType()->desc())` be more consistent with other changes?

src/foldermodel.cpp
```diff
@@ -264,7 +264,7 @@ QVariant FolderModel::data(const QModelIndex& index, int role/* = Qt::DisplayRol
             return (showFullNames_ && !item->name().empty() ? QString::fromStdString(item->name())
                                                             : item->displayName());
         case ColumnFileType:
-            return QString(info->mimeType()->desc());
+            return QString::fromUtf8((info->mimeType()->desc()));
```
tsujan: Has an extra pair pf parentheses.

src/foldermodel.cpp
luis-pereira: Yes, it should. That one escaped.

src/foldermodel.cpp
luis-pereira: nice catch

luis-pereira:
It's already done in a big part of LXQt.

src/filedialog.cpp
```diff
@@ -969,17 +969,17 @@ void FileDialog::FileDialogFilter::update() {
...
             right = nameFilter.length();
         }
         nameFilter = nameFilter.mid(left, right - left);
     }
     // parse the "*.ext1 *.ext2 *.ext3 ..." list into QRegularExpression objects
-    auto globs = nameFilter.simplified().split(' ');
+    auto globs = nameFilter.simplified().split(QLatin1Char(' '));
     for(const auto& glob: globs) {
 #if (QT_VERSION >= QT_VERSION_CHECK(5,12,0))
         patterns_.emplace_back(QRegularExpression("\\A(?:"
```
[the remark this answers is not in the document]
luis-pereira: Using 5.11 here. will update.

PR #425, opened 2019-05-09
File dialog: Smarter extension handling on saving

tsujan:
Closes https://github.com/lxqt/libfm-qt/issues/421 and closes https://github.com/lxqt/libfm-qt/issues/423 and fixes https://github.com/lxqt/lxqt-archiver/issues/29

This is how extensions are handled in the save mode:

 * If the file name has no extension and there isn't a default suffix but there is a filter in "File type" combo — like `Image Files (*.png *.jpg)` — the first extension provided by the filter will be added to the saved file name.
 * If the file name has an extension and the contents of "File type" combo is changed by the user, the extension will be updated accordingly.
 * Filters like `NAME (DESCRIPTION) (*.X *.Y)` are covered.

Overwrite prompts are also handled appropriately.

NOTE: Recompile libfm-qt based apps after this.

### DIFF
--- a/src/filedialog.cpp
+++ b/src/filedialog.cpp
@@ -72,7 +72,7 @@ FileDialog::FileDialog(QWidget* parent, FilePath path) :
         ui->folderView->selectionModel()->clearSelection();
         QStringList parsedNames = parseNames();
         for(auto& name: parsedNames) {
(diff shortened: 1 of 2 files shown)


tsujan:
This still needs to be a little smarter to cover cases like https://github.com/lxqt/lxqt-archiver/issues/29. I'll add another commit, squash and tell when it's ready.

tsujan:
It's ready now.

PR #439, opened 2019-07-17
Fixed DND with MTP

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/301

We get the drag info as "text/uri-list", which works most of the time. However, that mimetype changes the URL list by parsing its elements (see Qt → qmimedata.cpp → QMimeData::setData), while MTP needs the original list to query file info and support copying through DND. This patch gets the drag info without parsing it by making up the mimetype "libfm/files".


agaida:
GTM

PR #440, opened 2019-07-17
Update rubberband selection with smooth scrolling

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/978

This patch fixes the following minor issue by sending a mouse move event to the viewport after each smooth scroll:

 (1) In a folder with many files, select a few at the top by pressing the left mouse button and dragging the cursor over them but, after that, do not release the left mouse button;
 (2) While the cursor is motionless and the left mouse button is still pressed, turn the mouse wheel so that the view is scrolled down.

Previously, the selection wasn't updated correctly.


agaida:
GTM

PR #442, opened 2019-07-19
Fix crash when window is closed before finishing mount operation

tsujan:
Fixes https://github.com/lxqt/libfm-qt/issues/441

The explanation is in the code comment. The crash can happen only with bad mounts (like MTP); usually, the user can't close the window before the mount operation is finished.


agaida:
GTM

PR #443, opened 2019-07-20
Support DND to Places panel

tsujan:
Closes https://github.com/lxqt/pcmanfm-qt/issues/987

Files can be dragged and dropped to Places panel.

Also, if a folder is dropped on the Bookmarks item, it will be added as a bookmark.

NOTE 1: I excluded Computer, Applications and Network items because I didn't find a meaning for DNDing to them.

NOTE 2: For the sake of consistency, Trash is treated like other items but only with "Move here" and "Cancel" operations on the menu, regardless of prompt setting. Only if users threaten to kill me because of it and I can't kill them first, I might think about changing it.

### DIFF
--- a/src/placesmodel.cpp
+++ b/src/placesmodel.cpp
@@ -27,6 +27,8 @@
 #include <QStandardPaths>
 #include "utilities.h"
 #include "placesmodelitem.h"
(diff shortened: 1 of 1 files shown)


agaida:
GTM

PR #444, opened 2019-07-25
Easier and richer auto-completion

tsujan:
Closes https://github.com/lxqt/pcmanfm-qt/issues/201

Auto-completion always results in a path that ends with a slash. If an auto-completion has only one result, `Tab` will select it and open the next auto-completion popup (if any). Otherwise, Tab will select the next match on the popup.

When the popup is visible, pressing `Escape` not only closes it but also restores the last typed text.

`Backtab` is treated like `Tab` but for selecting the previous match.

To improve performance, matches are ordered alphabetically (see Qt doc).

As before, directory change requires pressing `Enter/Return`; `Tab` only puts the text into the path-bar.

NOTE: I tried to make it as logical as possible and it became very similar to what Dolphin did — but not identical because, in pcmanfm-qt, applying the path gives the focus to the main view. I think users could easily grasp it.

### DIFF
--- a/src/pathedit.cpp
+++ b/src/pathedit.cpp
@@ -21,12 +21,14 @@
 #include "pathedit.h"
 #include "pathedit_p.h"
 #include <QCompleter>
+#include <QAbstractItemView>
 #include <QStringListModel>
 #include <QStringBuilder>
 #include <QThread>
 #include <QDebug>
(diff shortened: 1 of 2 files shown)


tsujan:
@agaida You merged it but are you satisfied with it? If not, please tell why by opening a separate issue!

PR #446, opened 2019-07-27
Cleanup

shlyakpavel:
There is no sense in testing the 'delegate' pointer against null, as the memory was allocated using the 'new' operator. The exception will be generated in the case of memory allocation error.


tsujan:
GTM

PR #449, opened 2019-07-31
Fixes /* FIXME: we shouldn't trash a file already in trash:/// */

damianatorrpm:
Hopefully Fixes /* FIXME: we shouldn't trash a file already in trash:/// */


tsujan:
Thanks but if the check was needed, it shouldn't be done in a separate loop. Anyhow, it isn't needed, so we don't waste resources.

The reason why it isn't needed is that GLib takes care of it. If an implementation tries to move a trahsed file into trash (we don't), it goes into `unsupportedFiles_` and then, an appropriate prompt will be shown for deleting it in `FileOperation::onJobFinish()`.

PR #455, opened 2019-08-10
Don't show "untrust" emblem where it isn't needed

tsujan:
Don't show "untrust" emblem where it isn't needed

The patch does two things:

 1. It fixes the logic of "trust" and trusts all non-executable files. As a result, desktop entries inside `menu://applications/` or Trash don't have emblems anymore.

 2. It removes the "untrust" emblem from desktop entries inside `~/.local/share/applications` (as Dolphin does), although they aren't trusted by default (like in `/usr/share/applications`).

### DIFF
--- a/src/core/fileinfo.cpp
+++ b/src/core/fileinfo.cpp
@@ -405,8 +405,9 @@ bool FileInfo::isTrustable() const {
                 return (strcmp(data, "true") == 0);
             }
         }
+        return false;
     }
-    return false;
+    return true;
 }
 
 void FileInfo::setTrustable(bool trust) const {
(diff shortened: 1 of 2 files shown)


agaida:
i'm not nitpicking - but do you think that trusting ~/.local/share/applications per default is the right thing to do?

tsujan:
> do you think that trusting ~/.local/share/applications per default is the right thing to do?

Its files are still in the "untrust" state and the user can trust them if he wants -- exactly like `/usr/share/applications`.

I saw that in Dolphin. Dolphin's behavior is mostly logical (not always, of course).

agaida:
ok, was just curious about your first message

> It trusts desktop entries inside ~/.local/share/applications too.

tsujan:
Oh, thanks! My PR comment is wrong! I'm correcting it now...

tsujan:
Done, both in the commit and on this page.

PR #459, opened 2019-08-20
Only select first columns on selecting all items of a list view

tsujan:
This is the final fix.

The previous fix was incomplete because `QAbstractItemView::selectAll()` can be called directly by Qt, e.g, in `FileDialog` (see Qt → qabstractitemview.cpp → keyPressEvent).

Fixes https://github.com/lxqt/libfm-qt/issues/458

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -288,6 +288,19 @@ void FolderViewListView::activation(const QModelIndex& index) {
     }
 }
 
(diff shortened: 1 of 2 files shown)


tsujan:
I think it's better to recompile pcmanfm-qt and other libfm-qt based apps after this.

agaida:
GTM

PR #460, opened 2019-08-21
Wait: Fixed problems in graying-out of cut files

tsujan:
Fixes https://github.com/lxqt/libfm-qt/issues/457

Cut files are grayed out almost instantly once the signal `QClipboard::dataChanged` is emitted, even when their number is huge or they're cut by another app (based on libfm-qt or not).

The structure of the graying-out code is simplified, so that its process is started in `Fm::FolderModel` by connecting to `QClipboard::dataChanged` but the necessary updates are done by `Fm::Folder::updateCutFiles()`, which is very fast.

Please note that, when cutting is done in *another* app, `QClipboard::dataChanged` isn't always reliable: the signal may be emitted with a delay or may not be emitted until the cursor enters the window. That's a Qt issue and is especially visible in Qt 5.13 but it isn't so important for us because we want it to work with different folder models of the same app.

**NOTE:** All libfm-qt based apps should be recompiled.

P.S. IMO, this can wait until 0.15 is released.

### DIFF
--- a/src/core/fileinfo.h
+++ b/src/core/fileinfo.h
@@ -220,6 +220,10 @@ public:
 
     void setTrustable(bool trust) const;
 
+    GObjectPtr<GFileInfo> gFileInfo() const {
+        return inf_;
+    }
+
 private:
     GObjectPtr<GFileInfo> inf_;
     std::string name_;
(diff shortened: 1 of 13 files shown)


agaida:
@tsujan - merge or not merge now?

tsujan:
> @tsujan - merge or not merge now?

When a regression happens in Qt, I say, "Damn! They even don't test their patches!" I don't like a similar thing for LXQt ;) IMO, such a deep change needs more testing (by me and by those using the git version), although it has worked without problem.

tsujan:
@agaida 
After a week, I haven't found a single problem. So, I think it's safe to merge.

PR #466, opened 2019-09-07
Fixed a crash in renaming bookmark with dot in its name

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/1009

The context object was missing in a lambda connection.


tsujan:
It's straightforward but important; hence merging...

PR #472, opened 2019-09-22
Fix FTBFS with the latest lxqt-build-tools

tsujan:
Closes https://github.com/lxqt/libfm-qt/issues/471

NOTE: Will check libfm-qt based apps today.


tsujan:
libfm-qt based apps are OK. Merging...

PR #475, opened 2019-09-29
Select the current folder on showing the directory tree

tsujan:
Previously, when the directory tree was shown on the side pane, the current folder wasn't selected because the code didn't wait for the root items (`/home` and `/`) to be added.

Fixes https://github.com/lxqt/pcmanfm-qt/issues/1017

NOTE: Please recompile pcmanfm-qt after this!


tsujan:
@agaida Please test it when you find the time! I was half awake, half asleep. It worked but maybe in my dream.

PR #478, opened 2019-10-08
Corrected path button "hidden" names

tsujan:
Each path button has 2 names, one of which is shown on it and the other one is used to get the full path out of its components (between slashes). In most cases, these strings are identical but in the case of trash:/// and maybe some other places, they may be different. This patch, gets the hidden name in such a way that it can be used to remake the full path correctly.

Fixes https://github.com/lxqt/pcmanfm-qt/issues/1020

### DIFF
--- a/src/pathbar.cpp
+++ b/src/pathbar.cpp
@@ -261,9 +261,13 @@ void PathBar::setPath(Fm::FilePath path) {
             name = btnPath.toString();
         }
         else {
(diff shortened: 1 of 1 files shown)


tsujan:
@agaida, I'm reasonably sure that this is correct. But, IMHO, it's better to test it for a few days (by opening trahsed and other folders while path buttons are shown).

PR #498, opened 2019-12-19
Support for external copier as Ultracopier

alphaonex86:


tsujan:
Thanks for trying to make libfm-qt better but file copying/moving are very sensitive operations and I really don't think we should do this. The least that could happen is a confusion in bug reports.

However, most importantly:

libfm-qt can use external apps for jobs *it doesn't do* — like archiving and terminal emulation — but copy/move/delete operations are libfm-qt's job; letting them be done by another app *inside libfm-qt* doesn't make sense to me.

alphaonex86:
Hi,
Copy/move into libfm-qt is very buggy, it's why I wish use Ultracopier (Ultracopier is mature and used by 1 Millions user daily). The user nerver report bug to Windows (I have 95% Windows user) because it install explicitly Ultracopier, start it explicitly and clearly mentioned.
As you see, I have do dynamic detection, if Ultracopier is no started, then use internal libfm-qt function, **then no change when ultracopier is not started**.
**Replace copy/move it's the most wanted features on any OS**.
Maybe do you have better solution to do that's?
Cheers,

tsujan:
> Copy/move into libfm-qt is very buggy

This is how contribution is done in LXQt:

Probable bugs are reported and fixed. If you want to improve libfm-qt's code in any way (by fixing bugs or adding features), your PRs are very welcome.

But, for many reasons that come to mind easily, using an external app to do libfm-qt's job in its code is not an option.

alphaonex86:
@palinek
> Any possibly blocking operation must not be performed on the UI thread.
I'm right with you
>  it must be configurable by user (preferences)
Then why not directly not do a combobox, copy software: Internal, or binary path?

@tsujan
> Probable bugs are reported and fixed. If you want to improve libfm-qt's code in any way (by fixing bugs or adding features), your PRs are very welcome.
Copy and paste the code of Ultracopier (GPL3) into libfm-qt's code is no sense and do useless more to maintain. I need: full copy list management, speed limitation, pause, chart about the copy, advanced error management, ...
> But, for many reasons that come to mind easily
Can you express this reasons point by point? Maybe I be agree, maybe I have a solution.
Actually the user ask about this, Windows allow this and every user is happy.

tsujan:
> Copy and paste the code of Ultracopier (GPL3) into libfm-qt's code is no sense

Yes, of course, and it wouldn't be merged if done (see below).

> full copy list management, speed limitation, pause, chart about the copy, advanced error management, …

The argument if not about how good your app is — it may be the best app of its kind in the whole world — but about ruining libfm-qt's code. Your patch does that.

> Can you express this reasons point by point?

I don't have time for a long talk. Instead, you could see so many PRs that are merged and get an idea.

> Windows allow this and every user is happy.

Again, that's off the point.

agaida:
I follow @tsujan and closing this PR

PR #505, opened 2020-01-10
rename-dialog.ui: avoid translating placeholders

Fat-Zer:
Mark several placeholders strings as untranslatable so
avoid their pointless appearance on weblate/translation files.


tsujan:
GTM

PR #507, opened 2020-01-11
Fix filename renaming google drive and correct glib solution

damianatorrpm:
~~Fix non-UTF8 filename renaming~~
Provides the correct glib solution
Fixes google drive file renaming (dialog showing file name)
First Part of Fix for https://github.com/lxqt/libfm-qt/issues/138

Uses g_file_get_edit_name()
If there is a problem with the commit formats feel free to make your own PR,
with any git format you like.

### DIFF
--- a/src/core/fileinfo.cpp
+++ b/src/core/fileinfo.cpp
@@ -43,7 +43,8 @@ void FileInfo::setFromGFileInfo(const GObjectPtr<GFileInfo>& inf, const FilePath
         name_ = name;
 
     dispName_ = QString::fromUtf8(g_file_info_get_display_name(inf.get()));
-
+    editName_ = QString::fromUtf8(g_file_info_get_edit_name(inf.get()));
+  
     size_ = g_file_info_get_size(inf.get());
 
     tmp = g_file_info_get_content_type(inf.get());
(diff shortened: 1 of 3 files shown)


tsujan:
This is a good fix but needs a little work.

> If there is a problem with the commit formats feel free to make your own PR,

From a previous comment of yours, I guess you have problem with git. Since the fix is important, I'll make a superseding PR.

damianatorrpm:
> This is a good fix but needs a little work.
> 
> > If there is a problem with the commit formats feel free to make your own PR,
> 
> From a previous comment of yours, I guess you have problem with git. Since the fix is important, I'll make a superseding PR.
I do not use git. All these playing squash stuff is if you have a local git repo which I do not
and I won't change the buildsystem which works really well. The patches I have get auto-applied.

If I have a patch which upstream will benefit from I edit the files in github GUI and press commit 
(if it help I can edit the commits withing github GUI)

tsujan:
Superseded by https://github.com/lxqt/libfm-qt/pull/508

@damianatorrpm 
Git is not that terrible. Ignore its first impression! As an active contributor, you need to use command-line and master branching, cloning (branches), squashing and, especially, rebasing.

tsujan:
Oh, I forgot this:

@damianatorrpm
First, thanks for the fix! Second, please see whether https://github.com/lxqt/libfm-qt/pull/508 fixes the problem -- I don't have *any* file with invalid encoding in its name.

Issue #138, opened 2017-11-20
renaming files in google-drive:// - operation not supported


damianatorrpm:
renaming files in google-drive:// - operation not supported
(is supported by gvfs/nautilus)

damianatorrpm:
This is fixed #511

tsujan:
> This is fixed #511

My mistake; I had to add `closes` to the PR comment.

PR #508, opened 2020-01-11
Handle renaming of files, whose names have invalid encodings

tsujan:
The original fix was made by @damianatorrpm (see https://github.com/lxqt/libfm-qt/pull/507).

Closes https://github.com/lxqt/libfm-qt/issues/138

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -963,7 +963,11 @@ void FolderView::onClosingEditor(QWidget* editor, QAbstractItemDelegate::EndEdit
         QVariant data = index.model()->data(index, FolderModel::FileInfoRole);
         auto info = data.value<std::shared_ptr<const Fm::FileInfo>>();
         if (info) {
-            auto oldName = QString::fromStdString(info->name());
+            // NOTE: "Edit name" is used to handle invalid filename encoding.
+            auto oldName = QString::fromUtf8(g_file_info_get_edit_name(info->gFileInfo().get()));
+            if(oldName.isEmpty()) {
+                oldName = QString::fromStdString(info->name());
+            }
             if(newName == oldName) {
                 return;
             }
--- a/src/utilities.cpp
+++ b/src/utilities.cpp
@@ -171,8 +171,11 @@ bool renameFile(std::shared_ptr<const Fm::FileInfo> file, QWidget* parent) {
     FilenameDialog dlg(parent ? parent->window() : nullptr);
     dlg.setWindowTitle(QObject::tr("Rename File"));
     dlg.setLabelText(QObject::tr("Please enter a new name:"));
-    // FIXME: what's the best way to handle non-UTF8 filename encoding here?
-    auto old_name = QString::fromStdString(file->name());
+    // NOTE: "Edit name" seems the best way to handle non-UTF8 filename encoding.
+    auto old_name = QString::fromUtf8(g_file_info_get_edit_name(file->gFileInfo().get()));
+    if(old_name.isEmpty()) {
+        old_name = QString::fromStdString(file->name());
+    }
     dlg.setTextValue(old_name);
 
     if(file->isDir()) { // select filename extension for directories


damianatorrpm:
It does not close https://github.com/lxqt/libfm-qt/issues/138 just yet it shows the correct edit name to edit. The second part will be in a different PR implementing similar for Fm::FileOperation

tsujan:
> The second part will be in a different PR implementing similar for Fm::FileOperation

I'll wait for you to tell me why `g_file_move` doesn't work and what does -- as I said elsewhere, all file names have correct encodings here. In the meanwhile, I might consult Nautilus' code if I find the time.

Please also note that we don't want a too complex workaround.

damianatorrpm:
I already have an idea about that the most complex solution I can imagine would be 
an if (uriScheme=='google-drive') but I will look into this after setting up an issuer report and the gym.

tsujan:
As you might have guessed my reply, we shouldn't make an exception for google-drive.

We rename files by using `g_file_move`, which probably doesn't work with google-drive. Other ways of renaming are `g_file_set_display_name`  and `g_file_set_display_name_async`. To use them instead of `g_file_move`, we'll need a lot of code change and, more importantly, a lot of tests because we don't want to damage what works just for a special case.

I'm really not sure that it's worth the effort but I agree that it's worth considering.

I close this because it doesn't add anything to the current functionality. A clean solution will need a code restructuring by using `g_file_set_display_name(_async)`.

tsujan:
Update:

I took a very simple approach, replaced `g_file_move` with `g_file_set_display_name` and, contrary to what the functions name said, the *real* file name was changed, not its display name.

That's a little confusing but if `g_file_set_display_name` changes the real file name, it might be good for us. However, your test is needed to see if it solves https://github.com/lxqt/libfm-qt/issues/138.

To test, use this patch but change `changeFileName` as follows:

```c++
bool changeFileName(const Fm::FilePath& filePath, const QString& newName, QWidget* parent, bool showMessage) {
    Fm::GErrorPtr err;
    if(!g_file_set_display_name(filePath.gfile().get(),
                                newName.toLocal8Bit().constData(),
                                nullptr, /* make this cancellable later. */
                                &err)) {
        if (showMessage){
            QMessageBox::critical(parent ? parent->window() : nullptr, QObject::tr("Error"), err.message());
        }
        return false;
    }

    // reload the containing folder if it is in use but does not have a file monitor
    auto folder = Fm::Folder::findByPath(filePath.parent());
    if(folder && folder->isValid() && folder->isLoaded() && !folder->hasFileMonitor()) {
        folder->reload();
    }

    return true;
}
```

damianatorrpm:
> I took a very simple approach, replaced `g_file_move` with `g_file_set_display_name` and, contrary to what the functions name said, the _real_ file name was changed, not its display name.

I don't know what that with the Fm::FileOperation was (I have other work there).
Of course you are right and regarding the fix for https://github.com/lxqt/libfm-qt/issues/138
only the `changeFileName` function requires chaning.

> That's a little confusing but if `g_file_set_display_name` changes the real file name, it might be good for us. However, your test is needed to see if it solves #138.

Yes! It does change the file name now correctly in google drive.
I believe the reason for this behavior is that `g_file_set_display_name` 
sets the value of what the user visible file name should be and backends like google drive
which use internal hashes as real file name get adjusted accordingly on `g_file_set_display_name`.
Backends where display_name equals the real name i does the right thing and the function
adjusts the real name.

> To test, use this patch but change `changeFileName` as follows:
> 
> ```c++
> bool changeFileName(const Fm::FilePath& filePath, const QString& newName, QWidget* parent, bool showMessage) {
>     Fm::GErrorPtr err;
>     GFilePtr gfile{g_file_set_display_name(filePath.gfile().get(),
>                                             newName.toLocal8Bit().constData(),
>                                             nullptr, /* make this cancellable later. */
>                                             &err)};
>     if(gfile == nullptr) {
>         if (showMessage){
>             QMessageBox::critical(parent ? parent->window() : nullptr, QObject::tr("Error"), err.message());
>         }
>         return false;
>     }
> 
>     // reload the containing folder if it is in use but does not have a file monitor
>     auto folder = Fm::Folder::findByPath(filePath.parent());
>     if(folder && folder->isValid() && folder->isLoaded() && !folder->hasFileMonitor()) {
>         folder->reload();
>     }
> 
>     return true;
> }
> ```
> 
> EDIT: Corrected a mistake in the above code.

I think it may be possible  that the g_file_set_display_name reutrns an error but the gfile exists?
I am curious why there is GFilePtr when there is g_autoptr(GFile) but that does not matter that much.
GFilePtr works well.

This works also but I am no sure if it is possible that the gfile exists but an error occured.
line 1 and line 6 changed
 
> ```c++
> bool changeFileName(const Fm::FilePath& filePath, const QString& newName, QWidget* parent, bool showMessage) {
>     Fm::GErrorPtr err = nullptr;
>     GFilePtr gfile{g_file_set_display_name(filePath.gfile().get(),
>                                             newName.toLocal8Bit().constData(),
>                                             nullptr, /* make this cancellable later. */
>                                             &err)};
>     if(gfile == nullptr || err != nullptr) {
>         if (showMessage){
>             QMessageBox::critical(parent ? parent->window() : nullptr, QObject::tr("Error"), err.message());
>         }
>         return false;
>     }
> 
>     // reload the containing folder if it is in use but does not have a file monitor
>     auto folder = Fm::Folder::findByPath(filePath.parent());
>     if(folder && folder->isValid() && folder->isLoaded() && !folder->hasFileMonitor()) {
>         folder->reload();
>     }
> 
>     return true;
> }
> ```

tsujan:
> only the changeFileName function requires chaning.

The other changes are good too. GLib doc recommends them.

> Yes! It does change the file name now correctly in google drive.

Thanks for testing! I'll make a PR.

> This works also but I am no sure if it is possible that the gfile exists but an error occured.

GLIb doc explicitly says the gfile is NULL if an error occurs. So, there's no need to checking the error.

tsujan:
@agaida Will libfm-qt be released soon? I ask because the new file renaming code might need more tests (I haven't made a PR yet).

agaida:
i think end of february

tsujan:
> i think end of february

Good.

I think https://github.com/lxqt/libfm-qt/pull/511 is not only safe but a big step forward.

agaida:
so i can merge #511 ?

tsujan:
> so i can merge #511 ?

Yes! It's good.

PR #511, opened 2020-01-12
Use GIO's display/copy name when renaming/transferring files

tsujan:
Because, otherwise, renaming will not be possible in places like google-drive:/// and files in places like recent:/// or google-drive:/// will be transferred with their real names.

All credit goes to @damianatorrpm, who found the problem and its best solution.

Closes https://github.com/lxqt/libfm-qt/issues/137 and https://github.com/lxqt/libfm-qt/issues/138

NOTE: This may need more tests (by me). Please don't merge it soon if the next release comes in a few weeks.

### DIFF
--- a/src/core/filetransferjob.cpp
+++ b/src/core/filetransferjob.cpp
@@ -383,18 +383,24 @@ bool FileTransferJob::processPath(const FilePath& srcPath, const FilePath& destD
         return false;
     }
 
+    // Use GIO's copy name for destination if existing. This is especially good for copying
+    // from another file system or from places like recent:/// and also handles encoding.
+    const char* destCopyName = g_file_info_get_attribute_string(srcInfo.get(), "standard::copy-name");
+
     bool ret;
     switch(mode_) {
     case Mode::MOVE:
-        ret = moveFile(srcPath, srcInfo, destDirPath, destFileName);
+        ret = moveFile(srcPath, srcInfo, destDirPath, destCopyName ? destCopyName : destFileName);
         break;
     case Mode::COPY: {
         bool deleteSrc = false;
-        ret = copyFile(srcPath, srcInfo, destDirPath, destFileName, deleteSrc);
+        ret = copyFile(srcPath, srcInfo, destDirPath, destCopyName ? destCopyName : destFileName, deleteSrc);
         break;
     }
     case Mode::LINK:
-        ret = linkFile(srcPath, srcInfo, destDirPath, destFileName);
+        ret = linkFile(srcPath, srcInfo, destDirPath,
+                        // see setDestDirPath()
+                        srcPath.isNative() && destCopyName ? destCopyName : destFileName);
         break;
     default:
         ret = false;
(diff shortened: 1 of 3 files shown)


damianatorrpm:
glib correctly handles this case with / in display name by providing the correct copy name
which should be used as well here. I verified this behaviour with various local / non local / recent
copy move operations

```c++
    const char* copyName = g_file_info_get_attribute_as_string (srcInfo.get(), G_FILE_ATTRIBUTE_STANDARD_COPY_NAME);

    bool ret;
    switch(mode_) {
    case Mode::MOVE:
        ret = moveFile(srcPath, srcInfo, destDirPath, copyName ? copyName : destFileName);
        break;
    case Mode::COPY: {
        bool deleteSrc = false;
        ret = copyFile(srcPath, srcInfo, destDirPath, copyName ? copyName : destFileName, deleteSrc);
        break;
    }
    case Mode::LINK:
        ret = linkFile(srcPath, srcInfo, destDirPath,
                       // see setDestDirPath()
                       srcPath.isNative() && copyName ? copyName : destFileName);
        break;
    default:
        ret = false;
        break;
    }
    return ret;
}
```

tsujan:
> glib correctly handles this case with / in display name by providing the correct copy name

How does GLib replace a slash in the copy name is is used? With a dash or with a fraction slash?

damianatorrpm:
with dash

damianatorrpm:
[tux@localhost google-drive:host=gmail.com,user=damianatorrpm]$ gio info 1jy0lriNpz-goELFbQUSN0ntZinULh9Ih
display name: / / / / / / / slash /// / / / 
edit name: / / / / / / / slash /// / / / 
name: 1jy0lriNpz-goELFbQUSN0ntZinULh9Ih
type: directory
uri: google-drive://damianatorrpm@gmail.com/1jy0lriNpz-goELFbQUSN0ntZinULh9Ih
attributes:
  standard::type: 2
  standard::is-symlink: FALSE
  standard::name: 1jy0lriNpz-goELFbQUSN0ntZinULh9Ih
  standard::display-name: / / / / / / / slash /// / / / 
  standard::edit-name: / / / / / / / slash /// / / / 
  standard::copy-name: - - - - - - - slash --- - - -

tsujan:
Replacement with dash isn't so good because I want to implement Dolphin's behavior in this PR (if the new file name has slash, Dolphin changes it to "fraction slash"). But it's tolerable and, especially, "copy name" seems to be the safest option.

I'm very busy now; will consider all options and add a commit to the current PR later. Thanks for the discussion!

damianatorrpm:
@tsujan I understand. I prefer the dash behavior so I have done no further actions regarding this since it's the safest option and makes most sense to me and is what I want to use in the file manager for the soon upcoming distro. 

I wish you good luck if you think that the other behavior is better! 
I will have a look into the memory stuff and wish you good fortune until than!

> I'm very busy now; will consider all options and add a commit to the current PR later. Thanks for the discussion!

<>

tsujan:
Done. I updated the PR so that "copy name" is used for file transferring.

@damianatorrpm You're completely right about "copy name". We should follow GLib standards in our core code.

agaida:
@tsujan - ready to merge?

tsujan:
> @tsujan - ready to merge?

Yes, it's ready.

PR #516, opened 2020-01-21
Better DND from remote folders to external apps

tsujan:
Previously, only apps that could handle remote folders were able to receive a DND from them by using pcmanfm-qt (and other libfm-qt apps). For example, `lxqt-archiver` and `gedit` could do that with `smb` or `sftp`.

This patch doesn't change the internal DND of libfm-qt (for the sake of certainty) but makes use of local folders with DNDing to external apps as far as possible. In that way, for example, FeatherPad and Dolphin can also accept DND from a remote folder opened by pcmanfm-qt.

Nautilus has a similar behavior and, in general, it is better for external apps.

Closes https://github.com/lxqt/libfm-qt/issues/515

~~NOTE: I made the PR for it to be used by testers. So, please let it be tested for a while before merging (ask me)!~~

EDIT: After 3 weeks, all tests show that the PR is OK.


agaida:
LGTM

PR #519, opened 2020-02-01
Also open remote links in desktop entries

tsujan:
This patch just considers a case that was missing from the code: When, in a desktop entry, `Type=Link` and `URL` points to a remote place like `https://` or `sftp://`, `smb://`, etc., then libfm-qt will try to open the link with its handling app if there is one.

NOTE: Handlers can be defined in `~/.config/mimeapps.list` like:

```
[Default Applications]
…
x-scheme-handler/https=firefox.desktop
x-scheme-handler/smb=pcmanfm-qt.desktop
x-scheme-handler/sftp=pcmanfm-qt.desktop
…
```

I think we might want to add `x-scheme-handler/smb` and `x-scheme-handler/sftp` to pcmanfm-qt's desktop file later.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1065

### DIFF
--- a/src/core/basicfilelauncher.cpp
+++ b/src/core/basicfilelauncher.cpp
@@ -207,8 +207,11 @@ bool BasicFileLauncher::launchDesktopEntry(const FileInfoPtr &fileInfo, const Fi
             else {
                 if(target.empty()) {
                     filename = fileInfo->path().localPath();
+                    desktopEntryName = filename.get();
                 }
-                desktopEntryName = !target.empty() ? target.c_str() : filename.get();
+                else {
+                    desktopEntryName = target.c_str();
+                }
             }
             break;
         }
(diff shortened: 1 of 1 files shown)


wxl:
LGTM. Pretty simple, really.

Agree with your idea about adding handlers. We might want to look at the [IANA URI schemes](https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml) to see if there's anything else we haven't missed. I'm thinking it might be good to add `file` in case someone gets cute. Also do we support `afp` or `nfs`? Or, really crazy: `sshfs`?

tsujan:
> We might want to look at the IANA URI schemes to see if there's anything else we haven't missed.

I agree. I just don't know why other FMs don't do it (or do they?).

`file` is automatically covered.

wxl:
Here's stuff relevant to the file manager:
Gvfs: `afc` (not IANA), `afp`, `cdda` (not IANA), `dav`, `ftp`, `mtp` (not IANA), `nfs`, `sftp`, `smb`
KIO: `dav`, `fish`, `ftp`, `nfs`, `sftp`, `smb`

tsujan:
Opening smb and sftp with a FM is expected but I'm not sure about the other ones. I think we may need to discuss it as a separate issue (feature request) elsewhere. What do you think?

wxl:
I mention those because they all ultimately involve files (I guess `dav` is probably not appropriate in retrospect). Nearly all of them are some sort of remote filesystem (like `smb` and `sftp`). Either way, I agree a separate issue should be opened. See lxqt/pcmanfm-qt#1067.

PR #527, opened 2020-02-21
Avoid hard-coded paths

voidpin:
While packaging LXQt for NetBSD I came across a few hard-coded paths.
Please review.

### DIFF
--- a/src/core/fileinfo.cpp
+++ b/src/core/fileinfo.cpp
@@ -359,11 +359,11 @@ bool FileInfo::isExecutableType() const {
         if(isNative() && (mode_ & (S_IRUSR|S_IRGRP|S_IROTH))) {
             if(isShortcut() && !target_.empty()) {
                 /* handle shortcuts from desktop to menu entries:
-                   first check for entries in /usr/share/applications and such
+                   first check for entries in ${PREFIX}/share/applications and such
                    which may be considered as a safe desktop entry path
                    then check if that is a shortcut to a native file
                    otherwise it is a link to a file under menu:// */
-                if (!g_str_has_prefix(target_.c_str(), "/usr/share/")) {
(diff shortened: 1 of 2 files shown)


src/core/fileinfo.cpp
```diff
@@ -359,11 +359,11 @@ bool FileInfo::isExecutableType() const {
         if(isNative() && (mode_ & (S_IRUSR|S_IRGRP|S_IROTH))) {
             if(isShortcut() && !target_.empty()) {
                 /* handle shortcuts from desktop to menu entries:
-                   first check for entries in /usr/share/applications and such
+                   first check for entries in ${PREFIX}/share/applications and such
                    which may be considered as a safe desktop entry path
                    then check if that is a shortcut to a native file
                    otherwise it is a link to a file under menu:// */
-                if (!g_str_has_prefix(target_.c_str(), "/usr/share/")) {
+                if (!g_str_has_prefix(target_.c_str(), "${PREFIX}/share/")) {
```
palinek: What is `${PREFIX}` and how would it be evaluated in compile/run-time?

voidpin: On Linux is /usr
on NetBSD is /usr/pkg
on FreeBSD is /usr/local

Because of these differences hard-coding it causes portability issues. I don't have a Linux machine right now but on NetBSD ${PREFIX} will be evaluated as /usr/pkg, in this case /usr/pkg/share will be the result.
I can't test on Linux, that's why I asked to please review.

palinek: > ${PREFIX} will be evaluated as /usr/pkg

Who and when will it evaluate?

tsujan:
`${PREFIX}` doesn't have meaning and this breaks the code.

PR #528, opened 2020-02-21
Fix typo in hard-coded paths ("xdg" not "xgd")

dhgutteridge:
Assuming this is required, this should reference "xdg" not "xgd".


tsujan:
Yes, `xdg` is correct.

tsujan:
It should be related to app menu, whose code is old and should be replaced sooner or later.

PR #530, opened 2020-02-24
Sharp view icons with scale factors > 1

tsujan:
Since we draw view icons ourself, to have sharp view icons, we shouldn't get their pixmaps and use `QPainter::drawPixmap()` because — especially with SVG icons — the pixmap sizes may be too big. Instead, the current patch uses `QIcon::paint()`, letting the icon engine do the painting. In this way, view icons are always sharp, even with non-integer scale factors.


tsujan:
This needs https://github.com/lxqt/libqtxdg/pull/203 to work fine but the latter is needed for all apps. So, merging (also, see https://github.com/lxqt/lximage-qt/pull/306#issuecomment-612951827).

PR #533, opened 2020-03-26
Fixed disabled file DND in rare cases

tsujan:
We should call `QListView::setMovement()` only before showing the view; otherwise, in rare cases, the DND will be disabled.

Fixes https://github.com/lxqt/pcmanfm-qt/issues/1088


tsujan:
Merging (→ https://github.com/lxqt/lximage-qt/pull/306#issuecomment-612951827)

PR #534, opened 2020-04-02
Use g_list_free_full instead of casting function pointers

alexhenrie:
This is more straightforward and avoids both compiler warnings (with Clang) and the problems that could result if the code is compiled for a platform with an unusual calling convention.

### DIFF
--- a/src/appchooserdialog.cpp
+++ b/src/appchooserdialog.cpp
@@ -174,8 +174,7 @@ GAppInfo* AppChooserDialog::customCommandToApp() {
                 }
                 g_free(bin2);
             }
-            g_list_foreach(apps, (GFunc)g_object_unref, nullptr);
-            g_list_free(apps);
+            g_list_free_full(apps, g_object_unref);
             if(app) {
                 goto _out;
             }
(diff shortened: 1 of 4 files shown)


tsujan:
They're completely equivalent to each other. See `glib/glist.c`:
```c
void
g_list_free_full (GList          *list,
                  GDestroyNotify  free_func)
{
  g_list_foreach (list, (GFunc) free_func, NULL);
  g_list_free (list);
}
```

alexhenrie:
> They're completely equivalent to each other. See `glib/glist.c`:
> 
> ```c
> void
> g_list_free_full (GList          *list,
>                   GDestroyNotify  free_func)
> {
>   g_list_foreach (list, (GFunc) free_func, NULL);
>   g_list_free (list);
> }
> ```

Wow, I guess GLib doesn't support left-to-right argument passing. It must be too niche of a thing to worry about. Still, the code is cleaner with g_list_free_full, so thanks for merging this!

tsujan:
> Still, the code is cleaner with g_list_free_full

I agree. And, IMO, we don't need to worry about very old versions of GLib that don't support `g_list_free_full`.

PR #535, opened 2020-04-08
Allow saving mount passwords if gnome-keyring is present

tsujan:
Closes https://github.com/lxqt/lxqt/issues/774

NOTE: It works here when gnome-keyring is installed. Please test it *without* gnome-keyring; I think it should do nothing then.


McThump:
Thanks. This fixed the issue on all my test machines--goodbye Nautilus. When gnome-keyring is not installed, the credentials are not saved and no error is displayed.

tsujan:
@McThump 
Thanks for your test!

>  When gnome-keyring is not installed, the credentials are not saved and no error is displayed.

That was what I needed to know -- it's the expected behavior.

I merge it now...

PR #545, opened 2020-06-01
Update text color of places view if style changes in runtime

tsujan:
Otherwise, if the style is switched between dark and light in runtime (which will soon be possible in LXQt), the text color will not change and might not have enough contrast with its background.

### DIFF
--- a/src/sidepane.cpp
+++ b/src/sidepane.cpp
@@ -19,6 +19,7 @@
 
 
 #include "sidepane.h"
+#include <QEvent>
 #include <QComboBox>
 #include <QVBoxLayout>
 #include <QHeaderView>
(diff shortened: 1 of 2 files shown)


tsujan:
Preferably recompile libfm-qt based apps after applying the patch, although I saw no difference.

damianatorrpm:
but this doesn't work for dialogs?

tsujan:
> but this doesn't work for dialogs?

It does. It's in libfm-qt, which creates the place view, whether in dialogs or in pcmanfm-qt.

> ... at least via qt5ct it works for dialogs

OK but, in general, when you use qt5ct, you override LXQt's Qt plugin, so that what you see may not be what is seen with a standard LXQt installation. Problems that qt5ct may cause aren't in LXQt.

PR #547, opened 2020-06-13
Added a missing nullity check to file dialog code

tsujan:
Although, in the current code, the nullity check wasn't needed, its lack was a potential cause of crash — I checked that by using `qDebug()` after I changed the code during an experiment.

Long story short and contrary to appearances, the variable `folder_` can be null at the end of `FileDialog::setDirectoryPath`.


tsujan:
Merging because it's straightforward and I have a lot of work to do with this file (for remembering sorting and...).

PR #548, opened 2020-06-14
File dialog: Remember sorting and showing of hidden files

tsujan:
WARNING: pcmanfm-qt and lxqt-archiver should be recompiled because they use the file dialog directly (pcmanfm-qt in its wallpaper browser and lxqt-archiver everywhere).

NOTE: This is for LXQt's native file dialog when called by any Qt app, not for hard-coded file dialogs like those of pcmanfm-qt's wallpaper browser or lxqt-archiver's dialogs (which should manage their settings independently if needed).

### DIFF
--- a/src/filedialog.cpp
+++ b/src/filedialog.cpp
@@ -191,6 +191,78 @@ void FileDialog::setSplitterPos(int pos) {
     ui->splitter->setSizes(sizes);
 }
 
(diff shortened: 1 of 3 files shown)


tsujan:
Merging after 3 weeks of testing. An upcoming patch will need it.

PR #555, opened 2020-07-09
Fixed setting of default app in Properties dialog

tsujan:
GLib has a weird problem. On the one hand, `g_app_info_set_as_default_for_type()` writes to `~/.config/mimeapps.list`. On the other hand, the DE-specific list — e.g., `~/.config/lxqt-mimeapps.list` in LXQt — has priority over it and GLib respects that. Therefore, if the latter exists and contains some default apps, `g_app_info_set_as_default_for_type()` could not change them.

This patch solves the problem by adding a utility function that sets the default app directly inside the DE-specific list.

Also, see https://github.com/lxqt/libqtxdg/pull/219.

### DIFF
--- a/src/appchooserdialog.cpp
+++ b/src/appchooserdialog.cpp
@@ -20,6 +20,7 @@
 
 #include "appchooserdialog.h"
 #include "ui_app-chooser-dialog.h"
+#include "utilities.h"
 #include <QPushButton>
 #include <gio/gdesktopappinfo.h>
 #include <glib/gstdio.h>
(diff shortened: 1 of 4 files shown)


tsujan:
@luis-pereira 
This is very similar to https://github.com/lxqt/libqtxdg/pull/219 and is needed alongside it. There's only one difference: here we don't need to add associations first because they either already exist (otherwise they wouldn't show up in Properties dialog) or were added by the code before (see https://github.com/lxqt/libfm-qt/blob/928a85a173ab889e584a43df1eef54c37ee37322/src/appchooserdialog.cpp#L241 
 `g_app_info_set_as_last_used_for_type()` automatically adds the association; PCMan's comment: "add this app to the mime-type").

Since libfm-qt is the place where we use GLib most actively, there isn't a need to an extra dependency on `libqtxdg`.

PR #560, opened 2020-08-18
Translated using Weblate (Italian)

stefonarch:
Currently translated at 99.6% (286 of 287 strings)

Translation: LXQt-standreas/libfm-qt
Translate-URL: http://135.181.85.100/projects/lxqt/libfm-qt/it/


tsujan:
Congratulations! Keep up the good work!

Looking forward to more Weblate translations...

PR #564, opened 2020-08-23
Cleaned up the code for graying out cut files

tsujan:
Several redundant variables/methods were left from the first implementation. Their removal enhances code readability because their logic was absurdly complex and some of them did nothing.

There is no practical change in functionality though: cut files are grayed out inside a single folder — no multi-folder support yet but this cleanup may pave the way for its implementation.

WARNING: pcmanfm-qt and other libfm-qt based apps should be recompiled after this.

Closes https://github.com/lxqt/libfm-qt/issues/563

### DIFF
--- a/src/core/folder.cpp
+++ b/src/core/folder.cpp
@@ -34,8 +34,6 @@
 namespace Fm {
 
 std::unordered_map<FilePath, std::weak_ptr<Folder>, FilePathHash> Folder::cache_;
-QString Folder::cutFilesDirPath_;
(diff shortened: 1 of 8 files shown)


tsujan:
This can be much more simplified. In fact, there's no need to contaminate the core of libfm-qt. Its only advantage was the hash set, which made the CPU usage negligible, but it can be preserved in another form.

Will make a superseding PR.

tsujan:
Superseded by https://github.com/lxqt/libfm-qt/pull/566

PR #566, opened 2020-08-26
Better code for graying out cut files

tsujan:
The current patch simplifies and enhances the graying-out of cut files considerably by removing all traces of cut files from the core of libfm-qt and doing the job at the level of `FolderModel`. In this way, a better functionality is achieved by a much simpler logic.

Since a hash set is used for tracking cut files (instead of a path vector), the CPU usage is negligible even with huge numbers of cut files.

In addition:

 * The patch grays out cut files in multiple folders too (e.g., when multiple files are cut from a recursive search result).
 * Cut files are also grayed out in folders without file monitor (except for `search://`, which isn't a real folder).

WARNING: The changes affect both ABI and API. Therefore, **pcmanfm-qt and all other libfm-qt based apps should be cleanly recompiled after this.**

Closes https://github.com/lxqt/libfm-qt/issues/563

### DIFF
--- a/src/core/dirlistjob.cpp
+++ b/src/core/dirlistjob.cpp
@@ -6,8 +6,8 @@
 
 namespace Fm {
 
-DirListJob::DirListJob(const FilePath& path, Flags _flags, const std::shared_ptr<const HashSet>& cutFilesHashSet):
-    dir_path{path}, flags{_flags}, cutFilesHashSet_{cutFilesHashSet} {
+DirListJob::DirListJob(const FilePath& path, Flags _flags):
+    dir_path{path}, flags{_flags} {
 }
 
 void DirListJob::exec() {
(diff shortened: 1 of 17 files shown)


tsujan:
This is a big cleanup as well as enhancement and I want to merge it. Any objection?

palinek:
I'm glad to see the previous mess implementation to be removed (I was even surprised merging the mess previously). Go ahead with merging this.... even if there would be little problem it is way more easy to track it down with this implementation.

tsujan:
> I'm glad to see the previous mess implementation to be removed

Yes, it was a mess. I can't make up a good excuse for merging it -- I might have been hypnotized by the contributor's work and how he managed to do it ;) Since then, I've become much more cautious in reviewing PRs.

> even if there would be little problem...

I haven't found any but yes, problems may always surface and we'll fix them.

PR #567, opened 2020-08-27
File dialog: open mount targets of mountable directories

tsujan:
Previously, the mount targets weren't checked and, for example, `computer:///root.link` was opened as an empty directory.

Closes https://github.com/lxqt/libfm-qt/issues/565


tsujan:
This is simple and I want to work on `filedialog.cpp` to implement more dialog settings. Hence merging...

PR #568, opened 2020-08-29
File dialog: Added options for thumbnails, tooltips and icon sizes

tsujan:
And put all view options inside menus.

NOTE: pcmanfm-qt and lxqt-archiver should be recompiled but, to be on the safe side, recompile lximage-qt (and other libfm-qt based apps) too.

Closes https://github.com/lxqt/libfm-qt/issues/349


tsujan:
I've encountered no problem with this. Report it if you find any issue.

PR #571, opened 2020-09-17
More intuitive name sorting by excluding identical extensions

tsujan:
Qt's default sorting can be counterintuitive for files (e.g., `a_1.txt` comes before `a.txt` with the ascending order) because the concept of "extension" exists for them.

This patch sorts files in a natural way by treating the last dot of their names as the extension separator.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1197


tsujan:
Closed because of a logical inconsistency. See https://github.com/lxqt/pcmanfm-qt/issues/1197#issuecomment-699762430 for a detailed explanation.

PR #573, opened 2020-09-29
Natural (GTK-like) file sorting by treating dot as separator

tsujan:
Qt's default sorting can be counterintuitive for files (e.g., `a_1.txt` comes before `a.txt` with the ascending order). This patch sorts files by treating dot as a separator and comparing split strings with each other.

This is exactly how GTK (file dialog) sorts files; it considers all dots as separators (not just the last dot) and treats folders like files. Most Linux users are accustomed to GTK's file sorting, which is generally more intuitive than Qt's.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1201

### DIFF
--- a/src/proxyfoldermodel.cpp
+++ b/src/proxyfoldermodel.cpp
@@ -185,9 +185,27 @@ bool ProxyFolderModel::lessThan(const QModelIndex& left, const QModelIndex& righ
             }
             break;
         default: {
+            // To have a more natural sorting like that of GTK, we consider dot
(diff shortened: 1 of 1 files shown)


tsujan:
This has fulfilled my expectations of a "natural" sorting. Will merge it in a week if there's no objection.

Fortunately, I was wrong about one thing: the new sorting isn't the same as GTK's; it's better :) Check how GTK sorts `_b` and `a` (a very old bug that Gnome devs ignore).

PR #577, opened 2020-10-17
Avoid QListView's DND methods in our list view

tsujan:
Because, otherwise, they might interfere with DND when the item movement is not static (like on desktop).

This patch re-enables moving files into folders by DND on desktop.
Fixes https://github.com/lxqt/pcmanfm-qt/issues/1210


tsujan:
I'm merging the bug-fix PRs for the upcoming release...

PR #588, opened 2020-11-18
Fix compiling with Qt >=5.15.3

Chiitoo:
Not sure at all if this is the correct way to fix this, but at least it builds, and I haven't noticed anything broken thus far.  :]

I imagine it will compile on lower versions as well, but I have not tested that.

The failure looks like this:

```
/src/filedialog.cpp: In member function ‘void Fm::FileDialog::setMimeTypeFilters(const QStringList&)’:
/src/filedialog.cpp:1112:51: error: use of deleted function ‘QCharRef& QCharRef::operator=(char)’
 1112 |             nameFilter[nameFilter.length() - 1] = ')';
      |                                                   ^~~
In file included from /usr/include/qt5/QtCore/qhashfunctions.h:44,
                 from /usr/include/qt5/QtCore/qlist.h:47,
                 from /usr/include/qt5/QtCore/qhash.h:46,
                 from /usr/include/qt5/QtCore/qdebug.h:45,
                 from /usr/include/qt5/QtCore/QDebug:1,
                 from /src/core/gobjectptr.h:8,
                 from /src/core/filepath.h:5,
                 from /src/filedialog.h:5,
                 from /src/filedialog.cpp:1:
/usr/include/qt5/QtCore/qstring.h:1236:15: note: declared here
 1236 |     QCharRef &operator=(char c) = delete;
      |               ^~~~~~~~
```

### DIFF
--- a/src/filedialog.cpp
+++ b/src/filedialog.cpp
@@ -1109,7 +1109,7 @@ void FileDialog::setMimeTypeFilters(const QStringList& filters) {
                 nameFilter += suffix;
                 nameFilter += QLatin1Char(' ');
             }
-            nameFilter[nameFilter.length() - 1] = ')';
+            nameFilter[nameFilter.length() - 1] = QLatin1Char(')');
         }
         nameFilters << nameFilter;
     }


tsujan:
Thanks! Please wait until Qt 5.15.3 comes here. This needs to be examined.

tsujan (approved):
GTM.

It's strange that Qt didn't see the problem before.

@Chiitoo 
You mean Qt ≥ 5.15.2 (not 5.15.3); don't you? It hasn't come here yet but will soon.

If compilation fails with it, a point release will be needed.

Chiitoo:
Sorry, I should have been more clear about that: the build failure indeed does not happen until 5.15.3, which 'git master' has been for a while now (also with 6.0.0 (beta) I believe).

Figured I'd report it early since I already bumped into it.

tsujan:
But the Qt patch in question was merged before 5.15.2 was released.

Anyway, it makes no difference because the compilation will fail soon in distros like Arch and OpenSUSE if a point release isn't made.

PR #597, opened 2020-12-04
Fixed the localization of LXQt file dialog

tsujan:
Previously, it was localized only in Qt apps that were based on libfm-qt.

Fixed https://github.com/lxqt/libfm-qt/issues/596

### DIFF
--- a/src/filedialoghelper.cpp
+++ b/src/filedialoghelper.cpp
@@ -3,6 +3,7 @@
 #include "libfmqt.h"
 #include "filedialog.h"
 
+#include <QCoreApplication>
 #include <QWindow>
 #include <QDebug>
 #include <QTimer>
@@ -431,6 +432,8 @@ QPlatformFileDialogHelper *createFileDialogHelper() {
     if(!libfmQtContext_) {
         // initialize libfm-qt only once
         libfmQtContext_ = std::unique_ptr<Fm::LibFmQt>{new Fm::LibFmQt()};
+        // add translations
+        QCoreApplication::installTranslator(libfmQtContext_.get()->translator());
     }
     return new Fm::FileDialogHelper{};
 }


stefonarch:
Maybe I missed something, but I see no change here. Recompiled also pcmanfm-qt.

tsujan:
pcmanfm-qt doesn't need recompilation. The app you test (Falkon, FeatherPad,...) should be restarted.

I tested it on two systems (computers) and it worked fine. You may need a log-out and log-in. Please find "the hidden factor" in your case! It can be important.

tsujan:
Here is Falkon in Italian:

falkon

stefonarch:
Did restart session, to be sure. Recompiled all components against newest qt version on arch today too (screengrab was segfaulting).
No idea about else for now, will recheck tomorrow.

stefonarch:
No idea. Recompiled branch, tested even  a clean user and other languages with `LANGUAGE=de fpad` .Always in english.
Qt version here is now 5.15.2-1

ito32bit:
I thought this was a problem with my system. 
I tested it in Japanese under Debian sid (Qt 5.15.1).
Good!
File-dialog_screen-2020-12-05_17-59-11

stefonarch:
Good that you tested too, I doubt much more in me than in tsujan.
Still no clue why it doesn't work here.

tsujan:
@ito32bit, thank you for testing the patch!

@stefonarch, I wait until you find the hidden factor.

luis-pereira (approved):
GTM.
screen4

tsujan:
@luis-pereira, thanks!

stefonarch:
Found my  problem, as usual the reason sits was in front of the screen:
```
 ls -l /usr/lib/ | grep libfm
lrwxrwxrwx   1 root root    20  7 lug 14.45 libfm-extra.so -> libfm-extra.so.4.1.2
lrwxrwxrwx   1 root root    20  7 lug 14.45 libfm-extra.so.4 -> libfm-extra.so.4.1.2
-rwxr-xr-x   1 root root   30K  7 lug 14.45 libfm-extra.so.4.1.2
lrwxrwxrwx   1 root root    13  4 nov 09.28 libfm-qt.so -> libfm-qt.so.8
lrwxrwxrwx   1 root root    17 25 feb  2019 libfm-qt.so.6 -> libfm-qt.so.6.1.0
-rwxr-xr-x   1 root root  1,7M 22 feb  2019 libfm-qt.so.6.0.0
-rwxr-xr-x   1 root root   31M 22 apr  2020 libfm-qt.so.6.1.0
lrwxrwxrwx   1 root root    17 26 apr  2020 libfm-qt.so.7 -> libfm-qt.so.7.0.0
-rwxr-xr-x   1 root root   30M 24 ott 10.27 libfm-qt.so.7.0.0
lrwxrwxrwx   1 root root    17  4 nov 09.28 libfm-qt.so.8 -> libfm-qt.so.8.0.0
-rwxr-xr-x   1 root root   31M  5 dic 10.45 libfm-qt.so.8.0.0
lrwxrwxrwx   1 root root    11  1 dic 08.46 libfmt.so -> libfmt.so.7
lrwxrwxrwx   1 root root    15  1 dic 08.46 libfmt.so.7 -> libfmt.so.7.1.3
-rwxr-xr-x   1 root root  218K  1 dic 08.46 libfmt.so.7.1.3

```

removed libfm-qt.so.7* and libfm-qt.so.6* and it works

tsujan:
@stefonarch, thanks! I guess you'd directly installed from the source, without making a package with PKGBUILD.

stefonarch:
>  without making a package with PKGBUILD.
yeah, I should stop doing so...

tsujan:
> yeah, I should stop doing so...

In my old computer, that I use for testing, I once installed from the source and paid the price ;)

Another confusing mistake was unintentionally installing an app in `/usr/local/` (on my test computer). Later, it didn't want to be upgraded until I realized that I had two binaries of it.

PR #616, opened 2021-01-13
Move to new Qt5 signal/slot syntax.

adjamhub:
I also removed a QT_VERSION check pointing to Qt 5.11 version as
we have 5.12 as minimum version


tsujan (approved):
GTM

Thanks for removing the redundant workaround!

PR #619, opened 2021-01-21
Added the missing eject button in a special case

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/1295


stefonarch:
That was fast, works.
GTM

stefonarch (approved):

PR #621, opened 2021-01-29
Fix the wrong xarchiver command to create archives

ib:


tsujan:
I don't use xarchiver and I'm confused by its documentation. Please help me by explaining these:

 1. xarchiver 0.5.4.16 has the command-line option `--compress` in Arch but the documentation says otherwise: http://xarchiver.sourceforge.net/doc/ch03s02.html. I wonder whether `--compress` is Arch-specific or the doc is outdated.
 2. Moreover, the doc says, about `--add-to`, "If archive doesn't exist it is created"; while there's no mention of archive creation in `man xarchiver`. Is the archive created with `--compress` too?

ib:
I was asked to create this pull request. Please see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=932461#75 and https://github.com/lxde/libfm/pull/69.

http://xarchiver.sourceforge.net is outdated. It refers to version 0.5.4, more than six years old. The current version is 0.5.4.17.

Xarchiver switched from option `--add-to` to `--compress` 4 years ago with version 0.5.4.9.

tsujan:
@ib, thanks for the explanation!

PR #624, opened 2021-02-02
Launch multiple files associated with same app together

tsujan:
Previously, when multiple files were launched, they were grouped based on their mimetypes. That could cause a problem if different mimetypes were associated with the same app.

For example, if two video files with different mimetypes were launched, some video players might play them simultaneously, instead of adding the second file to their playlist; some others might even play the first file for a moment, stop playing it and start to play the second one after showing an error message.

This patch guarantees that all files associated with the same app are launched together and in the same order they are selected.

Closes https://github.com/lxqt/libfm-qt/issues/623

Also, a bad mistake (typo) is fixed in the same source file. Luckily, it had no consequence.


tsujan:
Tested for 3 weeks. Merging...

PR #627, opened 2021-02-03
Added support for creation time

tsujan:
GLib is responsible for reporting the correct creation time. Closes https://github.com/lxqt/libfm-qt/issues/397

WARNING: All libfm-qt based apps should be recompiled.

NOTE: This PR will be followed by another for pcmanfm-qt to complete the support for creation time.

### DIFF
--- a/src/core/fileinfo.cpp
+++ b/src/core/fileinfo.cpp
@@ -262,6 +262,7 @@ _file_is_symlink:
     mtime_ = g_file_info_get_attribute_uint64(inf.get(), G_FILE_ATTRIBUTE_TIME_MODIFIED);
     atime_ = g_file_info_get_attribute_uint64(inf.get(), G_FILE_ATTRIBUTE_TIME_ACCESS);
     ctime_ = g_file_info_get_attribute_uint64(inf.get(), G_FILE_ATTRIBUTE_TIME_CHANGED);
+    crtime_ = g_file_info_get_attribute_uint64(inf.get(), G_FILE_ATTRIBUTE_TIME_CREATED);
     if(auto dt = g_file_info_get_deletion_date(inf.get())){
         dtime_ = g_date_time_to_unix(dt);
     }
--- a/src/core/fileinfo.h
+++ b/src/core/fileinfo.h
@@ -93,6 +93,9 @@ public:
         return ctime_;
     }
 
+    quint64 crtime() const {
+        return crtime_;
+    }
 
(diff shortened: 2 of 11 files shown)


stefonarch:
While testing I stumpled upon this:
 
schermata-18-51-44


schermata-18-50-37

Looks like something gets inverted.

tsujan:
@stefonarch 
What's the problem? I can't see any in your screenshot.

stefonarch:
Creation time is after access time and modified time. And I see now that it's the same for all pictures: 06/06/2016  probably when I copied them on a this disk.

tsujan:
> Creation time is after access time and modified time.

That's understandable. The modification time can be long before the creation time because the file may have been modified elsewhere and then copied to your drive. The same is true for the access time, although the access time is less reliable because it can be disabled.

What GLib sees as the creation time is out of our control — the patch just shows and uses it — but I think it's correct.

stefonarch:
Thanks, I see. Just thought creation time is preserved by copying.

tsujan:
> Just thought creation time is preserved by copying.

Actually, at first, I thought so too but then found the logic by checking several files. Also, pages like this confirmed that: https://superuser.com/questions/1091134/can-a-file-appear-to-be-modified-before-it-was-created

stefonarch:
Just saw on my sd-card from the camera that in no-ext4 filesystem the creation time is shown as 1.1.1970
Would there be a way not to show it on filesystems  like fat32 ecc?

tsujan:
> Would there be a way not to show it on filesystems like fat32 ecc?

I first did so but removed it before making the patch. The reason: the creation time is like all other times; if some device doesn't support it, it's not our job to hide the problem.

tsujan:
I'm merging the PRs about Creation time. "1.1.1970" is a separate subject that may need to be discussed later.

stefonarch:
> "1.1.1970" is a separate subject that may need to be discussed later.

Just as idea: show it as "unknown" or "not readable" if it matches this date?

tsujan:
"Unknown" or any other string has this problem that when you sort files by the creation time, the files without that date may come after those having it. I don't know of such a scenario but prefer to be cautious.

In those cases, I'll try to show an empty string inside the window (pcmanfm-qt, file dialog,...) and not to show any info on the tooltip. But, for the sake of consistency, it should be done for all times, not just for the creation time.

PR #634, opened 2021-02-24
Fixed the size limit for thumbnail creation

tsujan:
It was forgotten in the code and had no effect.

Fixes https://github.com/lxqt/pcmanfm-qt/issues/1315


tsujan:
Merging (I want to fix pcmanfm-qt's spinbox for it).

PR #639, opened 2021-02-28
Don't show Unix start time for unsupported times

tsujan:
Closes https://github.com/lxqt/libfm-qt/issues/638


tsujan:
If there's no objection, I'd like to merge my PRs related to libfm-qt in a few days because I want to remove compilation warnings after that and before 0.17. I'm using them but please tell me if you see any problem in the codes or in practice!

stefonarch:
In practice it's working fine.

PR #653, opened 2021-03-13
updated *.ts files

stefonarch:


tsujan (approved):
GTM.

Like in https://github.com/lxqt/pcmanfm-qt/pull/1333, you don't need my approval because you know Weblate much better than I do.

PR #659, opened 2021-03-24
Fix (quite) all compilation warnings.

adjamhub:
To be fair, there is one missing... about glib g_slist_free_full.
But I had to give up ;)

### DIFF
--- a/src/filedialoghelper.cpp
+++ b/src/filedialoghelper.cpp
@@ -314,25 +314,31 @@ void FileDialogHelper::loadSettings() {
     dlg_->setSplitterPos(settings.value(QStringLiteral("SplitterPos"), 200).toInt());
     settings.endGroup();
 
-   settings.beginGroup (QStringLiteral("View"));
-   dlg_->setViewMode(viewModeFromString(settings.value(QStringLiteral("Mode"), QStringLiteral("Detailed")).toString()));
-   dlg_->sort(sortColumnFromString(settings.value(QStringLiteral("SortColumn")).toString()), sortOrderFromString(settings.value(QStringLiteral("SortOrder")).toString()));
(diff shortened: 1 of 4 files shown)


src/filedialoghelper.cpp
```diff
@@ -407,9 +413,17 @@ void FileDialogHelper::saveSettings() {
...
-        settings.setValue(QStringLiteral("HiddenPlaces"), sl);
+    else {
+#if (QT_VERSION >= QT_VERSION_CHECK(5,14,0))
+        QStringList hiddenPlacesList = settings.value(QStringLiteral("HiddenPlaces")).toStringList();
+        QSet<QString> hiddenPlacesSet = QSet<QString>(hiddenPlacesList.begin(), hiddenPlacesList.end());
+#else
+        QSet<QString> hiddenPlacesSet = settings.value(QStringLiteral("HiddenPlaces")).toStringList().toSet();
+#endif
+        if (hiddenPlaces != hiddenPlacesSet) {
+            QStringList sl = hiddenPlaces.values();
```
tsujan, line 424: `QSet::toList()` and `QSet::values()` are the same. If you want to use range constructors with Qt ≥ 5.14, you could write:

```
            QStringList sl (hiddenPlaces.begin(), hiddenPlaces.end());
```

src/folderview.h
```diff
@@ -130,7 +130,7 @@ class LIBFM_QT_API FolderView : public QWidget {
     void setCustomColumnWidths(const QList<int> &widths);
 
     QList<int> getHiddenColumns() const {
-        return hiddenColumns_.toList();
+        return hiddenColumns_.values();
```
tsujan, line 133: The same as above.

adjamhub: Ok. It's just that the compiler suggested me to use values :)
I'll ifdef it then

tsujan (approved):
Everything seems OK. Thanks!

PR #660, opened 2021-03-25
Added mount, unmount and eject actions to file context menu when possible

tsujan:
Particularly, the items inside `computer:///` can have those actions in their context menus.

Also, added an emblem to the icons of mounted files.

Closes https://github.com/lxqt/libfm-qt/issues/658

WARNING: All libfm-qt based apps — and especially pcmanfm-qt — should be recompiled.


tsujan:
Tested by me for a month. Merging...

PR #662, opened 2021-03-27
Added a missing nullity check in "mountoperation"

tsujan:


tsujan:
Straightforward. Merging...

PR #665, opened 2021-03-28
Fixed a memory leak in LXQt file dialog

tsujan:
`Fm::CachedFolderModel` wasn't unreffed. That could also cause a rare crash on logging out because of the connection to X11 clipboard.


src/filedialog.cpp
```diff
@@ -247,6 +247,12 @@ FileDialog::FileDialog(QWidget* parent, FilePath path) :
 
 FileDialog::~FileDialog() {
     freeFolder();
+    if(proxyModel_) {
```
palinek, line 250: delete nullptr is a noop by design. So there is no point to guard it by if...

tsujan: Correct. Copy-pasted from pcmanfm-qt → `tabpage.cpp`. There's no harm.

PR #672, opened 2021-04-15
Workaround for a bug is GFileMonitor regarding symlinks

tsujan:
If a single `GFileMonitor` is unreffed, those of open folders with the same (symlink) target will stop emitting signals and become useless. That's an old bug in GLib and this patch is a workaround for it by:

 1. Unreffing the old file monitor only after a new one is created when reloading a folder; and
 2. Finding and reloading all open folders that have the same target of the folder that is closed.

A new method is added to `Fm::FileInfo` to get the file identifier. By using a simple pointer comparison, it can be used to find files that have the same target.

Closes https://github.com/lxqt/pcmanfm-qt/issues/607

NOTE: All libfm-qt based apps should be recompiled.

### DIFF
--- a/src/core/fileinfo.cpp
+++ b/src/core/fileinfo.cpp
@@ -12,6 +12,7 @@ const char defaultGFileInfoQueryAttribs[] = "standard::*,"
                                             "access::*,"
                                             "trash::deletion-date,"
                                             "id::filesystem,"
+                                            "id::file,"
                                             "metadata::emblems,"
                                             METADATA_TRUST;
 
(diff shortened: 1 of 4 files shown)


tsujan:
This patch was made a long time ago and I really want to get rid of the nasty problem that it fixes. Although I've tested it many times, I prefer another LXQt member do it too.

@stefonarch, could you please test it? If yes, please first confirm the problem that's explained in https://github.com/lxqt/pcmanfm-qt/issues/607 and then see if the patch fixes it.

stefonarch:
Con confirm both issues.

Made some tests but do not have much time today, compiled this branch and pcmanfm-qt but it looks like 
that at point 3) the folder in the symlink DIR is not removed anymore when trashing originally folder, clicking it gives "invalid folder" (not translated?), restoring it it want's to open the symlink in ark and no way to change file association, `file` says `symbolic link to /home/stef/path/to/folder`, get's fixed when pressing F5.

Didn't have the time to test the second issues with renaming.
`

tsujan:
> Can confirm both issues.

Thanks!

> it looks like that at point 3) the folder in the symlink DIR is not removed anymore

That's not related to the problem in question. Yes, if the target of a symlink is removed, the symlink won't be updated until the parent folder is reloaded. It isn't a bug. We can't monitor targets, for several reasons.

The first issue is about trashing the file that's inside DIR, changing the directory from DIR to trash, restoring the file, and seeing the it doesn't *seem* to be restored in the symlink directory (which was open during the whole operation).

> Didn't have the time to test the second issues with renaming.

There's no rush. Do it whenever you find the time.

stefonarch:
I tested now the symlink-rename thing, it ends up with successfully renaming original file but the symlink then disappears.

tsujan:
Yes, if a file is overwritten by its symlink, it'll be lost.

You mean we should not allow that? I'm not sure what other file managers do (would you test?) but, IMO, it's a good idea to disallow it in `libfm-qt`. I'll investigate it and might open a report about it. Thanks!

PR #678, opened 2021-04-17
Distinguish between singular and plural translations of deletion prompt

tsujan:
It affects translations. In English, "file(s)" will be shown (unless an English translation file is added).

Closes https://github.com/lxqt/pcmanfm-qt/issues/1183


tsujan:
The sooner translators see it, the better. Merging...

stefonarch:
> The sooner translators see it, the better. Merging...

https://github.com/lxqt/libfm-qt/pull/681

PR #685, opened 2021-05-01
Enable Qt's alternating row colors for detailed list view

tsujan:
The colors depend on the widget style (we avoid hard-coded styles as far as possible).

Closes https://github.com/lxqt/pcmanfm-qt/issues/1368

Thanks to @ssokolow for the idea.


tsujan:
Any objection to this?

I think the file view of lxqt-archiver also needs alternate row colors; they're helpful in list-views.

PR #694, opened 2021-06-17
Workaround for launching apps that work with terminal

tsujan:
GLib has a hard-coded terminal list but no way of changing it in a code. Because of that, in the case of apps that only work through a terminal emulator, the user-defined terminal will be ignored if files are launched by `g_app_info_launch_uris()`.

PCMan had made a workaround a long time ago. It's still among libfm-qt's "legacy codes" and was used in https://github.com/lxqt/libfm-qt/commit/9af480812bdef4a7cca7db7416b24d96b179ebf1 for special cases. Here, it's used globally.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1399


tsujan:
If a problem exists — which is unlikely — it'll be found by git users. No point to waiting more.

PR #697, opened 2021-06-24
Fixed device usage of file properties dialog is rare cases

tsujan:
The device usage of a folder (like `/tmp`) may be different from that of its parent folder. In such cases, the file properties dialog showed different device usages on right clicking an empty space of the folder, depending on whether the parent folder was also opened (in another tab) or not.

This patch is rather a compromise between a total fix, which would require changes to the codes of `libfm-qt`-based apps, and the current situation. It practically fixes the issue without needing code changes to those apps.

WARNING: All `libfm-qt`-based apps should be recompiled; otherwise they might not run after applying this patch (and will give "symbol lookup error").

Fixes https://github.com/lxqt/libfm-qt/issues/696

### DIFF
--- a/src/filepropsdialog.cpp
+++ b/src/filepropsdialog.cpp
@@ -48,7 +48,7 @@ enum {
     ACCESS_FORBID
 };
 
-FilePropsDialog::FilePropsDialog(Fm::FileInfoList files, QWidget* parent, Qt::WindowFlags f):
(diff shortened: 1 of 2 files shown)


tsujan:
Found an elegant and complete solution that doesn't require recompilation of `libfm-qt`-based apps. Will make another PR...

PR #703, opened 2021-07-07
Fixed a memory leak in side-pane's volume items

tsujan:
Also applied the same logic to mount items.

Recompilation of `libfm-qt`-based apps is recommended.


tsujan:
Straightforward. Merging...

PR #704, opened 2021-07-08
Avoided a freeze on mounting encrypted volumes

tsujan:
For some reason unknown to me (GVFS or Qt bug?), having any local event loop in response to `GMountOperation`'s signal "ask-password" will cause a total app freeze if the mounting is done by `g_volume_mount()` *and* a volume is added/removed during the loop. Such a loop is needed for entering the password. The freeze also happens on auto-mounting a drive with both encrypted and non-encrypted partitions, for the same reason.

This patch works around the problem by *not* showing the password dialog in response to "ask-password" when `g_volume_mount()` is used. Instead, it first tries to mount the volume without a password. In case a password is needed, (a) it shows the password dialog only after the mount failure but without showing an error message, (b) saves the results of the user's actions and (c) mounts the volume again by using those results.

To the user, everything should seem like before but without the possibility of a freeze; he doesn't see the stages a, b and c.

NOTE: All `libfm-qt`-based apps should be recompiled.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1414

### DIFF
--- a/src/mountoperation.cpp
+++ b/src/mountoperation.cpp
@@ -24,7 +24,6 @@
 #include <QMessageBox>
 #include <QPushButton>
 #include <QEventLoop>
-#include "mountoperationpassworddialog_p.h"
 #include "mountoperationquestiondialog_p.h"
 #include "ui_mount-operation-password.h"
 #include "core/gioptrs.h"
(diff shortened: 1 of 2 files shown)


stefonarch:
Looks like this is also implemented in https://github.com/lxqt/libfm-qt/pull/707 (which I'ma using atm) as it doesn't freeze.
I found that unmounting by panel widget menu isn't possible now for encrypted volumes.

tsujan:
> Looks like this is also implemented in #707

No. You probably applied #707 over it, which is OK.

tsujan:
Let's get rid of this nasty freeze!

PR #706, opened 2021-07-15
Don't show impossible drop actions as far as possible

tsujan:
Closes https://github.com/lxqt/pcmanfm-qt/issues/1419


tsujan:
Tested by @stefonarch and me. I couldn't find a problem anywhere. Merging for git users...

PR #707, opened 2021-07-17
Support adding/removing emblems with File Properties dialog

tsujan:
When the emblem button is clicked, a file dialog is opened inside the active icon theme. The "emblems" folder will be prioritized if it's a top-level directory but any icon can be selected from the theme.

Also, a button is added for clearing emblems.

*WARNING:* `libfm-qt`-based apps should be recompiled.

Closes https://github.com/lxqt/libfm-qt/issues/700

### DIFF
--- a/src/core/fileinfo.cpp
+++ b/src/core/fileinfo.cpp
@@ -448,6 +448,43 @@ void FileInfo::setTrustable(bool trust) const {
                                     nullptr, nullptr);
 }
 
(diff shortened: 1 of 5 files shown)


stefonarch:
Works fine :)
schermata-07-17-10-34

tsujan:
Thanks for the test! I wanted to ask you to test it but forgot to add a comment.

Will merge it in a few days.

PR #708, opened 2021-07-21
updated *.ts files

stefonarch:


stefonarch:
Looks like also Turkish has similar issues as Hebrew:

```
Updating 'src/translations/libfm-qt_tr.ts'...
    Found 312 source text(s) (5 new and 307 already existing)
    Removed 1 obsolete entries
    Same-text heuristic provided 1 translation(s)
Removed plural forms as the target language has less forms.
If this sounds wrong, possibly the target language is not set or recognized.

```

tsujan:
> Looks like also Turkish has similar issues as Hebrew

Trust Qt. It has always been reliable in this regard.

tsujan:
BTW, a new translatable string is added by a recent merge (https://github.com/lxqt/libfm-qt/pull/706). So, another update is needed.

stefonarch:
> BTW, a new translatable string is added by a recent merge (#706). So, another update is needed.

Done, can you have a look?

tsujan:
It's good to merge.

tsujan (approved):

PR #710, opened 2021-07-24
Ensure corrrect selection order with Shift+mouse in icon view

tsujan:
Due to a problem in Qt, selection with Shift+mouse may sometimes have an incorrect order inside the icon view, resulting in an incorrect order of launching. This patch works around the problem by sorting the selection.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1421

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -102,13 +102,29 @@ void FolderViewListView::mousePressEvent(QMouseEvent* event) {
     // use the selection corner only with the extended and multiple selection modes
     // and change the mode to multiple temporarily if it is extended
     QAbstractItemView::SelectionMode sm = selectionMode();
-    if(sm == QAbstractItemView::ExtendedSelection
-       && cursorOnSelectionCorner_ && event->button() == Qt::LeftButton) {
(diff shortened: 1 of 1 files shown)


stefonarch:
Did not compile yet, but with .txt files  got a weird ordering always, different views, ctrl-a enter, ctrl+shift did not make difference.
schermata-13-50-22
schermata-13-50-01

stefonarch:
Compiling libfm-qt is enough? I can still reproduce different ordering in opening those files, depending on views and/or shift  or cltr+a

tsujan:
@stefonarch 
This is only about Shift-selection and yes, compiling `libfm-qt` is enough. But let me explain what you see: You open files in an app like FeatherPad. FeatherPad loads files in parallel, so that its tab order isn't predictable.

To test it, use Shift to select multiple video files with mouse inside pcmanfm-qt and right click them to open them in VLC or SMplayer. Then, take a look at the playlist: the order should be the same as in pcmanfm-qt. Video player enqueue files in the order they're given to them.

tsujan:
I've seen no problem in it; merging to afflict git users with it ;)

PR #721, opened 2021-09-03
Consider horizontal icon views, especially with smooth scrolling

tsujan:
They're possible, e.g., with lximage-qt.

The patch also takes horizontal angle data into account (mostly relevant with touchpads).

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -1652,14 +1652,16 @@ bool FolderView::eventFilter(QObject* watched, QEvent* event) {
                                                         : view->verticalScrollBar());
                 if(sbar != nullptr) {
                     QWheelEvent *we = static_cast<QWheelEvent*>(event);
-
+                    QPoint angleDelta = we->angleDelta();
+                    Qt::Orientation orient = (qAbs(angleDelta.x()) > qAbs(angleDelta.y()) ? Qt::Horizontal : Qt::Vertical);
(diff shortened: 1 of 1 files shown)


tsujan:
@stefonarch 
Could you do an unusual test if you have a touchpad? First apply this patch; then test the vertical thumbnail bar of https://github.com/lxqt/lximage-qt/pull/441 and see if you can scroll it with touchpad when the cursor is inside the thumbnail view?

stefonarch:
Done.
* to apply changes in preferences (show left|right...) pressing shortcut T twice is needed.
* With folders containing large number of files the scroll with touch pad  is unusable - too fast (same as in file manager with pictures, there is a issue open somewhere about those 3-lines).

Smooth scroll would be nice, as in Shotwell.

tsujan:
Is it better now, with icon sizes ≥ 96? WARNING: Saying "no" may have lethal consequences.

stefonarch:
It's much better but still needs carefull scroll, is this enough to avoid lethal consequences?

tsujan:
> is this enough to avoid lethal consequences?

Certainly ;)

You could also check pcmanfm-qt. I made the scrolling (with mouse and touchpad alike) row-by-row in the icon and thumbnail views when the icon size is greater than or equal to 96px. That's incompatible with the global settings but, after so many complaints, I thought it should be done.

Issue #722, opened 2021-09-05
Icon/thumbnail view's smooth scrolling may not work with some touchpads


xiota:
**Description:** Scrolling in the icon and thumbnail modes freezes up when the cursor is over the folder view area.  Scrolling works fine when the cursor is positioned over the scrollbar.  

I first noticed the problem in [lximage-qt 4cd4e57](https://github.com/lxqt/lximage-qt/commit/4cd4e574da25302b56af3724b1747fd85173a104) using libfm-qt 0.17.1.  The problem also affects libfm-qt 1c38ef3.  Both pcmanfm-qt and lximage-qt are affected.  A video to demonstrate the problem is attached.

This problem occurs in two different desktop environments (KDE Plasma, Xfce) on two computer I tested.  I used both a touchpad and a trackball for testing.  I could not find a normal mouse.

**Expected Behavior:** Scrolling should "work".  It should not freeze up.

**Possible Solution:**  When I added IconMode and ThumbnailMode to the if statement, scrolling works as expected.  This may be related to lxqt/pcmanfm-qt#85, since a potential fix is the same.

Potentially related to #611.

**Environment:** 
* OS: Kubuntu 21.04
* Kernel: 5.10.??
* Qt Version: 5.15.2
* libqtxdg Version: 3.7.1 and [dcac08a](https://github.com/lxqt/libqtxdg/commit/dcac08a03317e544152570b954df41352618e6f9)
* lxqt-build-tools Version: 0.9.0
* Package version: 0.17.1 and 1c38ef3

https://user-images.githubusercontent.com/8353098/132129929-bb886b0a-e4ef-4d98-871d-f2a1427d0ad8.mp4

tsujan:
Not reproducible.

Please add  System Information; it was in the template for a reason.

The report can be reopened if you give us a way of reproducing it.

xiota:
@tsujan I added the info.  I just tested a third computer, Kubuntu 18.04, with an ancient version of libfm-qt and pcmanfm-16 and it does *not* appear to be affected.

tsujan:
As I said, we need a way of reproducing it. I suspect you may be using bad builds —`libqtxdg` isn't the latest version, let alone the latest git, you may have mixed packages (old and git) or badly compiled packages, etc.

xiota:
I'll try building libqtxdg from git to see if it makes a difference.

xiota:
[...]

PR #723, opened 2021-09-05
Treat scrolling in IconMode and ThumbnailMode the same as CompactMode

xiota:
Potential fix for #722

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -1669,14 +1669,15 @@ bool FolderView::eventFilter(QObject* watched, QEvent* event) {
                     return true;
                 }
             }
-            // This is to fix #85: Scrolling doesn't work in compact view
+            // This is to fix pcmanfm-qt#85: Scrolling doesn't work in compact view
+            // This also fixes libfm-qt#722: Scrolling doesn't work in icon and thumbnail modes
             // Actually, I think it's the bug of Qt, not ours.
             // When in compact mode, only the horizontal scroll bar is used and the vertical one is hidden.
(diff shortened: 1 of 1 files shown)


tsujan (changes requested):
Here, you've actually removed all the code below those lines.

xiota:
@tsujan  I don't understand your comment.  This PR is based on 1c38ef3.  No code was removed.  Only a couple lines in the comment to refer to the issue number and the condition of the if statement are modified.

tsujan:
> I don't understand your comment.

You changed the condition so that its following block will never be reached, which is equivalent to its removal.

xiota:
Okay.  I don't know how many modes there are.  Is that section of code necessary on your computers?

tsujan:
> Is that section of code necessary on your computers?

The computer is irrelevant.

xiota:
@tsujan
> The computer is irrelevant.

Apparently not, since you stated in #722:

> Please add System Information; it was in the template for a reason.

tsujan:
> Apparently not, since you stated in #722:

It was about system info, as included in the template. We didn't ask about your computer model, how much RAM it has, its graphic card, etc.

xiota:
Okay.  Is it necessary on your computer *setup*?  It looks like smooth scrolling might have been a hack to get around Qt limitations.  Is it even necessary anymore?

tsujan:
If a problem is encountered, first its existence should be proven by providing a way of reproducing it everywhere and, second, it should be fixed. As far as I know, only Gnome devs throw out whole codes when they encounter a problem (like their old codes of desktop icons, systray,...).

PR #724, opened 2021-09-05
Fixed smooth scrolling with some touchpad devices

tsujan:
Some touchpad devices may trigger wheel events with too small angle deltas, resulting in jumpy scrolling. The patch fixes the issue by waiting until the total delta reaches an acceptable value.

Closes https://github.com/lxqt/libfm-qt/issues/722

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -1706,9 +1706,9 @@ bool FolderView::eventFilter(QObject* watched, QEvent* event) {
                     // NOTE: Some touchpad devices may trigger wheel events with angle deltas
                     // less than "scrollAnimFrames", resulting in jumpy movements. Therefore,
                     // we wait until the total delta value is enough.
-                    static int totalDelta = 0;
-                    totalDelta += delta;
-                    if(abs(totalDelta) < scrollAnimFrames) {
+                    static int _delta = 0;
+                    _delta += delta;
+                    if(abs(_delta) < scrollAnimFrames) {
                         return true;
                     }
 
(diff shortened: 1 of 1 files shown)


xiota:
I'll test this, but you [had said before](https://github.com/lxqt/lximage-qt/pull/448#issuecomment-912989608) that this change wouldn't have any effect on scrolling in libfm-qt.

tsujan:
@stefonarch 
If scrolling from inside icon/thumbnail view of pcmanfm-qt with touchpad is erratic for you, please test it with this patch whenever you found the time.

stefonarch:
I already did, but cannot see any difference here. Pcmanfm-qt needs to reompiled or not?

tsujan:
> Pcmanfm-qt needs to reompiled or not?

No, it doesn't.

Could you describe what you see?

stefonarch:
I did so, restartet pcmanfm-qt, testing again I can see the improvement in touchpad_smooth_scrolling, tested in /usr/lib woth both icon&thumbnail view:
* with master 2 finger scroll does start only at some point, needs more care
* in this branch scrolling is perfect, slowly or faster depending on the movements.
* in compact view I get a horizontal scroll, working fine with the same movement, probably always been like this but I hardly ever use this

tsujan:
> in this branch scrolling is perfect, slowly or faster depending on the movements.

This was what I needed to be sure about. Thanks a lot!

stefonarch:
I recompiled all components on my PC with arch,weel scrolling with largest thumbnails size is working perfectly now, before it was unusable because of this 3 lines jumping.
:1st_place_medal:

tsujan:
Yes, after seeing several complaints, I decided to not respect LXQt's general setting when the view had big icons. I might add a more fine-tuned algorithm later; for now, when the icon size is ≥ 96, the scrolling is done by one item, but only in icon and thumbnail views.

xiota:
@tsujan I just re-tested this.  It's but much better, but still a little glitchy.  It may have been the icon size ≥ 96 part.

tsujan:
> but still a little glitchy.

Yes, it may still be jumpy under special circumstances and with some touchpads. I'm working on it (it isn't about icon size).

tsujan:
Done in https://github.com/lxqt/libfm-qt/pull/730

PR #726, opened 2021-09-06
Discard embedded thumbnails that are too small.

xiota:
Discard embedded thumbnails that are too small, and prepare to support larger thumbnails.  On 4k displays, embedded thumbnails can be too small and are shown with a large white border.  This PR causes thumbnails that are smaller than requested to be regenerated.  It also enables use of the "x-large" thumbnail folder.

### DIFF
--- a/src/core/thumbnailjob.cpp
+++ b/src/core/thumbnailjob.cpp
@@ -75,8 +75,18 @@ QImage ThumbnailJob::loadForFile(const std::shared_ptr<const FileInfo> &file) {
         return QImage();
     }
 
-    QLatin1String subdir = size_ > 128 ? QLatin1String("large") : QLatin1String("normal");
-    thumbnailDir += subdir;
+    if(size_ > 512) {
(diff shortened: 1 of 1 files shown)


tsujan (changes requested):
Superseded by https://github.com/lxqt/libfm-qt/pull/727

The reasons (sorry, didn't have time to add comments under code lines):

 1. Sometimes a thumbnail cannot be generated with the requested size. Trying to regenerate it would waste CPU time.
 2. Thumbnails should be scaled only when they're larger than the requested size. We don't sacrifice sharpness for size.
 3. Exif should be used when its data exist (as in the original code); otherwise, thumbnails might have wrong orientations.
 4. 256 should be in the "large" folder (other apps put it there).
 5. Thumbnails replace icons and icons have specific sizes. We shouldn't fill the thumbnail directory with images that have arbitrary sizes (with non-integer scale factors). For the same reason, 512px is the largest size that we may want to support as "x-large".

PR #727, opened 2021-09-07
Support x-large thumbnails

tsujan:
They may be needed with scale factors > 1.

Supersedes https://github.com/lxqt/libfm-qt/pull/726

### DIFF
--- a/src/core/thumbnailjob.cpp
+++ b/src/core/thumbnailjob.cpp
@@ -75,7 +75,9 @@ QImage ThumbnailJob::loadForFile(const std::shared_ptr<const FileInfo> &file) {
         return QImage();
     }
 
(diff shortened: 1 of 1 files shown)


xiota:
Can you add support for the `xx-large` folder also?

tsujan:
`xx-large` would be an overkill. The maximum icon size that we have in `libfm-qt` based apps (and can be found in good file managers like Dolphin) is 256 and no one uses a scale factor of 3 (unless for testing). However, it could be added easily whenever a 1024×1024 image is seen like a thumbnail by most.

xiota:
Why not add it now?  It won't be used if no apps request it, but will be future ready.

tsujan:
Overkill

PR #728, opened 2021-09-07
handle floating file dialogs

antis81:
This provides a concept for a window manager agnostic integration for PCManFM-Qt.

:bulb: Note:
Window pointers can leak if not used in conjunction with [pcmanfm-qt#1440](https://github.com/lxqt/pcmanfm-qt/pull/1440)

@stefonarch Could you test this on xfwm4?

### DIFF
--- a/src/filedialog.cpp
+++ b/src/filedialog.cpp
@@ -21,9 +21,8 @@
 
 namespace Fm {
 
-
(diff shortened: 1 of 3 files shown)


antis81:
See comment on pcmanfm-qt#1440!

PR #729, opened 2021-09-07
Stop scrolling on view scrollbars during inline renaming

xiota:
This change prevents scrolling during inline renaming when the mouse cursor is over the scrollbars.

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -1130,6 +1130,8 @@ void FolderView::setViewMode(ViewMode _mode) {
     if(view) {
         // we have to install the event filter on the viewport instead of the view itself.
         view->viewport()->installEventFilter(this);
+        view->verticalScrollBar()->installEventFilter(this);    // needed to prevent scrolling during rename
+        view->horizontalScrollBar()->installEventFilter(this);
         // we want the QEvent::HoverMove event for single click + auto-selection support
         view->viewport()->setAttribute(Qt::WA_Hover, true);
         view->setContextMenuPolicy(Qt::NoContextMenu); // defer the context menu handling to parent widgets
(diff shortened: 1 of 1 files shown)


tsujan:
Elsewhere, I asked you to check Dolphin for a reason. While I dedicate time to reviewing your good PRs as fast as possible, I don't think it'll be too much if I expect a little trust from you, in order to avoid explanations like the following.

If the file name is too long, it'll be scrollable inside the renaming editor. The user may want to add something to its start/end. The most natural way is scrolling it with the wheel. Now, if the file is inside a scrollable view, as soon the scrolling reaches a limit inside the editor, it might scroll the view. This annoying behavior is prevented by the code.

Wheel scrolling on the scrollbar can't cause such an interference because the cursor can't be inside the editor at the same time. There's no reason to prevent the user from having it and taking a look up or down before renaming a file.

List and compact views show the same behavior just for the sake of consistency. It isn't needed with them.

xiota:
I opened this as a PR because you appear to treat PRs more favorably than issues.  With an issue, this would have been closed along with a snide remark.  At least this way, I got an explanation.

tsujan:
> you appear to treat PRs more favorably than issues.

No; as you see now ;)

But I really appreciate the works of a good contributor like you. Since your first PRs, I felt that your work was reliable. In one case, your PR was about a duplication but, seeing its elegance, I couldn't help merging it ;)

xiota:
More favorably doesn't mean that you'll accept them.  For instance, you actually wrote an explanation.  With an issue, you would have said (did say) go check out dolphin and closed it.

tsujan:
>  With an issue, you would have said...

I put more time into LXQt than the free time I have. I manage it by avoiding long-winded comments.

PR #730, opened 2021-09-08
Enhanced wheel scrolling inside view

tsujan:
Particularly with smooth scrolling as well as touchpad.

The scroll speed is fine-tuned and rapid touchpad scrolling is considered.

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -1628,7 +1628,7 @@ bool FolderView::eventFilter(QObject* watched, QEvent* event) {
                 setCursor(Qt::ArrowCursor);
             }
             break;
-        case QEvent::Wheel:
+        case QEvent::Wheel: {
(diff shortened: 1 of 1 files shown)


tsujan:
@stefonarch 
I'd be grateful if you could test it — especially with touchpad — whenever you have time for it. pcmanfm-qt needs to be restarted (not recompiled).

stefonarch:
Done, it works very well (tested all views and scrolling with max thumbnails and narrow window, just 2 pictures in a row) but I cannot say if there is any diff to master, at least with my touchpad (quite recent laptop).

tsujan:
Thanks a lot!

> but I cannot say if there is any diff to master

Your touchpad may be like mine. I have another touchpad that works reliably only with this patch.

There's a small difference with mouse too: with ≥ 256-px thumbnails, half the image is scrolled by a single wheel turn.

PR #731, opened 2021-09-11
Smooth scrolling in compact and detailed list modes

tsujan:
Qt's jumpy scrolls with mouse wheel from inside the view can get annoying in those modes. The patch extends the smooth scrolling to them. Moreover, the code is made a little simpler and tidier this way.


stefonarch:
Here I see quite a difference!  Very nice!

tsujan:
The small problem is fixed now.

PR #732, opened 2021-09-11
Remember hidden columns of LXQt file dialog

tsujan:
The user may not need all columns of the detailed list mode.

WARNING: If you don't recompile apps that are based on `libfm-qt`, you might encounter paranormal events.


tsujan:
It's very simple. Merging...

PR #733, opened 2021-09-21
Don't show "Select All" and "Invert Selection" if they're impossible

tsujan:
They're impossible with single-selection mode (as in lximage-qt thumbnails bar or some file dialogs) and also with no-selection mode.

Also fixed a problem with "Invert Selection" in detailed list mode (the selection flag was forgotten).

Closes https://github.com/lxqt/lximage-qt/issues/466

### DIFF
--- a/src/foldermenu.cpp
+++ b/src/foldermenu.cpp
@@ -71,15 +71,23 @@ FolderMenu::FolderMenu(FolderView* view, QWidget* parent):
         separator2_ = addSeparator();
     }
 
-    selectAllAction_ = new QAction(tr("Select &All"), this);
-    addAction(selectAllAction_);
-    connect(selectAllAction_, &QAction::triggered, this, &FolderMenu::onSelectAllActionTriggered);
+    auto selMode = view_->childView()->selectionMode();
+    if(selMode == QAbstractItemView::SingleSelection || selMode == QAbstractItemView::NoSelection) {
+        selectAllAction_ = nullptr;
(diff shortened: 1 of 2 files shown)


stefonarch:
Works fine il lximage-qt, but I wonder what "no selection mode" is.

tsujan:
> I wonder what "no selection mode" is

It's there just because Qt allows it; we don't use it anywhere.

Thanks for the test!

tsujan:
@stefonarch 
I wonder what's the use of "Invert Selection" in folder context menu. As soon as the user shows the folder context menu, all selections will be cleared. IMO, that's the correct behavior because, otherwise, the user might mistake the menu for that of selected items.

Dolphin's behavior is the same but Dolphin shows the folder context menu without clearing the selectiion when `Ctrl` is also pressed. I don't think it can be considered a standard.

Any thought?

stefonarch:
Didn't notice that, as I use the shortcut or the "view" menu item to invert selections (rarely).
Probably the best would be not to show it.

tsujan:
It just looks strange to me; otherwise, it doesn't do any harm. I thought I was missing something; your reply showed me that I wasn't.

Sadly, I can't find a good place in LXQt file dialog for adding something like the Edit menu of pcmanfm-qt (with "Select All" and "Invert Selection" in it). I don't want to add a tool button for it and the Options menu doesn't seem like a good place...

PR #749, opened 2021-11-11
Prepare libfm-qt for implementing "Recent Files"

tsujan:
These simple changes are needed for implementing "Recent Files" in a consistent way (→ https://github.com/lxqt/pcmanfm-qt/issues/1483).

WARNING: Recompile all libfm-qt based apps after this, especially pcmanfm-qt!

### DIFF
--- a/src/filelauncher.cpp
+++ b/src/filelauncher.cpp
@@ -41,13 +41,28 @@ FileLauncher::~FileLauncher() {
 bool FileLauncher::launchFiles(QWidget* parent, const FileInfoList &file_infos) {
     GObjectPtr<FmAppLaunchContext> context{fm_app_launch_context_new_for_widget(parent), false};
     bool ret = BasicFileLauncher::launchFiles(file_infos, G_APP_LAUNCH_CONTEXT(context.get()));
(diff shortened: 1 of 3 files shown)


stefonarch:
Hm, compiled it and pcman/lximage restarted desktop module but I get still this and empty files when click.
schermata-11-12-07-30

tsujan:
> Hm, compiled it together with pcman/lximage,restarted desktop module but I get still this and empty files

Sorry, I should have said two things more clearly:

 * This patch changes nothing by itself. It just provides the necessary tools for the implementation of "Recent Files" in pcmanfm-qt, *in a future PR*.
 * "Recent Files" won't be related to `recent:///`. I've attached a screenshot at https://github.com/lxqt/pcmanfm-qt/issues/1483#issuecomment-966525077.

`recent:///` is useless to us. I'll mention that in the comment of the PR that I'll make for pcmanfm-qt, after merging this PR.

tsujan:
I think there's no need to wait: it's only about two empty virtual methods as well as using a method instead of repeating its contents in `FileMenu::openFilesWithApp`.

PR #752, opened 2021-11-15
Corrected a typo that could have caused crash

tsujan:
@cepamoi found it. Fortunately, the function that contained it was never used.

Closes https://github.com/lxqt/libfm-qt/issues/751


tsujan:
Straightforward. Merging...

PR #758, opened 2021-11-23
Consider the special case of QApplication::wheelScrollLines() == 0

tsujan:
This is a very rare and unusual case but it's still possible. It didn't cause a crash on wheel scrolling because no wheel event was produced with it. However, a division by zero should be prevented *explicitly*.


tsujan:
My bad; division by zero was already prevented (by me).

PR #763, opened 2021-12-16
Fixed "Move Bookmark Down" in bookmark context menu

tsujan:
It had two old bugs: one in showing the item; the other in moving it.

Fixes https://github.com/lxqt/pcmanfm-qt/issues/1506

### DIFF
--- a/src/placesview.cpp
+++ b/src/placesview.cpp
@@ -357,9 +357,10 @@ void PlacesView::onMoveBookmarkDown() {
     PlacesModelBookmarkItem* item = static_cast<PlacesModelBookmarkItem*>(model_->itemFromIndex(action->index()));
 
     int row = item->row();
-    if(row < model_->rowCount()) {
+    QModelIndex indx = proxyModel_->mapFromSource(model_->bookmarksRoot->index());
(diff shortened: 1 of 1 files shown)


tsujan:
@stefonarch 
Please test it.

stefonarch:
Didn't find issues now, thanks!
No idea if it's possible with DND to show only the line between the bookmarks as it's impossible to nest them (one bookmark into the other).

tsujan:
> No idea if it's possible with DND to show only the line between the bookmarks

I think I know what you mean but the current behavior is intended. Dropping a bookmark exactly between other bookmarks isn't easy. Because of that, the drop height is extended to cover bookmarks too.

Suppose that X, Y and Z are three bookmarks with that order, from top to bottom:
```
X
Y
Z
```
Then, the small space between X and Y *and* the space taken by Y mean "Y's position". Similarly, the small space between Y and Z *and* the space taken by Z mean "Z's position". Therefore, if you want to put X after Y, you could drop it on Z; you won't need to drop it into the small space between Y and Z, which is difficult.

> Wondering how long this was around and nobody reported it...

I think it was there from start.

PR #764, opened 2021-12-17
Added some icons to bookmark context menu

tsujan:
Why being shy of adding appropriate icons?! They're more useful than texts.


stefonarch:
Nice :)

PR #766, opened 2021-12-28
Fixed file association with empty `XDG_CURRENT_DESKTOP`

tsujan:
When `XDG_CURRENT_DESKTOP` is empty, GLib's function `g_app_info_get_default_for_type()` reads the default app from `~/.config/mimeapps.list`. Therefore, we should also write the default app to that file in this special case.

Fixes https://github.com/lxqt/pcmanfm-qt/issues/1510


tsujan:
It's straightforward. Merging...

PR #777, opened 2022-02-03
filedialoghelper: Add protected API to internal dialog

palinek:
To allow descendants access to internal dialog. This is needed for e.g.
xdg-desktop-portal-lxqt for ability to inject additional options widget.

ref. lxqt/xdg-desktop-portal-lxqt#7

### DIFF
--- a/src/filedialoghelper.h
+++ b/src/filedialoghelper.h
@@ -36,6 +36,9 @@ public:
 
     bool isSupportedUrl(const QUrl &url) const override;
 
+protected:
+    inline Fm::FileDialog & dialog() { return *dlg_; }
+
 private:
     void applyOptions();
     void loadSettings();


CMakeLists.txt
```diff
@@ -10,7 +10,7 @@ set(LIBFM_QT_LIBRARY_NAME "fm-qt" CACHE STRING "fm-qt")
 
 set(LIBFM_QT_API_VERSION_MAJOR 1)
 set(LIBFM_QT_API_VERSION_MINOR 0)
-set(LIBFM_QT_API_VERSION_PATCH 0)
+set(LIBFM_QT_API_VERSION_PATCH 1)
```
tsujan, line 13: Please don't change this. Because of the changes that are made to libfm-qt and pcmanfm-qt, we can't make a point release. The versions will be set when LXQt 1.1.0 is released.

palinek: Are there ABI incompatible changes after 1.0.0 release?

palinek: OK. Then we can't do the proper release of xdg-desktop-portal-lxqt :(

stefonarch: What about the initial release now?

tsujan:
> Then we can't do the proper release of xdg-desktop-portal-lxqt :(

It'll be done with LXQt 1.1.0. There's no way to predict these things.

tsujan:
> What about the initial release now?

If it's really urgent, we might be able to break the rule and release it alongside libfm-qt, pcmanfm-qt, lximage-qt, lxqt-archiver and lxqt-qtplugin 1.0.1 (and maybe another component -- I'm not sure). Is it so urgent to justify that?

palinek:
> If it's really urgent, we might be able to break the rule and release it alongside libfm-qt, pcmanfm-qt, lximage-qt

It's the view of packaging.... if there would be only the library ( libfm-qt.so.X ) in the "binary" package and other files (archivers.list, terminals.list, libfm-qt-mimetypes.xml) split into different package(s), then you could have installed libfm-qt10 alongside with libfm-qt11 and there would be no need to release pcmanfm-qt, lximqge-qt.

I know the above approach from debian based distros, but arch and derivates has one-source-package == one-binary-package AFAIK

stefonarch:
I meant the 0.1.0 already released - can it be packaged e/o should we announce it (with comment about future release)?
Releasing all together takes time that can be better used I guess, and  two month isn't a long time.

palinek:
Yes... don't announce anything and we release xdg-desktop-portal-lxqt together with 1.1 and just then make the public announcement.

stefonarch:
Ok.

PR #781, opened 2022-02-11
Fix MIME types - cbz, cbr, rar

menaechmi:
The MIME types "application/x-cbr" and "application/x-cbz" have been deprecated for a while. They have been replaced with "application/vnd.comicbook-rar" and "application/vnd.comicbook+zip" respectively. See https://www.iana.org/assignments/media-types/application/vnd.comicbook-rar and https://www.iana.org/assignments/media-types/application/vnd.comicbook+zip.

Additionally, the MIME type "application/x-rar-compressed" was standardized to "application/vnd.rar". See https://www.iana.org/assignments/media-types/application/vnd.rar

So, I replaced the old versions with the corrected new versions, this should fix the right-click context menu for those items. As far as I could tell, those are the only ones that were incorrect, but there may be additional ones.


tsujan:
Thanks! GTM.

The review would take a lot of time if FeatherPad didn't have selection highlighting.

PR #792, opened 2022-03-03
IconEngine: Forward API to child icon engine

palinek:
..instead of doing own implementation.

### DIFF
--- a/src/core/iconinfo_p.h
+++ b/src/core/iconinfo_p.h
@@ -25,6 +25,7 @@
 #include "../libfmqtglobals.h"
 #include "iconinfo.h"
 #include <gio/gio.h>
+#include <private/qicon_p.h>
 
 namespace Fm {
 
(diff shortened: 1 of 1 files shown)


tsujan:
I'd experimented with almost exactly the same thing myself but I didn't make a PR for 2 reasons:

 1. Except for `xdndworkaround.cpp`, we never used a Qt private header anywhere. I really prefer to keep it that way; otherwise, it wouldn't be worth the probable future hassles.
 2. What follows was very confusing to me and I didn't have time to investigate: it worked with my icon set (https://github.com/tsujan/Plataro) but not with the Breeze set. My guess was that either Breeze had a problem or the problem wasn't *completely* in `Fm::IconEngine`.

Since the patches are practically identical, I'm sure I'd see the same things with this too.

tsujan:
Oh, right now, I investigated 2 :) Yes, the Breeze set only includes `actions/24@2x` in its `index.theme`, while it does have `places/24@2x` inside its directory tree. So, 2 is Breeze's problem and I was wrong at https://github.com/lxqt/libfm-qt/issues/788#issuecomment-1057797121 saying, "Breeze is better for testing".

1 is still valid and worries me. But yes, this works.

PR #798, opened 2022-04-12
Prerelease changes

tsujan:


CHANGELOG
```diff
@@ -1,3 +1,19 @@
+libfm-qt-1.1.0 / 2022-04-15
+============================
+ * Prepared libfm-qt for implementing "Recent Files".
+ * Silenced Valgrind about an uninitialized value.
+ * Corrected a typo that could have caused crash.
```
Chiitoo, line 5: Maybe 'caused **a** crash'?

tsujan: > Maybe 'caused a crash'?

It didn't cause any crash because it wasn't used anywhere.

Chiitoo: Sorry, I meant that part of the sentence looks a bit odd (to me at least) without the 'a' in there.

tsujan: Oh, sorry for my misunderstanding!

I think articles can be dropped sometimes. There's no exact rule, but here "crash" is meant very generally — although I omitted the "a" without thinking. (see https://english.stackexchange.com/questions/38759/whats-the-general-rule-for-dropping-articles-in-article-section-titles-or-in → back to disk).

PR #802, opened 2022-04-26
Allow disabling smooth scrolling entirely

galaxyhaxz:
After this change, here is a test scrolling from top to bottom in `/usr/bin` (directory with lots of files) while in Icon View mode:

With smooth scrolling:
- ~3.70 seconds to scroll all the way down
- ~4.49% CPU usage (reported by QPS)

Without smooth scrolling:
- ~3.20 seconds to scroll all the way down
- ~2.83% CPU usage (reported by QPS)

It feels much more responsive overall and it's less obnoxious without the animation.

String update: https://github.com/lxqt/pcmanfm-qt/pull/1573

### DIFF
--- a/src/filedialog.cpp
+++ b/src/filedialog.cpp
@@ -232,8 +232,6 @@ FileDialog::FileDialog(QWidget* parent, FilePath path) :
         showHiddenAction->setChecked(proxyModel_->showHidden());
         thumbnailsAction->setChecked(proxyModel_->showThumbnails());
         tooltipsAction->setChecked(!noItemTooltip_);
-        perPixelAction->setVisible(viewMode_ == FolderView::DetailedListMode
-                                   || viewMode_ == FolderView::CompactMode);
         perPixelAction->setChecked(scrollPerPixel_);
     });
 
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -1690,7 +1690,7 @@ bool FolderView::eventFilter(QObject* watched, QEvent* event) {
             bool horizontal(qAbs(angleDelta.x()) > qAbs(angleDelta.y()));
             if(event->spontaneous()
                && we->source() == Qt::MouseEventNotSynthesized
-               && (scrollPerPixel_ || (mode != DetailedListMode && mode != CompactMode))
+               && scrollPerPixel_
                // To have a simpler code, we enable horizontal smooth scrolling with mouse wheel
                // only in horizontal list views.
                && (horizontalListView || !horizontal)) {


tsujan:
I'm afraid I'd explained the following (and more) in other places.

When the mouse wheel is used, the user wants to see files while scrolling. In the icon/thumbnail view, a jumpy movement would interfere with that; hence the animation.

The option for disabling smooth scrolling in list and compact views was added only because of a bug in Qt, which could cause a delay in loading a view with a huge number of items.

Also, the patch doesn't care about the rest of the code.

galaxyhaxz:
> Also, the patch doesn't care about the rest of the code.

It works perfectly fine, I tested it extensively.

> When the mouse wheel is used, the user wants to see files while scrolling.

You can't read them while they are scrolling, because the velocity is too fast. It would have to scroll really slow.

I knew you'd reject this either way, but i put it up so anyone in the future can refer back to it and compile their own version. Such a shame you are so closed minded about a simple change that won't even bother you since you clearly like it and wouldn't turn it off to begin with. This is precisely why the linux community is so divided and as a result has far lower quality software than windows. Everyone does their own thing.

galaxyhaxz:
>  the user wants to see files while scrolling.

LXQT is the only DE that does this. I've tried Gnome, Mate, KDE, LXDE, TDE, and XFCE. Guess what?

none of them have it. so clearly the user doesn't want it....

tsujan:
> It works perfectly fine

Not enough.

> Such a shame you are so closed minded about a simple change

We don't add options just for the sake of adding options. In this regard, I'll remain "close minded" indefinitely.

> since you clearly like it and wouldn't turn it off

It isn't about my likes and dislikes, but you're free to try mind reading.

The goal isn't winning a race by wheel scrolling but seeing the files while scrolling. Even if the above-mentioned Qt bug didn't exist, disabling smooth scrolling in the compact and list modes would make sense (although useless) because of the small row height. It doesn't make sense in the icon and thumbnail views.

galaxyhaxz:
> The goal isn't winning a race by wheel scrolling but seeing the files while scrolling

This is where we differ. I want everything to be fast and instantaneous, and LXDE was by far the best at achieving that. LXQT has thus far exceeded by expectations in being a solid, snappy DE that lets you turn off all the crap you see in Gnome. If you move the mouse wheel down, then back up, it has to finish animating before it will then scroll up which is annoying. Without smooth scrolling it's instant. Do you not remember? You were the one who added it in the first place back in what 2017 based on commit history?

tsujan:
> You were the one who added it...

I explained why it was added.

> Here's an idea, open an issue and let everyone take a vote...

Open-source ≠ "Voted-source".

galaxyhaxz:
> I explained why it was added.

Not the option to disable smooth scrolling, the feature entirely. It was added into the QT version a few years ago.

> Open-source ≠ "Voted-source".

I'm well aware of that. How do you expect people to want to contribute to your projects though if you can't reach any sort of agreement with them? Before you say, just fork it and do your own thing. That's not what we need, yet another hundreth linux distro. I WANT to help out this project, I feel it is the best DE around. But smooth scrolling is super annoying and I can't stand it and having to clone/update/recompile every time a new release comes out is ridiculous.

These projects fall on the backs of hobbyists who often struggle finding time and enough people to keep them alive indefinitely. Why make it any harder? I have C/C++ skills...

tsujan:
> I WANT to help out this project

Contributions are always welcome and very appreciated. So many good changes wouldn't be possible without contributors' good works.

However, there's a difference between contributing and demanding.

galaxyhaxz:
> However, there's a difference between contributing and demanding.

And what do you lose by accepting a PR like this? Can you elaborate? Does my change open up a bug somewhere? When QT fixes the bug, do you still plan to keep the option?

tsujan:
Sorry for the delay; I was busy.

> And what do you lose by accepting a PR like this? 

 1. None of your arguments went beyond personal preferences or general ideas. Time recording or talking about less than %2 of CPU usage is hardly practical.
 2. Also, as I said, the patch ignores other parts of the code and damages them. I'd discuss them in a review if the change was justifiable.

My several replies on this page — this one included — can be interpreted as my appreciation of your interest in LXQt. I'm afraid I don't have time for more.

galaxyhaxz:
Fair enough. Looking at other closed PRs/issues in the past has shown you have a very tight view of what is practical vs "taste" and it always ends up with pointless arguments like this one. Changing icons is taste, changing theme is taste; I get not wanting to have 100 different options for every possible thing, but in this case it just requires updating what's already there. The file and web browser are the two programs people deal with more than anything else, and having them work the way you want improves overall usability.

What about something like time/date format? AFAIK you can't change it from MMDDYYYY which could confuse people living in countries with different layouts. Thunar let's you change it.

tsujan:
>  AFAIK you can't change it from MMDDYYYY

It's locale dependent.

> Thunar lets you change it.

LXQt may implement practical features of other DEs. Having app-specific options for date format can't be among them.

> It would also be nice to change how many items scroll per wheel tick

→ https://github.com/lxqt/pcmanfm-qt/wiki#mouse-wheel-scrolling

galaxyhaxz:
> Having app-specific options for date format can't be among them.

I agree, everything should follow the system locale. Just couldn't get it to display YYYYMMDD myself though I probably didn't change locale correctly. Using Lubuntu so i'll go back and give it a try.

>  https://github.com/lxqt/pcmanfm-qt/wiki#mouse-wheel-scrolling

With this and my patch things are working exactly the way I like it now! Thanks!

PR #811, opened 2022-06-23
Silenced compilation warnings about deprecated GLib functions

tsujan:
`g_pattern_match_string()` has been deprecated since GLib 2.70.


luis-pereira (approved):
GTM

PR #816, opened 2022-07-17
Don't show a deletion prompt without a selection

tsujan:
I guess I was the first user who pressed `Shift+Delete` without a selection and was surprised by a prompt.


tsujan:
Straightforward; merging...

PR #817, opened 2022-07-18
Allow remembering the result of execution prompt for multiple files

tsujan:
When multiple executable files are going to be launched, the execution prompt dialog gets a checkbox for remembering the user's action for next files of the same kind.

If you've selected 10 desktop entries inside `/usr/share/applications` and pressed `Enter` by mistake, you know why this is necessary.

WARNING: `pcmanfm-qt` and other apps that depend on `libfm-qt` should be recompiled after applying this patch.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1616

### DIFF
--- a/src/exec-file.ui
+++ b/src/exec-file.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>487</width>
-    <height>58</height>
+    <height>87</height>
    </rect>
   </property>
   <property name="windowTitle">
@@ -86,6 +86,17 @@
        </property>
        <property name="icon">
         <iconset theme="dialog-cancel"/>
+       </property>
+      </widget>
+     </item>
+    </layout>
+   </item>
+   <item>
+    <layout class="QHBoxLayout" name="remLayout">
+     <item>
+      <widget class="QCheckBox" name="remBox">
+       <property name="text">
+        <string>Remember for next files of this kind</string>
        </property>
       </widget>
      </item>
--- a/src/execfiledialog.cpp
+++ b/src/execfiledialog.cpp
@@ -54,10 +54,20 @@ ExecFileDialog::ExecFileDialog(const FileInfo &fileInfo, QWidget* parent, Qt::Wi
         ui->open->hide();
     }
     ui->msg->setText(msg);
+    ui->remBox->hide();
 }
 
 ExecFileDialog::~ExecFileDialog() {
     delete ui;
+}
+
+void ExecFileDialog::allowRemembering() {
+    ui->remLayout->setContentsMargins(0, 10, 0, 0);
+    ui->remBox->show();
+}
+
+bool ExecFileDialog::isRemembered() {
+    return ui->remBox->isChecked();
 }
 
 void ExecFileDialog::accept() {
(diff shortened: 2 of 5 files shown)


tsujan:
@stefonarch, I requested your review. I've tested it many times, but it'll be better if you test it too. Don't forget to recompile pcmanfm-qt after it.

stefonarch (approved):
Works as expected. I wonder if `Remember for next files of this type` would be a clearer text.

> Don't forget to recompile pcmanfm-qt after it.

I tested without  but saw the checkbox. Is it really needed?

tsujan:
Thanks!

> I tested without but saw the checkbox. Is it really needed?

Since the added methods are private, it shouldn't be needed. But better safe than sorry ;)

> I wonder if `Remember for next files of this type` would be a clearer text.

I we remove "next", the user might think that it's also for all future operations (which would be illogical, especially if Cancel is chosen). If "of this type" is removed, he might think that it's for all executable files in this operation, while it's only for this specific type, namely, desktop file, script or just executable (= one of 3). The user may have selected a collection that contains all 3 types; if so, he'll see 3 prompts, regardless of the number of files.

stefonarch:
It was only about changing `kind` in `type`.

tsujan:
> It was only about changing kind in type not about remove anything.

Oh, sorry! Wrestling with another code, and having thought about that label several times, I not only misread your comment but also used "type" in my previous comment by mistake.

I intentionally used "kind" (it could be "sort" too) because "type" usually means "mime type". An executable python script and an executable bash script are treated in the same way here, but they have different types.

I don't know if this choice of words can be exactly reflected by translations, but I guess that's a general problem.

stefonarch:
`Remember for all selected files` ?

tsujan:
> Remember for all selected files ?

It isn't for all of them. For example, there may be only 2 executable scripts among the selected files. Or the elected files may consist of 2 scripts, 3 desktop file and 4 executables: three prompts will be shown, each for one kind.

tsujan:
BTW, the implementation won't care if only one executable file exists among 10 selected files. Finding the number of executables of each kind would be a waste of CPU time, IMO.

stefonarch:
I see, I never open a bunch of different files so didn't even tested this.

tsujan:
> I never open a bunch of different files so didn't even tested this.

Me neither, but it's a valid scenario. `libfm-qt` gives different sets of actions to those 3 kinds; so, the check-box should appear for each.

stefonarch:
If I get it right:
* there is a limit of the number of files to open which triggers a confirmation
* this new checkbox appears only if there is at least one executable among them and the setting "ask before execute files" is checked

tsujan:
The check-box appears when

 1. There are multiple files; and
 2.  The first executable script, desktop entry or executable file is processed.

It may appear three times at most, each for every kind.

tsujan:
> there is a limit of the number of files to open which triggers a confirmation

Yes, a question message will always appear if the user has selected more than 20 files. The reason is that, opening 300 ordinary text files in a text editor by mistake can take a lot of CPU time. This isn't related to "ask before execute files".

> this new checkbox appears only if there is at least one executable among them and the setting "ask before execute files" is checked

Yes.

PR #819, opened 2022-07-18
Launch trashed files with `%u` or `%U` too

tsujan:
Previously, trashed files could be opened by non-GLib apps only when they had `%f` or `%F`in the exec keys of their desktop entries.

Remote trashed files may be possible, but I think this change is good for consistency. Dolphin can open trashed files with `%u` or `%U` too.

NOTE: Unfortunately, GVFS doesn't give info on sub-directories of trash. So, only files that are immediately under `trash:///` can be opened.

Closes https://github.com/lxqt/libfm-qt/issues/818


tsujan:
@stefonarch, please take a look at this.

stefonarch (approved):
Works as expected here too.

PR #822, opened 2022-07-28
Fixed error handling in creating symlinks by DND

tsujan:
Previously, an infinite loop was created if a symlink was going to overwrite a non-empty directory.

The error messages are those of GLib.

Fixes https://github.com/lxqt/pcmanfm-qt/issues/1618


stefonarch (approved):
Works fine.

PR #829, opened 2022-08-14
Fixed file deselection in LXQt file dialog

tsujan:
Previously, if a single file was selected and then deselected, the file name entry wouldn't be emptied and the deselected file would be opened on pressing `Open`. There were other scenarios too.

Fixes https://github.com/lxqt/libfm-qt/issues/828

NOTE: I'll rebase the Qt6 branch soon.


tsujan:
Straightforward. Merging....

PR #830, opened 2022-08-14
Rubber-band selection in list mode from non-name columns

tsujan:
Opening items by clicking a non-name column is kept intact in the single-click mode.

Also, fixed a bug in the rubber-band selection with RTL layouts.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1633

NOTE: I'll rebase the Qt6 branch soon.


stefonarch:
> and, I suppose, @stefonarch

Yes, didn't see issues.

PR #833, opened 2022-08-16
No activation in list mode if selection rectangle exceeds drag distance

tsujan:
This is another improvement to UX in the detailed list mode.

If the user drags the cursor on a non-name column and releases the left mouse button on the same column, activation will occur only with a small enough selection rectangle (whose "Manhattan length" doesn't exceed the dragging distance).


tsujan:
Already tested by by @akinokonomi and me. Merging...

PR #835, opened 2022-08-29
Restored deselection behavior of LXQt file dialog

tsujan:
It's the behavior of the default Qt file dialog (and perhaps other file dialogs).

Contrary to my arguments at https://github.com/lxqt/libfm-qt/issues/828, there are several reasons as to why it's the best behavior (although, perhaps, not intuitive at first glance).

This just reverts https://github.com/lxqt/libfm-qt/commit/0719f8d9b07a071017b7814b10d3719fe1f2cc68

Sorry for the inconvenience!


tsujan:
Because it's a reverting commit, I'm merging it immediately.

PR #837, opened 2022-09-17
Corrected the position of drop menu under Wayland

tsujan:
During DND under Wayland, `QCursor::pos()` isn't reliable for getting the cursor's "global" position (which is really not global but calculated relative to the parent widget). Instead, `QDropEvent::pos()` should be mapped by using `QWidget::mapToGlobal()`.

Although this is logical under X11 too, it makes no difference there.


stefonarch (approved):
GTM, works as expected.

tsujan:
Reminder: The Qt6 branch should be updated because `QDropEvent::pos()` is deprecated in Qt6.

PR #839, opened 2022-09-18
Fixed crash with DND into side-pane under Wayland

tsujan:
The patch is safer with X11 too because it gets the required info before showing the DND menu.

Fixes https://github.com/lxqt/libfm-qt/issues/838

NOTE: The misplacement of the DND menu can't be fixed in this case because the relevant classes don't include the required info. Unless Qt finds solutions/workarounds for Wayland nonsenses, Wayland users should tolerate them.


tsujan:
Straightforward. Merging...

PR #852, opened 2022-11-24
Specified the parents of context menus for use on Wayland

tsujan:
On Wayland, a parentless context menu might have a window decoration and pop up in an incorrect position if the window is inactive when right clicked. The patch prevents the problem by specifying the parents of folder and side-pane context menus.

There will be no change on X11.

### DIFF
--- a/src/placesview.cpp
+++ b/src/placesview.cpp
@@ -471,7 +471,7 @@ void PlacesView::contextMenuEvent(QContextMenuEvent* event) {
         // Do not take the ownership of the menu since
         // it will be deleted with deleteLater() upon hidden.
         // This is possibly related to #145 - https://github.com/lxqt/pcmanfm-qt/issues/145
-        QMenu* menu = new QMenu();
+        QMenu* menu = new QMenu(this);
         QAction* action = nullptr;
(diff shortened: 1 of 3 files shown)


src/placesview.cpp
```diff
@@ -471,7 +471,7 @@ void PlacesView::contextMenuEvent(QContextMenuEvent* event) {
         // Do not take the ownership of the menu since
         // it will be deleted with deleteLater() upon hidden.
         // This is possibly related to #145 - https://github.com/lxqt/pcmanfm-qt/issues/145
-        QMenu* menu = new QMenu();
```
palinek, line 474: Just minor nitpick. Did you consider the above comment? Or if it is irrelevant as of today, then it probably needs to be removed to not confuse.

tsujan: Thank you so much for bringing this to my attention! I'll check that old problem soon.

tsujan: To be on the safe side, I just changed that comment.

The code is correct and has worked reliably. Because, in another place, I had the experience of avoiding a weird hang that was related to `GLib`, I prefer not to change the rest of this code.

PR #855, opened 2022-12-11
appchooser: don't disable the OK button

kreijack:
When we are not in the app view tab don't change the enable state of the OK button, only because we lost the focus from the app on the app view.

In the appchooser dialog, sometime we loose the active status of the OK button in the "custom command" tab.
This happens because if in the "Installed Application" tab the selection is lost, then the "OK button enable status" is disabled.

This patch ensures that the OK button state is changed only when the "Installed Application" tab is shown.

### DIFF
--- a/src/appchooserdialog.cpp
+++ b/src/appchooserdialog.cpp
@@ -253,6 +253,11 @@ void AppChooserDialog::accept() {
 }
 
 void AppChooserDialog::onSelectionChanged() {
+    if(ui->tabWidget->currentIndex() != 0) {
+        // the selection may be reset by menu-cache,
+        // while the app menu view is not shown
+        return;
+    }
     bool isAppSelected = ui->appMenuView->isAppSelected();
     ui->buttonBox->button(QDialogButtonBox::Ok)->setEnabled(isAppSelected);
 }


src/appchooserdialog.cpp
```diff
@@ -253,6 +253,9 @@ void AppChooserDialog::accept() {
 }
 
 void AppChooserDialog::onSelectionChanged() {
+    // if we are not in app menu view then don't change the state of OK button
+    if(ui->tabWidget->currentIndex() != 0)
+        return;
```
tsujan, line 258: Thanks for finding and fixing the problem!

The root cause of the problem is `AppMenuView::onMenuCacheReload`: menu-cache may be reloaded while the tab has been switched.


To follow the existing code style, please usse curly brackets and put the comment inside it. Please also change the code comment to something like this:

```c++
    if(ui->tabWidget->currentIndex() != 0) {
        // the election may be reset by menu-cache,
        // while the app menu view is not shown
        return;
    }

```

kreijack:
Hi I updated the comment as per request.

tsujan (approved):
Thanks!

PR #857, opened 2022-12-12
Increased the minimum width of name column in list view

tsujan:
Previously, it was 200px, which might not be enough with wide fonts. This patch changes it to 50 times the space width plus icon width. It's usually more than 200px even with a small font size of 9.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1696


tsujan:
Merging because it's straightforward. I think an exact calculation (by considering text-icon spacing, text margin,...) would clutter the code for no good reason.

PR #859, opened 2022-12-13
Fixed DND menu position with DND between windows under Wayland

tsujan:
Made the DND menu a child of the viewport; otherwise, the DND menu position will be incorrect under Wayland when files are dragged from a window and dropped into another.

NOTE: To be on the safe side, recompile pcmanfm-qt and apps that are based on libfm-qt.

### DIFF
--- a/src/dndactionmenu.cpp
+++ b/src/dndactionmenu.cpp
@@ -46,9 +46,9 @@ DndActionMenu::~DndActionMenu() {
 
 }
 
-Qt::DropAction DndActionMenu::askUser(Qt::DropActions possibleActions, QPoint pos) {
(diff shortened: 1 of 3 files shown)


stefonarch:
Works fine under wayland - one thing less in the wrong position ;)

tsujan:
Thanks for testing it!

The only remaining issue is a crash under Wayland, that may happen when files are dragged from another app (like lxqt-archiver) and dropped into pcmanfm-qt. It's an old problem Qt has had with Wayland, and I haven't found a workaround for it... I'm getting tired of basic problems Wayland has created in the name of "security".

PR #861, opened 2022-12-14
Prevent DND crash from outside app under Wayland

tsujan:
By accepting the proposed action and not showing a DND menu.

Showing a DND menu and waiting for the user's response makes Wayland free the mime data and create a dangling pointer, which will cause a crash on accepting the DND.

Closes https://github.com/lxqt/libfm-qt/issues/860


tsujan:
Found a better solution, without sacrificing anything. Closing this PR....

PR #862, opened 2022-12-14
Fixed crash with DND from outside app under Wayland

tsujan:
A drag and drop from outside the app resulted in a crash under Wayland because Wayland freed the DND mime data once the DND menu was shown, leaving a dangling pointer, which caused the crash on accepting the DND.

This patch fixes the issue by processing the DND mime data only before showing the DND menu. X11 sees no change.

NOTE: This is rather a workaround for a Qt problem. Qt's DND needs fixes for Wayland, so that such a situation becomes impossible.

Fixes https://github.com/lxqt/libfm-qt/issues/860


stefonarch:
Confirming - crash is avoided now.

PR #864, opened 2022-12-15
Fixed unresponsive DND menu under Wayland

tsujan:
Under Wayland, not only the DND menu was unresponsive to mouseover, it might also cause a crash if Escape was pressed (twice).

The reason was that the menu was shown synchronously, before the DND was finished. This patch fixes the issue by showing the menu after finishing the DND (which is OK under X11 too).

Also, corrected the position of the side-pane DND menu under Wayland.

NOTE: Avoiding `QMenu::exec()` and using `QMenu::popup()` with signals and slots would be a cleaner solution. However, the current solution is much simpler and works fine. At some point, I might change the code to avoid `QMenu::exec()` as far as possible.

Fixes https://github.com/lxqt/libfm-qt/issues/863

### DIFF
--- a/src/dirtreemodel.cpp
+++ b/src/dirtreemodel.cpp
@@ -19,7 +19,6 @@
 
 #include "dirtreemodel.h"
 #include "dirtreemodelitem.h"
-#include "dndactionmenu.h"
 #include "fileoperation.h"
 #include "utilities.h"
 #include <QDebug>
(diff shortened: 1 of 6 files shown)


tsujan:
@stefonarch, could you confirm that the patch does its job? I think its the last big change needed for a good Wayland experience.

stefonarch:
I can't notice any difference between master and this, DND from archiver to pcmanfm-qt, cannot produce any crash.
Can you please give a detailed test instruction?

tsujan:
1. Drag an item and drop it inside the same folder. The DND menu appears. Previously, if you moved the cursor over the menu, no menu item would be selected; it was unresponsive to mouseover. Also, pressing Escape a few times might cause a crash.
 2. Drag an item from the main view and drop it into an item in the side-pane, whether in the Liists mode or in the Directory Tree mode. Previously, the DND menu was positioned incorrectly.

tsujan:
If you didn't see 1, that might be because you didn't have server-side decorations. But you should see 2 everywhere.

stefonarch:
Ok, I don't use server side decoration but can reproduce 1) with crash and 2) wrong position now. Those issues are fixed now here I see.

tsujan:
A good thing with Wayland is that it may show the weak points of the code in general.

PR #866, opened 2022-12-22
Workaround for DND keyboard modifiers under Wayland

tsujan:
To copy/move/link files directly by DND, `Ctrl` and/or `Shift` modifiers should be pressed only *after* dragging and before dropping because, if they are pressed before dragging, they might mess with the item selection.

However, Wayland doesn't see a keyboard modifier if it's pressed after dragging, rendering DND modifiers impractical. This patch is a simple workaround.

### DIFF
--- a/src/folderview.cpp
+++ b/src/folderview.cpp
@@ -1635,8 +1635,25 @@ void FolderView::childDropEvent(QDropEvent* e) {
             }
             auto curPos = view->viewport()->mapToGlobal(e->pos());
             QTimer::singleShot(0, view, [this, curPos, actions, srcPaths, destPath] {
-                // a parent is needed under Wayland for correct positioning
-                Qt::DropAction action = DndActionMenu::askUser(actions, curPos, view);
+                Qt::DropAction action;
+                // Wayland does not see the modifier if it is pressed after dragging.
+                // Therefore, it should be checked now, when the DND is finished.
+                switch(QApplication::keyboardModifiers()) {
+                case Qt::ControlModifier:
+                    action = Qt::CopyAction;
+                    break;
+                case Qt::ShiftModifier:
+                    action = Qt::MoveAction;
+                    break;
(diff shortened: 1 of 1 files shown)


stefonarch:
Working fine.

tsujan:
So fast! Thanks!

tsujan:
I feel that we might need to release libfm-qt/pcmanfm-qt 1.2.1 before starting an official Qt6 port -- because of Wayland fixes. If so, I'm not sure that the other components that depend on libfm-qt should also have point releases or their current releases should be recompiled by distros.

@yan12125, any suggestion?

yan12125:
> their current releases should be recompiled by distros.

This sounds more natural. To me it's kind of strange to have a new release just for a rebuild. On the other hand, distro packagers are familiar with rebuilding packags against new dependencies. A clear statement for packagers in release notes should be enough.

tsujan:
> A clear statement for packagers in release notes should be enough.

I'll do it. Thanks for the reply!

tsujan:
@yan12125 
If I'm not mistaken, the ABI hasn't changed. The only relevant change is here:

https://github.com/lxqt/libfm-qt/blob/2fd35a5e5495993abfa316cd9ef345c687ecfe88/src/dndactionmenu.h#L36

The argument `QWidget* parent` has been added, which is `nullptr` by default. So, there will be no need to recompilation of `libfm-qt` based apps; only `pcmanfm-qt` will be released because it completes the Wayland changes made to `libfm-qt`.

yan12125:
Generally speaking, changing number or type of C++ function arguments changes the ABI. In this case, the class `DndActionMenu` is not part of public ABI (not declared with `LIBFM_QT_API`), so ABI is not changed.

> So, there will be no need to recompilation of `libfm-qt` based apps

The conclusion is correct :)

tsujan:
Thanks!

PR #867, opened 2022-12-23
Check mime data of clipboard for nullity

tsujan:
`QClipboard::mimeData()` can be null under Wayland (although it hasn't been with Wayland compositors I've tried). Moreover, Qt doesn't guarantee that it's valid everywhere.

The patch adds a nullity check.


tsujan:
Straightforward. Merging...

PR #870, opened 2023-01-01
Announce user's decision about DND

tsujan:
This will be used in `pcmanfm-qt` to fix a small regression that was caused by showing DND menu after finishing DND (for Wayland).


tsujan:
It's a signal, that's used only by pcmanfm-qt. Merging....

PR #872, opened 2023-01-25
Allow typing text tab with inline renaming in all modes

tsujan:
Closes https://github.com/lxqt/pcmanfm-qt/issues/1719


tsujan:
Any objection to merging?

PR #873, opened 2023-01-27
Fixed the thumbnail/icon rectangle in list views

tsujan:
Previously, its width depended on the thumbnail size and was zero for desktop entries without icon. Now it's fixed, like in icon views.

Fixes https://github.com/lxqt/pcmanfm-qt/issues/1720


tsujan:
@stefonarch
I'm using the patch, but could you please confirm that it works fine?

stefonarch:
Sorry forgot about testing that. It is fixed indeed I see.

tsujan:
Thanks!

PR #875, opened 2023-01-29
Generalized the option for disabling smooth scrolling

tsujan:
Now, the option disables smooth scrolling in all view modes.

Unlike before, the tweaks for wheel scroll speed will be preserved if smooth scrolling is disabled (e.g., slower scrolling of big thumbnails).

This PR will be followed by a simple PR for pcmanfm-qt, in which only a label will change.


stefonarch (approved):
I had to test it under X11 as under wayfire it was hard to spot any difference between disabled and enabled (afaik because there's smooth scrolling everywhere, in thunderbird, firefox at least).

GTM

PR #886, opened 2023-03-02
Fixed pressing of ENTER in file dialog

tsujan:
Previously, if ENTER/RETURN was pressed on an active item, the dialog might be accepted twice because the item activation might accept the dialog too. This patch filters out the key press and only activates the item.

Fixes https://github.com/lxqt/libfm-qt/issues/885, among other things.


tsujan:
Any objection to merging?

PR #890, opened 2023-03-12
Add Name=  and Type=  keys for File Properties .directory files

isf63:
According to the XDG Desktop Entry Specification, `Name=` and `Type=` are required desktop entry keys (including for `.directory` files):
https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#recognized-keys

This PR adds a `Type=Directory` key for `.directory` files generated when changing the folder icon with 'File Properties'. Pending any comments on what `Name=` should be, that should be added as well.


tsujan:
Thanks, but that file is used rather like a simple config file; it isn't a real desktop entry. `.directory` has been chosen only for not adding yet another hidden file besides what Dolphin adds. Writing redundant key-value pairs isn't a good idea.

PR #891, opened 2023-03-18
Treat zero-sized files based on their extensions

tsujan:
And only if not possible, use GLib's proposed type. By giving priority to extensions, `libfm-qt` handles zero-sized files like other files (as KDE does, and unlike GNOME)

Previously, they were considered having the same type (`text/plain` before GLib 2.75.1, and `application/x-zerosize` after it).

NOTE: This patch does **not** reverse https://gitlab.gnome.org/GNOME/glib/-/merge_requests/3053 but enhances it. Text editors still need to cover `application/x-zerosize` with GLib ≥ 2.75.1.

### DIFF
--- a/src/core/fileinfo.cpp
+++ b/src/core/fileinfo.cpp
@@ -51,9 +51,22 @@ void FileInfo::setFromGFileInfo(const GObjectPtr<GFileInfo>& inf, const FilePath
 
     size_ = g_file_info_get_size(inf.get());
 
+    type = g_file_info_get_file_type(inf.get());
+
     tmp = g_file_info_get_content_type(inf.get());
     if(tmp) {
-        mimeType_ = MimeType::fromName(tmp);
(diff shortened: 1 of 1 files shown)


stefonarch:
Works fine. Wording has to be changed:

screen_area_sab_20:18:37_

_Please enter file name and extension_?
`new file.extension`

tsujan:
> Wording has to be changed

I agree. I'll do it after this PR is merged.

But we first need an appropriate name. Since it's supposed to be a "blank file", an extension isn't needed, although the user can add one. I think "New file" is enough — like "New folder" for directories.

We can also use "New document", because the name of `application/x-zerosize` is "empty document".

I prefer "New file" just because it's a file, in contrast to a folder. "New document" is also acceptable to me, but I don't know why it sounds a little weird to me.

What do you prefer?

tsujan:
I'll change it to "New file" in the next PR if there's no objection.

PR #896, opened 2023-04-02
Ensure toolbar background is shown behind path-bar

tsujan:
Some Qt styles may fill the path-bar with the button color. This patch removes the auto-filling.

Related to https://github.com/lxqt/pcmanfm-qt/issues/1755

### DIFF
--- a/src/pathbar.cpp
+++ b/src/pathbar.cpp
@@ -81,6 +81,10 @@ PathBar::PathBar(QWidget* parent):
     buttonsLayout_->setSpacing(0);
     buttonsLayout_->setSizeConstraint(QLayout::SetFixedSize); // required when added to scroll area according to QScrollArea doc.
     scrollArea_->setWidget(buttonsWidget_); // make the buttons widget scrollable if the path is too long
+
+    // ensure the parent's background is shown behind the bar
+    scrollArea_->viewport()->setAutoFillBackground(false);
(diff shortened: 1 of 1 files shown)


tsujan:
@stefonarch 

Fusion's tool-bar has a slight color gradient, which should be shown behind the path-bar buttons after this patch (it does here). Previously, the path-bar had a flat background, although its flatness wasn't so noticeable with Fusion.

stefonarch:
I can't spot a difference here but maybe I got something wrong.
screen_area_mar_18:45:40_

tsujan:
> I can't spot a difference here 

That's because you've used the split view ;) Please remove the split view in order to show the path-bar on the toolbar. Then you should see toolbar's mild gradient inside the area of path-bar too (from `#35343b` at the bottom of toolbar to `#37363d` at its top).

tsujan:
Like this, with the default palette:

pcmanfm-qt

stefonarch:
Like that?
screen_area_mar_19:26:48_

Witht dark palettes I can't spot anything.

tsujan:
> Like that?

Exactly. The gradient is visible in your screenshot too.

tsujan:
> Witht dark palettes I can't spot anything.

It exists with the dark palette too; you need a color picker to detect the mild gradient.

PR #900, opened 2023-04-14
Prevent long content search texts from widening search dialog

tsujan:
Previously, if the search history was enabled, long content search texts could widen the search dialog the next time it was shown. If the user entered a very long content search text (by mistake), the search dialog might become unusable the next time.

This is an old issue in Qt and also exists in Qt6. The patch circumvents it by setting the minimum widths of the search boxes.


tsujan:
@stefonarch, sorry for bothering you with testing it, but I think it should be in 1.3.0.

stefonarch (approved):
Works fine.

PR #905, opened 2023-05-13
Use `lxqt-menu-data` instead of `lxmenu-data`

tsujan:
Also, it is made required for compilation because important functionalities depend on it.


stefonarch:
Testing it with `$ export XDG_MENU_PREFIX=` I can see still the lxqt-applications.menu  in `menu://applications` with leave menu and so on, shouldn't it use the fm menu in that case?

stefonarch:
It needed in fact deleting of the cache.

GTM

tsujan:
Thanks for testing it!

PR #909, opened 2023-06-05
Don't select extension on creating new template file

tsujan:
Closes https://github.com/lxqt/pcmanfm-qt/issues/1777

### DIFF
--- a/src/utilities.cpp
+++ b/src/utilities.cpp
@@ -29,6 +29,10 @@
 #include <QStandardPaths>
 #include "fileoperation.h"
 #include <QEventLoop>
+#include <QDialogButtonBox>
+#include <QVBoxLayout>
+#include <QLabel>
+#include <QLineEdit>
 
 #include <pwd.h>
 #include <grp.h>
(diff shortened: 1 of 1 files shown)


isf63:
Tested and works great, thanks! Couldn't find any changes other than the selection.

stefonarch:
Looks ok. I just noticed (again) that those libreoffice files generated automatically (don't remember how) have `.desktop` extensions on arch, while on debian with 1.3 they are ok.
screen_area_mar_17:34:31_

screen_area_mar_17:35:29_

tsujan:
>  I just noticed (again) that those libreoffice files generated automatically

They're a mess that LibreOffice creates. To filter them, check `Preferences → Advanced → Templates → Show only user defined templates in menu`

And thanks for testing! I'll merge it after using it for a while.

stefonarch:
I just checked, on debian I've custom files for calc and writer, and no automatically generated ones, but without that option checked. Forgot about the option though, thanks.

PR #910, opened 2023-06-07
Dialog window icons for Create New folder/blank file

isf63:
Add window icons to the dialogs of Create New → Folder, and Create New → Blank File.

Create New  → \<template\> dialog icons are unchanged (as there seemingly isn't an appropriate icon).

### DIFF
--- a/src/utilities.cpp
+++ b/src/utilities.cpp
@@ -228,16 +228,19 @@ void createFileOrFolder(CreateFileType type, FilePath parentDir, const TemplateI
 
     switch(type) {
     case CreateNewTextFile:
+        parent->setWindowIcon(QIcon::fromTheme(QStringLiteral("document-new")));
         prompt = QObject::tr("Please enter a new file name:");
         defaultNewName = QObject::tr("New file");
         break;
 
     case CreateNewFolder:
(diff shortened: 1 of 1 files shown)


tsujan (changes requested):
A modal dialog inherits the icon of its parent window *if existing*, and the parent icon shouldn't be touched.

modal

isf63:
Ah I see. I was testing on the Desktop and did not test in file manager windows. With my changes, the dialog spawned from file manager windows has the inherited icon. It does work nice on the Desktop however. Not sure what to do.

tsujan:
> Not sure what to do.

Here, there's no problem to solve :) `libfm-qt` isn't limited to pcmanfm-qt, and a basic file like `utilities.cpp` shouldn't be changed because of pcmanfm-qt's desktop.

Moreover, there will be no problem with the icon of that dialog on Desktop if your icon set has an icon for pcmanfm-qt, as mine does:

desktop

As I said in https://github.com/lxqt/pcmanfm-qt/pull/1758, I'll add an icon for pcmanfm-qt. A single SVG icon will solve all problems, without a single code change.

isf63:
Ok, sounds good 👍

PR #913, opened 2023-06-28
Prevent invisible "Create New" dialog under Wayland

tsujan:
Under Wayland and with Qt6, `QWidget::minimumHeight()` is zero for a dialog before it's shown. It's better to be cautious with Qt5 too.


tsujan:
Straightforward; merging....

PR #915, opened 2023-07-15
File dialog: Also add suffixes containing dot on saving

tsujan:
This patch changes a too strong condition which was used in https://github.com/lxqt/libfm-qt/commit/6575eff8ac65ddc7c82005b044f80284e449bd03 to fix a problem. With `\W` in that commit, good extensions like `.tar.gz` weren't added automatically on saving.


tsujan:
To test this, you can open lxqt-archiver, click "New", type a name with extension inside the name entry (like `haha.haha`), and change the file type combo with mouse wheel. Previously, the extension didn't change automatically with types that had dot inside them (like `.tar.xz`).

tsujan:
The change is simple. Merging...

PR #917, opened 2023-08-29
Remember password and anonymity settings of mount dialog

tsujan:
… if they are relevant.

Also, a tooltip is added about the session length.

NOTE: I'll apply the patch to the Qt6 branch after this is merged.


stefonarch:
Tested and all 3 options are working fine (after reboot using `gnome-keyring`).

PR #918, opened 2023-09-03
Add splitIndex parameter to FileLauncher methods

riverbl:
Add splitIndex parameter to some methods on BasicFileLauncher and FileLauncher.

Allows fixing https://github.com/lxqt/pcmanfm-qt/issues/1329.

See also https://github.com/lxqt/pcmanfm-qt/pull/1806.

### DIFF
--- a/src/core/basicfilelauncher.cpp
+++ b/src/core/basicfilelauncher.cpp
@@ -23,7 +23,7 @@ BasicFileLauncher::BasicFileLauncher():
 BasicFileLauncher::~BasicFileLauncher() {
 }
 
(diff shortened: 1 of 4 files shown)


tsujan:
Thanks for trying to add the related feature to `pcmanfm-qt`!

`libfm-qt` is a file management library; `pcmanfm-qt` is just one app which uses it. I'm afraid adding methods to its core classes for such a special use case isn't a good approach.

riverbl:
Cheers for the quick response.

I've modified my pull request to pcmanfm-qt to not need this change.

PR #920, opened 2023-09-06
Fixed a crash on deleting pathbar

tsujan:
The crash would happen if a pathbar was deleted while its path-edit had focus (e.g., by pressing F6 inside path-edit in pcmanfm-qt). The reason is explained in the code comment.

Fixes https://github.com/lxqt/libfm-qt/issues/919


stefonarch (approved):
Can confirm the issue and the  fix for it.

PR #921, opened 2023-09-07
add foot to terminals.list

SArpnt:
only tested on artix linux (with the arch linux package) in pcmanfm-qt but i'm guessing this should work just about everywhere since it's a simple change

### DIFF
--- a/data/terminals.list
+++ b/data/terminals.list
@@ -87,3 +87,7 @@ open_arg=-e
 open_arg=-e
 noclose_arg=--hold -e
 desktop_id=Alacritty.desktop
+
+[foot]
+noclose_arg=--hold
+desktop_id=org.codeberg.dnkl.foot.desktop


tsujan (approved):

Kagukara:
Can you add footclient as well?

```
[footclient]
noclose_arg=--hold
desktop_id=org.codeberg.dnkl.footclient.desktop
```

tsujan:
> Can you add footclient as well?

If you've tested it and it works, yes, of course.

Kagukara:
> > Can you add footclient as well?
> 
> If you've tested it and it works, yes, of course.

Yes it works, I keep having to manually add it every time there's an update.

tsujan:
> Yes it works,

Thanks, it'll be added soon.

> I keep having to manually add it every time there's an update.

Yes, that's a problem that should be fixed at some point, by supporting a user-defined list (→ https://github.com/lxqt/libfm-qt/issues/922).

tsujan:
@Kagukara
If you use git, please test https://github.com/lxqt/libfm-qt/pull/924, with which, you could create your own `~/.local/share/libfm-qt/terminals.list` with just that entry. Otherwise, `footclient` will be in the next release, regardless of the PR.

Kagukara:
I'm using the AUR package `libfm-qt-git`, so I'm unable to test the terminals_data_dirs branch via git. But thank you for adding `footclient` to the next release, very much appreciated.

tsujan:
> I'm using the AUR package libfm-qt-git

OK. @stefonarch and I have tested the PR and will do more tests. Accompanied by https://github.com/lxqt/pcmanfm-qt/pull/1810, it'll enable the users of v1.4.0 to add terminals without needing to edit `~/.local/share/libfm-qt/terminals.list` manually. You might get it sooner, after the two PRs are merged.

Kagukara:
Just got an update for `libfm-qt-git`, and adding: 
```
[footclient]
noclose_arg=--hold
desktop_id=org.codeberg.dnkl.footclient.desktop
```
to `~/.local/share/libfm-qt/terminals.list` adds `footclient` to the list of Terminal emulators. Thank you.

tsujan:
Thanks for testing!

You could have also entered `footclient` in `PCManFM-Qt → Preferences → Advanced → Terminal emulator` if you also updated PCManFM-Qt. Then, if you wanted to edit that file manually (usually not necessary), you could right click inside the combo-box.

Kagukara:
> You could also have entered `footclient` in `PCManFM-Qt → Preferences → Advanced → Terminal emulator` if you also updated PCManFM-Qt.

I had tried this a while ago and just now, it doesn't work. Only way I got it working was to add it to the `terminals.list`.

Kagukara:
Oh, I had to put `footclient -e %s` in, just like the example for terminal. and it worked.

Added this:
```
[footclient]
open_arg=-e
```
to `~/.local/share/libfm-qt/terminals.list`

tsujan:
> (can't just put footclient)

I added it here and it worked:

```
[footclient]
open_arg=
```

Kagukara:
> > (can't just put footclient)
> 
> I added it here and it worked:
> 
> ```
> [footclient]
> open_arg=
> ```

Ah, just did it as well. Having the file deleted with footclient left in Terminal emulator causes it to not work. I had to switch to a different option then add `footclient`.

tsujan:
> Having the terminals.list file deleted with footclient left in Terminal emulator text box causes it to not work.

I think I know what you mean. That file is read only when the dialog is shown; so, the dialog can't know about manual editing before being cancelled and opened again.

stefonarch:
Tried if `$TERM` or exec `$TERM` would work, it doesn't as it gets sanitized. 
Wondering where I've this env set to xterm - in `lxqt-session-config` it's qterminal.

tsujan:
Neither `libfm-qt` nor `libqtxdg` consult `$TERM`. There was an old discussion as to why `$TERM` was useless (in short, there's no standard way of setting the default terminal in Linux).

Apps that depend on `libqtxdg` (e.g., LXQt Panel) use what's set by `qtxdg-mat` (see `qtxdg-mat def-terminal`). PCManFM-Qt can't use it because it doesn't depend on `libqtxdg`; hence it has its own method — that's what we pay for modularity (which is OK, IMO).

tsujan:
Oh, in addition, `libfm-qt` may need more info on terminals (although that doesn't happen usually). That's another reason.

luis-pereira:
> Neither `libfm-qt` nor `libqtxdg` consults `$TERM`. There was an old discussion as to why `$TERM` was useless (in short, there's no standard way of setting the default terminal in Linux).

The  naming for the `TERM`  variable is very misleading. 
`The environment variable TERM contains a identifier for the text window’s capabilities. You can get a detailed list of these cababilities by using the ‘infocmp’ command, using ‘man 5 terminfo’ as a reference. 
`
Source: https://www.gnu.org/software/gettext/manual/html_node/The-TERM-variable.html

> Apps that depend on `libqtxdg` (e.g., LXQt Panel) use what's set by `qtxdg-mat` (see `qtxdg-mat def-terminal`). PCManFM-Qt can't use it because it doesn't depend on `libqtxdg`; hence it has its own method — that's what we pay for modularity (which is OK, IMO).

Couldn't pcmanfm-qt when running in a LXQt desktop read qtxdg-mat config file ?
No additional dependencies and better user experience.

tsujan:
> Couldn't pcmanfm-qt when running in a LXQt desktop read qtxdg-mat config file ?

Yes, that occurred to me too. But I encountered a problem: Should an app like PCManFM-Qt *write* to that file? IMO, the answer is "no".

That being said, it can solve https://github.com/lxqt/lxqt-archiver/issues/366. It can also be used by any app which needs a terminal emulator.

luis-pereira:
> Yes, that occurred to me too. But I encountered a problem: Should an app like PCManFM-Qt _write_ to that file? IMO, the answer is "no".

There is no need to use the `XdgMimeApps::setDefaultTerminal()` API that would introduce a build time dependency, if that's an issue.
`qtxdg-mat` can be used under LXQt. Runtime dependency only.

tsujan:
No, I meant, "Does an app like PCManFM-Qt have *the right* to change the default terminal of LXQt?" I think it doesn't. Of course, it can programmatically, but that doesn't seem OK to me.

tsujan:
@luis-pereira 
Now that you're here, would you please take a look at https://gitlab.freedesktop.org/xdg/xdg-utils/-/merge_requests/48?

luis-pereira:
> No, I meant, "Does an app like PCManFM-Qt have _the right_ to change the default terminal of LXQt?" I think it doesn't. Of course, it can programmatically, but that doesn't seem OK to me.

Under LXQt it can call LXQt default app settings.
p,s. I know that LXQt default app GUI doesn't support Terminal, yet.

tsujan:
> Under LXQt it can call LXQt default app settings.

That seems like a good approach. As you said, we need to add a section to LXQt default app settings. `libfm-qt` will still need its own list, as it contains more info than the terminal command. But its default terminal will be in sync with that of LXQt under LXQt.

All in all, I think it's practical — perhaps the best way.

luis-pereira:
> @luis-pereira Now that you're here, would you please take a look at https://gitlab.freedesktop.org/xdg/xdg-utils/-/merge_requests/48?

Done.

PR #924, opened 2023-09-16
Get terminals list also from data directories

tsujan:
The files `terminals.list` inside data directories take priority over the one provided by `libfm-qt` if existing. How data directories are handled is left to Glib.

In this way, the user can create `~/.local/share/libfm-qt/terminals.list` and see the result in PCManFM-Qt on the fly. (Later, PCManFM-Qt might be patched to create or append to that file.)

Closes https://github.com/lxqt/libfm-qt/issues/922


stefonarch (approved):
Works as expected. I thought first the local list will be the only one displayed but the are merged. The syntax has to be documented maybe.

screen_area_sab_19:44:05_

tsujan:
You were so fast that you tested it before I made a small change and squashed the commits :) That change won't make a difference in most cases, but it's logical.

tsujan:
> The syntax has to be documented maybe.

Maybe. Or I could add the needed functionality to PCManFm-Qt without the user needing to know the syntax.

stefonarch:
I said this because I started with just a list with terminal names without looking at the other file ;)

tsujan:
@stefonarch 
While copy-pasting, I'd missed a line of code, without which, all terminal apps (like htop) were opened in `xterm` from inside PCManFM-Qt. Added it just now. Please update your installation because I'll need it with an upcoming PR for PCManFM-Qt.

stefonarch:
Done, although it had to be in the missed commit as it worked fine.
In a perfect world only installed terminals would be shown, filtering the default list by `.desktop` files in`/usr/share/applications`.

tsujan:
Giving it to git users for more tests...

w-jablonski:
This seems to be breaking my setup. 
My terminal is `wezterm cli spawn --new-window --cwd .` I tried to include this (with or without the initial `wezterm` and with or without quotes ) in the `open_arg=` line but the effect is as if I only passed the single word `wezterm` -- i.e. wezterm starts in $HOME no matter what directory it is invoked from. It used to work before. Should I file a bug report or is there something I am missing? Thanks

tsujan:
@w-jablonski 

Your problem isn't related to this commit but about how wezterm works — libfm-qt didn't accept your command option before this commit either.  At least in In Arch, wezterm comes with `/usr/bin/open-wezterm-here`. Remove the previous command and add `open-wezterm-here` instead.

If there is no `/usr/bin/open-wezterm-here` in your distro, this is what Arch provides:

```sh
#!/bin/sh
# This script is a helper that starts a new terminal window
# in the cwd of the calling process, rather than using the
# default cwd which is usually the home directory.
exec wezterm start --cwd "$PWD" -- "$@"
```

You could add it to your PATH.

BTW, please use https://github.com/lxqt/lxqt/discussions for questions, so that other users could benefit from it too.

w-jablonski:
Thanks!

PR #941, opened 2023-11-30
Add "Copy Location" context menu action

gfgit:
Hello,
I've made a small patch which adds "Copy Location" to file context menu when only one item is selected.
I find it super useful when using terminal a lot, to paste file path in the command line.
Previously when copying files and pasting them inside terminal, it would paste an URI beginning with `file:///` which would break command line tools expecting a normal file path and thus has to be manually removed.

If `edit-copy-path` icon is not present, it automatically falls back to `edit-copy` icon.

### DIFF
--- a/src/filemenu.cpp
+++ b/src/filemenu.cpp
@@ -61,6 +61,7 @@ FileMenu::FileMenu(Fm::FileInfoList files, std::shared_ptr<const Fm::FileInfo> i
     separator1_ = nullptr;
     cutAction_ = nullptr;
     copyAction_ = nullptr;
+    copyLocationAction_ = nullptr;
     pasteAction_ = nullptr;
     deleteAction_ = nullptr;
     unTrashAction_ = nullptr;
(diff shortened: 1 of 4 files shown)


tsujan:
I'm afraid I disagree, on account of the following reasons:

 1. We don't clutter the code of a library because of a use case related to terminals.
 2. PCManFM-Qt already has a "Copy Full Path" action in its Tools menu, and it has a shortcut.
 3. Even if PCManFM-Qt didn't have that action, the user could add a custom action, or he could simply drag and drop file(s) into the terminal…

gfgit:
Ok...
So I can add a custom action to the context menu?

gfgit:
And I can't move this existing action to context menu right?

tsujan:
> And I can't move this existing action to context menu right?

In PCManFM-Qt, that action is intentionally put in the Tools menu for not cluttering the context menu. You could duplicate it or generalize it to the case of multiple files, etc.

You might hear the word "cluttering" from me a lot. It has two meanings: (1) Don't add unnecessary codes (especially codes which duplicate functionalities); and (2) Don't clutter GUIs (as they are in KDE).

PR #943, opened 2023-12-09
Reorder XML .ui files to correct tab orders

isf63:
Reordered XML .ui files to correct tab orders throughout libfm-qt.

There may be one or two .ui files that require <tabstops>, and/or changes to prevent StandardButtons causing incorrect ordering.

Affected files:
- app-chooser-dialog.ui
- edit-bookmarks.ui
- file-operation-dialog.ui
- file-props.ui
- filedialog.ui
- filesearch.ui
- mount-operation-password.ui
- rename-dialog.ui

### DIFF
--- a/src/app-chooser-dialog.ui
+++ b/src/app-chooser-dialog.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>432</width>
-    <height>447</height>
+    <height>387</height>
    </rect>
   </property>
   <property name="windowTitle">
--- a/src/app-chooser-dialog.ui
+++ b/src/app-chooser-dialog.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>432</width>
-    <height>387</height>
+    <height>447</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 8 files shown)


isf63:
This one is more practical as people somewhat-more-often use tabs on customary dialogs. For myself I use tabbing frequently in Firefox, but no so much elsewhere.

tsujan:
Will check it soon; have been very busy with other codes recently.

tsujan:
This is misplaced in `edit-bookmarks.ui`:
```xml
   <item row="2" column="0" colspan="2">
    <widget class="QDialogButtonBox" name="buttonBox">
     <property name="orientation">
      <enum>Qt::Horizontal</enum>
     </property>
     <property name="standardButtons">
      <set>QDialogButtonBox::Cancel|QDialogButtonBox::Ok</set>
     </property>
    </widget>
   </item>
```
In `file-props.ui`, the rows under `groupBox_2` start with `1`, instead of `0`

In `filesearch.ui`, the rows under `groupBox_6` don't start from `0`.

In `mount-operation-password.ui`, rows jump (row 1 is missing).

isf63:
Done. Can `<tabstops>...` be removed in `mount-operation-password.ui`?

tsujan:
> Can <tabstops>... be removed in mount-operation-password.ui?

I haven't checked that part, but if the order is correct and makes sense in practice now, why not?

tsujan:
On second thought, those tabstops are good and should be kept. I'm reviewing your latest changes now…

PR #950, opened 2024-01-05
Disallow input containing '/' in Fm::createFileOrFolder

isf63:
For the "Create New" menu, handling of entry containing '/' is inconsistent: relative paths work if the subdirectories exist; if the path doesn't exist then an error message is shown. Absolute paths do nothing, and lastly there is some funky behavior when the final character is '/'.

Renaming (`Fm::changeFileName`) also disallows '/': https://github.com/lxqt/libfm-qt/blob/eb1de483c106d6ba7786fe543969d487a5155f6f/src/utilities.cpp#L152

This change disallows entry containing '/' because the behavior is inconsistent, and it seems the dialogs weren't intended to be used with paths.

### DIFF
--- a/src/utilities.cpp
+++ b/src/utilities.cpp
@@ -270,6 +270,7 @@ _retry:
         QDialogButtonBox *btns = new QDialogButtonBox(QDialogButtonBox::Ok | QDialogButtonBox::Cancel);
         QObject::connect(btns, &QDialogButtonBox::accepted, &dlg, &QDialog::accept);
         QObject::connect(btns, &QDialogButtonBox::rejected, &dlg, &QDialog::reject);
+
         // layout
         QVBoxLayout *layout = new QVBoxLayout;
         layout->addWidget(label);
@@ -285,6 +286,10 @@ _retry:
         default:
             return;
         }
+    }
+
+    if(new_name.contains(QString::fromUtf8("/"))) {
+        goto _retry;
     }
 
     auto dest = parentDir.child(new_name.toLocal8Bit().data());


tsujan:
Yes, I've known about this for a long time — and thanks for paying attention to it. IMO, it isn't our problem. If it was a bad GLib bug, I'd add a workaround, as I did in at least two severe cases (I never report bugs at GNOME), but this one isn't a big deal.

On the one hand, retrying without showing a message isn't informative, but in this case, we don't want to add a handmade message. On the other hand, it's good to include relative paths and also to let GLib show an error message if they don't exist.

The only inconsistency is about the cases of a starting and an ending `/`: GLib's function `g_file_get_child` ignores the latter case (good) but neither ignores nor shows a message in the former (bad).

isf63:
What does GLib do (if anything) when its relevant functions are called starting with `/`?

Regarding ending with `/`, templates and Blank File shouldn't have a trailing `/`, as that implies a folder.

>it's good to include relative paths and also to let GLib show an error message if they don't exist.

That sounds good.

tsujan:
> What does GLib do (if anything) when its relevant functions are called starting with /?

Handling of `/` depends of the file system (which can be that of MS Windows too). Knowing what GLib does in each case requires reading its code, which is beyond my free time for now.

All I'm saying is this: let GLib handle `/`!

Also, in the case of renaming, it isn't libfm-qt but GLib that handles  `/`.

isf63:
Okay, it's an issue with GLib.

tsujan:
> Okay, it's an issue with GLib.

Yes, an issue that almost no user encounters, and if a user does, nothing disastrous will happen.

PR #952, opened 2024-02-09
File properties: Check same parent dir, timestamps for multiple files

isf63:
Same behavior as single files, but if the values are identical then the value is displayed.
Otherwise, "\<Multiple Values\>" is displayed.

### DIFF
--- a/src/filepropsdialog.cpp
+++ b/src/filepropsdialog.cpp
@@ -284,12 +284,12 @@ void FilePropsDialog::initGeneralPage() {
         }
     } // end if(singleType)
     else { // not singleType, multiple files are selected at the same time
-        ui->fileType->setText(tr("Files of different types"));
+        ui->fileType->setText(tr("<Multiple Values>"));
+        ui->mimeType->setText(tr("<Multiple Values>"));
         ui->target->hide();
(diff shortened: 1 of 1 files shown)


tsujan:
Calling Properties dialog of a multiple selection is useful for setting common permissions or checking total sizes. As for timestamps, they may very rarely be the same, and there are much better ways of comparing them.

So, I'm afraid this extra computation is unnecessary in practice, and that was why it wasn't included in the first place.

isf63:
Timestamps could be left blank. What about the parent directory (Location)? I don't know if it would ever differ in a multiple selection, but leaving Location blank leaves out information.

tsujan:
> What about the parent directory (Location)?

Not needed in practice either. Redundant computation (CPU usage) is avoided, especially with a large number of files. I left this comment intentionally, just as a notice:

```
// FIXME: check if all files has the same parent dir, mtime, or atime
```

isf63:
I still would like to fill some of these fields, but I understand the performance requirements.

Please take a look at my straightforward PRs in libfm-qt, lximage-qt, pcmanfm-qt, and lxqt-panel; mainly dealing with tab order and missing icons.

PR #960, opened 2024-03-18
Workaround for Qt bug in a context menu invoked by Menu key

tsujan:
`QContextMenuEvent` reports the cursor position incorrectly when a context menu is invoked by Menu key in detailed list mode, i.e., in `QTreeView`. Since the bug has nasty effects for us, a small workaround is included.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1882


stefonarch:
Works as expected, GTM

tsujan:
Thanks for the confirmation. As it'll be quite safe if the Qt bug is fixed, merging...

PR #964, opened 2024-04-20
Allow paths in name entry of file dialog

tsujan:
A path can be absolute or relative and can be used with file opening and saving alike.

Also, two minor issues are revealed and fixed after this change:

 * A mistake is corrected in the logic of showing error message about nonexistent files.
 * An issue is fixed in updating accept button on changing directory.

NOTE: Although this isn't self-consistent, Qt's default file dialog supports it. It can also be useful when the path is already copied from somewhere. The extra code is kept to a minimum.

Closes https://github.com/lxqt/libfm-qt/issues/962

### DIFF
--- a/src/filedialog.cpp
+++ b/src/filedialog.cpp
@@ -83,7 +83,12 @@ FileDialog::FileDialog(QWidget* parent, FilePath path) :
                 name += QLatin1Char('.');
                 name += defaultSuffix_;
             }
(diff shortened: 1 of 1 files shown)


tsujan:
@stefonarch 
I've used it and found no problem since the last commit. Please check it whenever you find the time.

stefonarch:
I've used it some time now, didn't notice issues.

PR #972, opened 2024-05-07
Fixed a regression in executing in terminal

tsujan:
Fixes https://github.com/lxqt/pcmanfm-qt/issues/1902

### DIFF
--- a/src/core/terminal.cpp
+++ b/src/core/terminal.cpp
@@ -21,7 +21,7 @@ static void child_setup(gpointer user_data) {
 bool launchTerminal(const char* programName, const FilePath& workingDir, Fm::GErrorPtr& error) {
     /* read user and system terminals files */
     GKeyFile* kf = g_key_file_new();
-    bool found = g_key_file_load_from_data_dirs(kf, "libfm-qt/terminals.list", nullptr, G_KEY_FILE_NONE, &error);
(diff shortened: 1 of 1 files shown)


stefonarch (approved):
I had to rename  `~/.local/share/libfm-qt/terminals.list` to make the error appear. `/usr/share/libfm-qt/`  didn't exist here for some reason, maybe I'm missing the qt5 version.

EDIT: I removed the packages for qt5 yesterday as pacman wanted to update them. instead of adding them to the ignore list...

The fix works, but I don't really get why it produced now an error and first not.
https://docs.gtk.org/glib/struct.KeyFile.html

tsujan:
> The fix works, but I don't really get why it produced now an error and first not.

(1) libfm-qt first searches inside data directories to find a file with a relative path `libfm-qt/terminals.list`, and (2) if no such file exits or if it doesn't contain the searched app, libfm-qt searches in `/usr/share/libfm-qt6`.

Now, GLib has an "error basket" (a name I give to it, not a real name).  Previously, the error basket was filled with the error of the stage (1) when  `libfm-qt/terminals.list` didn't exist, and it remained full at (2), while no error happened at (2). Hence showing that error message. This patch uses the error basket only for (2), ignoring the probable error at (1).

The first version of this PR emptied the error basket before starting (2), but then I remembered that we don't need the basket for (1) at all.

PR #974, opened 2024-05-17
Select name without extension in overwrite prompts

tsujan:
Qt6 automatically selects the whole name in overwrite prompts (while Qt5 selected nothing). This patch excludes the extension — if any — from the selection.


tsujan:
@stefonarch 
Would you please check it before merging?

PR #982, opened 2024-05-30
Give parent to File Properties dialog from context menu

tsujan:
Make it a child of the current window for it to be centered in front of the window.

The same behavior can be implemented for `Alt+Return/Enter` in pcmanfm-qt but, for now, let's have 2 ways of displaying it.

Closes https://github.com/lxqt/libfm-qt/issues/981


stefonarch (approved):
Could reproduce the issue only with firefox/thunderbird filedialogs with `xdg-desktop-portal-lxqt` and the patch works fine.

PR #984, opened 2024-06-08
Add parent to Choose an Application dialog

isf63:
Same as #982 but for the Choose an Application dialog.

Fixes a bug in the Open File dialog where right-clicking a file → Open With → Other Applications, the dialog is inaccessible.

Note: There remains one `AppChooserDialog` ctor with a null parent: https://github.com/lxqt/libfm-qt/blob/ed5fd2f206b31c80b4c2d11e1fbc305bfc7b2144/src/filelauncher.cpp#L80 As I've never used launchers I was unsure what to do there.

### DIFF
--- a/src/filemenu.cpp
+++ b/src/filemenu.cpp
@@ -373,7 +373,7 @@ void FileMenu::onOpenTriggered() {
 }
 
 void FileMenu::onOpenWithTriggered() {
-    AppChooserDialog dlg(nullptr);
+    AppChooserDialog dlg(nullptr, parentWidget() ? parentWidget()->window() : nullptr);
     if(sameType_) {
         dlg.setMimeType(info_->mimeType());
     }


tsujan:
I'm not sure we want to go so far. While https://github.com/lxqt/libfm-qt/pull/982 served as a workaround for file-dialogs that were opened in Firefox and similar apps through portals, App-chooser dialog shouldn't be closed on closing another window, because it's about file association. Yes, I know that the same thing may happen with it through portals, but that's only a special case.

isf63:
The app-chooser dialog is primarily about "How do you want to open this file right now? (And do you want to set its types' default action?)". It's not like lxqt-config-file-associations and primarily about settings. It's unlikely that someone would close the FM before completing the dialog, and also expect the dialog to still be active.

Benefits to this:
- UX improvement with centering, like https://github.com/lxqt/libfm-qt/pull/982.
- Fixes a layering bug in the Open File dialog.

Concerns:
- Closing a FM window with this dialog open triggers a Desktop refresh.
- Closing a Open File dialog with it open, doesn't close the app-chooser.

tsujan:
> UX improvement with centering....

In the case of properties dialog, we have a file/folder inside the view. So, it makes sense to center the dialog, *although it isn't necessary* (and that was why I hadn't made a PR until https://github.com/lxqt/libfm-qt/issues/981 was reported). Associating an app with a mime type goes beyond a single file, and so, centering can't be considered as a benefit.

> Fixes a layering bug in the Open File dialog.

That's not our bug. It may happen with portals or some WMs — it doesn't happen here with Qt apps (= no portal).

PR #991, opened 2024-07-08
Improve side pane spacing

SnAgCu:
Currently the spacing of items in the side pane list view is a bit off, there is less space between the headers and the items next to them than between the child items:
screenshot-07a5d3cb

That happens because the headers have no icons.
This changes that by making the headers taller.
Picture of before and after with the first header focused:
focusbefore focusafter

Tested with different icon and font sizes and styles and scaling.

### DIFF
--- a/src/folderitemdelegate.cpp
+++ b/src/folderitemdelegate.cpp
@@ -46,6 +46,9 @@ FolderItemDelegate::FolderItemDelegate(QAbstractItemView* view, QObject* parent)
     removeIcon_(QIcon::fromTheme(QStringLiteral("list-remove"))),
     fileInfoRole_(Fm::FolderModel::FileInfoRole),
     iconInfoRole_(-1),
+    isHeaderRole_(-1),
+    headerMarginTop_(5),
+    headerMarginBottom_(4),
     margins_(QSize(3, 3)),
     shadowHidden_(false),
     hasEditor_(false) {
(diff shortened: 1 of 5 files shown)


tsujan:
> Currently the spacing of items in the side pane list view is a bit off

It isn't. It's calculated by Qt and `QStyle` and depends on the font size, icon size, and the presence of an icon, among other factors. There's no good reason to add codes for changing it. What you've marked in your screenshot is what you see because of those letters that appear in names.

Also, the "headers" are already distinguished by indentation.

SnAgCu:
>It's calculated by Qt and QStyle and depends on the font size, icon size, and the presence of an icon, among other factors.

Yes, but because the hedaers don't have icons it _looks_ off.

>What you've marked in your screenshot is what you see because of those letters that appear in names.

No, I've measured from the edges of the small letters, it is independent of the actual text:

what2

The "headers" shouldn't be closer to the top item than the space between items, at least all the spacing should be the same.

tsujan:
These screenshots show how the spacing between view items depends on the Qt style (with Fusion, Breeze and Kvantum — Kvantum can give any spacing, depending on its active theme):

1-fusion
2-breeze
3-kvantum


Fusion may draw the most compact item view — and some users may not like it — but it's how it is. We don't add codes to make Fusion or any other style look better ;)

> No, I've measured from the edges of the small letters,

Those are the letters that exist there. Qt calculates the text height for all possible letters.

SnAgCu:
I am not changing the space between the child items, only between the headers, and this is a custom view. All your screenshots show the same issue.

SnAgCu:
Please forget about the implementation for a moment and just look at it as lines of text.
Text line spacing is measured baseline to baseline.
Look at [https://lxqt-project.org/release/2024/06/27/point-releases-qtermwidget-qterminal-2-0-1/](https://lxqt-project.org/release/2024/06/27/point-releases-qtermwidget-qterminal-2-0-1/):

Screenshot 2024-07-09 at 16-34-01 Point Releases Qtermwidget and QTerminal 2 0 1 LXQt

The headings have more space above them.
If we remove that space to make it less than the space between the lines in a paragraph it looks wrong:

Screenshot 2024-07-09 at 16-34-01 Point Releases Qtermwidget and QTerminal 2 0 1 LXQt2

Because it is wrong. And that is how the side pane is. Yes it is less pronounced on some configurations, but it is there.

tsujan:
Please report anything you see as a problem to Qt. Closing this...

PR #992, opened 2024-07-11
Fixed a vulnerability about trusted locations

tsujan:
With the master code, if you made a desktop entry like this (replace YOUR_USER_NAME with your user name):

```
[Desktop Entry]
Name=Test
Comment=
URL=/usr/share/../../home/YOUR_USER_NAME/Downloads/any.desktop
Type=Link
Icon=application-x-desktop
Terminal=false
```

then it would be considered "safe" and execute `/home/YOUR_USER_NAME/Downloads/any.desktop` without prompt. This patch fixes the issue by not relying on the relative paths.

This vulnerability was kindly reported by an attentive user via email. I wasn't sure if he wanted his name to be mentioned here.

NOTE: We can also trust `/usr/local/share/` or more paths, but IMHO there's no need to be too meticulous in this regard.

### DIFF
--- a/src/core/fileinfo.cpp
+++ b/src/core/fileinfo.cpp
@@ -390,15 +390,14 @@ bool FileInfo::isExecutableType() const {
          they are native and have read permission */
         if(isNative() && (mode_ & (S_IRUSR|S_IRGRP|S_IROTH))) {
             if(isShortcut() && !target_.empty()) {
-                /* handle shortcuts from desktop to menu entries:
-                   first check for entries in /usr/share/applications and such
-                   which may be considered as a safe desktop entry path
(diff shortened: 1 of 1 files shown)


stefonarch:
>This vulnerability was kindly reported by an attentive user via email. I wasn't sure if he wanted his name to be mentioned here.

He doesn't:

> Please do not use my name if you plan to credit this issue/vulnerability.

tsujan:
> He doesn't:...

I have telepathy.

stefonarch (approved):
This needs a point release IMO,

tsujan:
> This needs a point release IMO,

A desktop entry with `/usr/share/../../` as its URL is practically nonexistent; there's no real threat.

PR #994, opened 2024-07-11
Prevent truncated labels in prompt for executing

tsujan:
Due do a problem that I think is a Qt bug, the prompt label might not be shown completely, unless the dialog was resized.

Closes https://github.com/lxqt/libfm-qt/issues/993


stefonarch (approved):
Works fine now.

tsujan:
There is a checkbox that appears only when multiple executable files are launched. The problem still happened when it was shown. I fixed it in a commit, which also adds the above-mentioned 10-px spacing.

PR #1001, opened 2024-08-03
Prevent truncated labels in file properties dialog

tsujan:
Due do a Qt problem, long location labels might not be shown completely, unless the dialog was resized. (See https://github.com/lxqt/libfm-qt/pull/994 for a similar case.)


isf63:
Works as intended. The other issue seems to be related to related to `formLayout_2` and `girdLayout_3` when Location has more than one line. (Although it may happen in other cases)

tsujan:
Qt has an old problem in setting the height of a dialog when it contains long wrapped labels, and its workaround is simple, as done here. Other layout problems of Qt don't affect UX to this extent and should be fixed in Qt itself.

PR #1004, opened 2024-08-10
Enabled the setting about local-only thumbnails

tsujan:
The code relies on the attribute `G_FILE_ATTRIBUTE_FILESYSTEM_REMOTE`; whether it's set correctly or not is the responsibility of GLib/GVFS.

Closes https://github.com/lxqt/libfm-qt/issues/643

NOTE1: To be on the safe side, recompile all apps that are based on libfm-qt — but I didn't :P

NOTE2: I haven't seen an external thumbnailer that works with remote files. So, this is mostly about libfm-qt's internal thumbnailer.

### DIFF
--- a/src/core/filesysteminfojob.cpp
+++ b/src/core/filesysteminfojob.cpp
@@ -8,7 +8,8 @@ void FileSystemInfoJob::exec() {
             g_file_query_filesystem_info(
                 path_.gfile().get(),
                 G_FILE_ATTRIBUTE_FILESYSTEM_SIZE","
-                G_FILE_ATTRIBUTE_FILESYSTEM_FREE,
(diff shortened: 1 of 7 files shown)


tsujan:
@stefonarch 
I tested it with `sftp` and, only years ago, with `smb`. Please test it with `smb` when you find the time (there's no rush).

tsujan:
Unfortunately this approach has a problem: If the first remote folder contains PNG/JPG images, some of them will be thumbnailed regardless of the setting, although that will be corrected on reloading. The reason is that the code gets the value of `G_FILE_ATTRIBUTE_FILESYSTEM_REMOTE` shortly after loading the first folder, while it supposes that folders are local by default.

tsujan:
Closing this PR because of the above-mentioned problem...

PR #1007, opened 2024-08-13
Detect remote and read-only file systems

tsujan:
The patch does it only by a single info query for the parent directory. So, it shouldn't affect the speed.

Also, the setting about local-only thumbnails is enabled in this way.

Closes https://github.com/lxqt/libfm-qt/issues/643 and fixes https://github.com/lxqt/libfm-qt/issues/1006

NOTE: To be on the safe side, recompile libfm-qt based apps — although the LXQt apps don't seem to need that.

REMINDER: Add an option to lximage-qt for local-only thumbnails if this PR is merged.

### DIFF
--- a/src/core/dirlistjob.cpp
+++ b/src/core/dirlistjob.cpp
@@ -51,6 +51,29 @@ _retry:
         return;
     }
     else {
+        // First set the attributes "filesystem::readonly" and "filesystem::remote",
+        // which will be queried by FileInfo and are useful only for the parent dir.
+        if(GFileInfoPtr fs_info{
+            g_file_query_filesystem_info(dir_gfile.get(),
+                                         G_FILE_ATTRIBUTE_FILESYSTEM_READONLY","
+                                         G_FILE_ATTRIBUTE_FILESYSTEM_REMOTE,
+                                         cancellable().get(), nullptr),
+            false
+        }) {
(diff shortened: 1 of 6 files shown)


stefonarch (approved):
Works perfect fine in samba, my only ftp connection is somewhat slow so no reliable test possible.

Detecting read-only is just a second feature, right? I can't see any thumbnails ever atm in lximage-qt.

stefonarch:
Checked in master lximage-qt/libfm-qt and remote thumbnails are always shown. Does it need a setting now?

tsujan:
> Checked in master lximage-qt/libfm-qt and remote thumbnails are always shown.

On the contrary, they'll never be shown after this.

> Does it need a setting now?

Yes. See `REMINDER` in  the PR comment.

tsujan:
Although this works as intended and I'm happy that I've found the shortest solution, I ask myself this question: Do we need to do it, while the user can disable thumbnails whenever he wants? Add to this the fact that remoteness is detected only if GVFS has a role. For example, Google Drive isn't seen to be remote when mounted by rclone.

The read-only part isn't used anywhere. And the logical inconsistency of https://github.com/lxqt/libfm-qt/issues/1006 can be solved in an elementary way.

@stefonarch, what do you think? If we decide to ignore remoteness (as it's already ignored), we should remove its settings from pcmanfm-qt.

stefonarch:
I think making work a non-working option is nice, although nobody complained about it.

PR #1010, opened 2024-08-15
Use our own bookmarks file

tsujan:
The new file is `~/.local/share/libfm-qt/bookmarks`. If it doesn't exist or is empty, as soon as a libfm-qt based app is launched, the GTK3 file `~/.config/gtk-3.0/bookmarks` will be consulted for backward compatibility, then the new file will be created and the GTK3 file will be ignored.

Closes https://github.com/lxqt/libfm-qt/issues/1009

NOTE: The portal file dialogs will see the change after a reboot (or after restarting user portal services).


tsujan:
Oops! I forgot to consider the case where the folder `~/.local/share/libfm-qt/` doesn't exist yet. Will fix it soon.

stefonarch:
>as soon as a libfm-qt based app is launched

Here it starts using the new bookmarks file only after adding/removing one, did even a relogin.

tsujan:
> Here it starts using the new bookmarks file only after adding/removing one

Yes, it does, but even before that, although the user may not notice. Provided that `~/.config/gtk-3.0/bookmarks` exists, pcmanfm-qt creates new bookmarks file by copying it on startup and uses it since then.

You can check that by removing the folder `~/.local/share/libfm-qt/`, stopping the process of pcmanfm-qt, and starting it again: the new file should be created as a copy of the GTK3 file, and any manual change of it should be reflected by pcmanfm-qt.

stefonarch:
>You can check that by removing the folder ~/.local/share/libfm-qt/, stopping the process of pcmanfm-qt, and starting it again: the new file should be created as a copy of the GTK3 file, and any manual change of it should be reflected by pcmanfm-qt.

As I said: No new file here that way
```

stef@archlinux:libfm-qt$ pwd
/home/stef/.local/share/libfm-qt
stef@archlinux:libfm-qt$ ls
terminals.list
```
`bookmarks` only appears after changing one.

What about the gtk- filemanager issue? I think that's a showstopper, having different bookmarks in different apps/filedialogs.

tsujan:
> As I said: No new file here that way

That shouldn't be possible if you have a valid, non-empty `~/.config/gtk-3.0/bookmarks`.

> What about the gtk- filemanager issue?

It isn't an issue but a "fix" (as I mentioned in https://github.com/lxqt/libfm-qt/issues/1009): A non-GTK app shouldn't affect GTK bookmarks. I remember that someone had complained about it a long time ago.

tsujan:
> having different bookmarks in different apps

That's why portals exist.

stefonarch:
>That's why portals exist.

Portal isn't working in libreoffice, gimp, shotwell and every other GTK app I tried, except firefox/thunderbird.

tsujan:
> Portal isn't working in libreoffice, gimp...

I know. That's not our problem. Only by chance we used a GTK-only setting before this patch: a remnant of the GTK-based pcmanfm. As I said in https://github.com/lxqt/libfm-qt/issues/1009, even GTK4 seems to use the portal now.

I don't think KDE/GNOME users ever complain about the difference between KDE and GTK bookmarks. They may only complain that this or that GTK/KDE app doesn't use KDE/GTK file dialog inside KDE/GNOME respectively.

stefonarch:
For me it's a regression, as I can't anymore update bookmarks shown in gtk filedialogs. If `xdg-desktop-portal-lxqt` would work everywhere ok, but it doesn't.

tsujan:
> For me it's a regression...

"Regression" has a definition far from that. This change breaks a "bad habit" ;)

> We have that feature of shared bookmarks - why remove it?

For two reason (already mentioned — they motivated making of the PR):

 1. Previously we changed a config file that wasn't related to us at all (a "bad habit", or even a bug); and
 2. GTK changes in a backward incompatible way (as it has already changed in its version 4) and we have no reason to follow it.

stefonarch:
I'm pragmatic: regression is when something worked and after it doesn't anymore.

For 2) Afaik the only GTK4 app which writes to bookmarks file in foreseeable future  is nautilus, and it still writes to `~/.config/gtk-3.0/bookmarks`.
 I would agree to a "regression" if there would be an incompatibility in sight.

tsujan:
Not only `~/.config/gtk-3.0/bookmarks` isn't ours to change, but also the original code had this problem that it wouldn't create it if the folder `~/.config/gtk-3.0/` didn't exist; instead, it used the terrible file `~/.gtk-bookmarks`.

However I look at it, changing the contents of a GTK3 config file seems like an old mistake. It should be corrected now.

And I don't see any reason to care for apps that want to have their own file dialogs. This is Krita's — check it's bookmarks:

krita

tsujan:
I imagine ourselves becoming like Wayland devs who talk for months before going forward ;) GNOME devs are at the other extreme and may talk only after the fact.

Let's focus here on whether this PR does its job as intended, i.e., the review. Of course, I won't merge it if you or other members don't agree with the philosophy behind it.

stefonarch (approved):
It works as intended. If it's a good idea needs more discussion IMO. We are writing also to other gtk files (after setting a hook) in `lxqt-appearance`.
An option "share bookmarks with GTK apps"?

tsujan:
> If it's a good idea needs more discussion IMO.

I won't merge it without a discussion. But to me, it's inevitable; if not now, in future.

> An option "share bookmarks with GTK apps"?

An overdoing, IMO. Sooner or later, we should say goodbye to such things.

tsujan:
I use this without problem since I made it. Should we wait more? I have no problem with waiting though.

tsujan:
And this is `gtk4-widget-factory` opening a file:

gtk4

The old times have passed ;)

stefonarch:
In theory you're right, in practise users would loose bookmarks over time in all libreoffice apps, inkscape and whatever else.

PR #1012, opened 2024-08-18
Fixed memleaks with not free alloc MountOperation

GermanAizek:
@luis-pereira @tsujan make code review PR changes because I'm not sure if everything is correct, this is the first time I understand code LxQt.


tsujan:
`MountOperation` auto-destroys.

> this is the first time I understand code LxQt.

Good PRs are appreciated, but please read codes before making PRs.

PR #1014, opened 2024-10-13
Enhanced the detection of changes in folder config

tsujan:
… and so, prevented redundant writing.

Confession: I've used this patch for 4 years but forgot to make a PR ;)


tsujan:
Straightforward. Merging...

PR #1015, opened 2024-10-15
Fixed the scaled pixmap of `Fm::IconEngine`

tsujan:
Scaling is taken into account separately for Qt < 6.8 and Qt ≥ 6.8.

Closes https://github.com/lxqt/libfm-qt/issues/788

### DIFF
--- a/src/core/iconinfo_p.h
+++ b/src/core/iconinfo_p.h
@@ -103,9 +103,17 @@ bool IconEngine::isNull() {
     return info ? info->internalQicon().isNull() : true;
 }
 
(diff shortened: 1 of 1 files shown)


tsujan:
With Qt 6.7.3 and a scale factor of 2, the symbolic icons of my icon set remain symbolic (because the set has 24x2 folders):

scaled

This shows that the scaled pixmap is correct.

tsujan:
Luckily I was able to test this with both Qt 6.7.3 and Qt 6.8.0 (see version checks) , and it worked fine. Any objection to merging?

PR #1016, opened 2024-10-20
Silenced a compilation warning about `GDestroyNotify`

tsujan:
The problem was that `menu_cache_item_unref` returned `gboolean` (`FALSE`) but we needed `void` to cast to `GDestroyNotify`.

### DIFF
--- a/src/appmenuview.cpp
+++ b/src/appmenuview.cpp
@@ -67,6 +67,11 @@ AppMenuView::~AppMenuView() {
     }
 }
 
+// To avoid incompatible cast to GDestroyNotify:
+static inline void menu_cache_item_unref0(MenuCacheItem* item) {
+    menu_cache_item_unref(item);
+}
+
 void AppMenuView::addMenuItems(QStandardItem* parentItem, MenuCacheDir* dir) {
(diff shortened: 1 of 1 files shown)


stefonarch (approved):
Only one left:
```
[ 83%] Building CXX object src/CMakeFiles/fm-qt6.dir/filedialoghelper.cpp.o
In file included from /usr/include/qt6/QtCore/6.8.0/QtCore/private/qobject_p.h:28,
                 from /usr/include/qt6/QtGui/6.8.0/QtGui/private/qdnd_p.h:27,
                 from /tmp/makepkg/libfm-qt-git/src/libfm-qt/src/xdndworkaround.cpp:33:
/usr/include/qt6/QtCore/6.8.0/QtCore/private/qproperty_p.h:311:16: warning: ISO C++ prohibits anonymous structs [-Wpedantic]
  311 |         struct {
      |                ^
[ 84%] Linking CXX shared library libfm-qt6.so
[ 84%] Built target fm-qt6
````

tsujan:
> Only one left:

That one belongs to Qt, not us.

PR #1017, opened 2024-10-21
Modernized string literals

tsujan:
`QLatin1StringView` is used instead of `QLatin1String`.

Also, instead of `QLatin1StringView` and `QStringLiteral`, the operators `_L1` and `_s` are used respectively when a source file contains many string literals. This is supported since Qt 6.4 and makes codes more readable as well as easier to write.

NOTE: This may seem like a heavy patch, but the changes should be made sooner or later — also in other sources. Here I preferred sooner rather than later. Will do it for pcmanfm-qt too.


tsujan:
Should this "monstrosity" be merged before 2.1.0?

tsujan:
I read it again and it was OK. And I've used it since 2.0.0. Merging...

PR #1028, opened 2024-12-21
Removed menu-cache version checks from `vfs-menu.c`

tsujan:
Since libfm-qt depends on menu-cache 1.1.0, those version checks were redundant and only made the code hard to read.

### DIFF
--- a/src/core/vfs/vfs-menu.c
+++ b/src/core/vfs/vfs-menu.c
@@ -30,21 +30,8 @@
 #include <glib/gi18n-lib.h>
 #include <menu-cache.h>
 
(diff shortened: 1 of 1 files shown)


isf63:
There are also checks for old versions of Glib in other spots

From CMakeLists.txt
https://github.com/lxqt/libfm-qt/blob/58ca7b5f8038a31468e7ffd5f6eec879e0a06bfe/CMakeLists.txt#L29

tsujan:
I know. They don't damage code readability. But these checks made the code of that file unreadable, for no good reason.

PR #1030, opened 2025-01-05
File Properties dialog: Use IBeam cursor for selectable labels

isf63:
Currently the user may not know the labels are selectable until happening to test.

Windows has this behavior, I think it's good UI/UX.

### DIFF
--- a/src/file-props.ui
+++ b/src/file-props.ui
@@ -98,6 +98,9 @@
          <property name="textInteractionFlags">
           <set>Qt::LinksAccessibleByMouse|Qt::TextSelectableByMouse</set>
          </property>
+         <property name="cursor">
+          <set>Qt::IBeamCursor</set>
+         </property>
         </widget>
        </item>
        <item row="3" column="0">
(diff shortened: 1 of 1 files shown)


tsujan:
Intentionally avoided (by me) because (1) it's ugly, and (2) in general, showing a text cursor inside an uneditable text can be confusing.

And now, there's one more reason: it doesn't work on Wayland.

isf63:
>avoided, because (1) it's ugly

It _is_ sort of. Also selecting text with a regular cursor is visually unappealing (and incorrect).

> (2) in general, showing a text cursor inside an uneditable text can be confusing

The main confusion with these labels is that people will only select them once knowing they are selectable - there is no visual cue.

> Wayland

There must be a bug there.

tsujan:
I added selectability only later, as an extra feature. If it's confusing (which isn't, IMO), it shouldn't be made worse by adding an ugly text cursor there.

isf63:
I suspected this was a debatable change by seeing that this wasn't done anywhere else. I'll close this.

If you find time please view my other active PRs.

tsujan:
Yes, lack of time… but they won't be forgotten.

Issue #1031, opened 2025-01-07
Inconsistency about navigating into devices when mounting


isf63:
**Is your feature request related to a problem? Please describe.**

This is about the side pane and clicking under Devices to mount devices.

After a fresh boot, clicking a device (and optionally entering credentials) and mounting does not navigate to the device root folder. The mount was successful however, but the device must be clicked again.

Unmounting, and clicking the device again to mount goes directly into the device root folder.

**Describe the solution you'd like**
Click and middle click on devices consistently go to the root folder.

**Describe eventual alternatives you've considered**

##### Context
I assumed this happened on the first mounts per login. I tested and it was first mount per device, after first booting.
I tested whether it was about entering credentials, and it's not, after letting authentication timeout.

tsujan:
@isf63 
Your test would be appreciated: https://github.com/lxqt/libfm-qt/pull/1032

PR #1032, opened 2025-01-07
Try better to change directory on mounting by clicking in side-pane

tsujan:
Sometimes the signal "mount-changed" is emitted by GIO with a delay when a volume is mounted by activating its side-pane item, and so, the directory isn't changed. This patch is a workaround for such cases.

Closes https://github.com/lxqt/libfm-qt/issues/1031


tsujan:
@stefonarch 
If you can reproduce the issue (see https://github.com/lxqt/libfm-qt/issues/1031), please review it by testing.

stefonarch:
> If you can reproduce the issue (see #1031), please review it by testing.

Done on both machines, GTM
 In fact it's easy to reproduce, didn't notice it first.

tsujan:
Ah, thanks!

isf63:
Works! Also tested middle click, and a few cold boots

tsujan:
Thanks for testing!

PR #1033, opened 2025-01-25
Add colon to remaining label

isf63:
All labels that should have colons (per pointing to an input) had colons, except one.

**Steps:**
1. `designer6 $(find . -name "*.ui")` to check labels. Thankfully, there was only one label with a missing colon.
2. `grep ":</translation>" src/translations/*` to ensure the existing translations all use the ASCII colon.

The only differing language convention was French, with `<space>:` instead of `:`.

The `line="75"` in `<location filename="../file-operation-dialog.ui" line="75"/>` is the only thing that looks suspicious.

### DIFF
--- a/src/file-operation-dialog.ui
+++ b/src/file-operation-dialog.ui
@@ -82,7 +82,7 @@
      <item row="3" column="0">
       <widget class="QLabel" name="label_4">
        <property name="text">
-        <string>Progress</string>
+        <string>Progress:</string>
        </property>
       </widget>
      </item>
--- a/src/translations/libfm-qt.ts
+++ b/src/translations/libfm-qt.ts
@@ -158,7 +158,7 @@
     </message>
     <message>
         <location filename="../file-operation-dialog.ui" line="75"/>
-        <source>Progress</source>
+        <source>Progress:</source>
         <translation type="unfinished"></translation>
     </message>
     <message>
--- a/src/translations/libfm-qt_ar.ts
+++ b/src/translations/libfm-qt_ar.ts
@@ -164,8 +164,8 @@
     </message>
     <message>
         <location filename="../file-operation-dialog.ui" line="75"/>
-        <source>Progress</source>
-        <translation>التقدّم</translation>
+        <source>Progress:</source>
+        <translation>التقدّم:</translation>
     </message>
     <message>
         <location filename="../file-operation-dialog.ui" line="95"/>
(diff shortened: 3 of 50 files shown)


isf63:
Searching for "U+00A0" in Featherpad highlights normal spaces, there also is no visual distinction when I enable Preferences > Text > Show whitespaces.

I tried using grep but it failed to highlight matches.

tsujan:
spaces

tsujan:
By checking the translations of my apps, I saw that `:` is replaced by `&#xa0;:` in French.

This has always given us a headache because
 1. Some French translators use a simple space, which can't be correct; and
 2. Some other languages may use something similar but different.

isf63:
I used Python to get the bytes of `src/translations/libfm-qt_fr.ts`:
```
>>> file = open('src/translations/libfm-qt_fr.ts')
>>> contents = file.read()
>>> [pos for pos, char in enumerate(contents) if char == ':']
[ ... ]
>>> memoryview(contents[n:m][0].encode()).tolist() # Position of the "space" preceding the colon in "<translation>Plus petits que :</translation>"
[32]
# Checked several other strings
```

For all the existing translations in `_fr` it looks like they use a normal space.

stefonarch:
It's weblate which replaces `&#xa0;:` at every new translation in that file, not the translators, and I use to fix it by updating all translation files immediately before a release. So as long it's in the source it should be ok. It also seen in catalan and some other language I don't remember.

I didn't look into it but for the LTR languages the position of the colon is ok?

tsujan:
> and I use to fix it by updating all translation files immediately before a release

No fix is needed. It's correct, and the French GUI is OK.

So, it's *you* that insert those spaces ;)

tsujan:
> I didn't look into it but for the LTR languages the position of the colon is ok?

You mean RTL. That could be found by comparison; there are other colons in the same files.

isf63:
Arabic, for instance, puts `:` immediately before `</translation>`.
```
[ ~/Programming/git/libfm-qt/src/translations] grep ":<" libfm-qt_ar.ts
```
Same for `_he`, and all others IIRC.

stefonarch:
> > and I use to fix it by updating all translation files immediately before a release
> 
> No fix is needed. It's correct, and the French GUI is OK.

Repeating: Every time a french or other file containing those narrow spaces is coming in from weblate all those spaces are reverted to normal ones and running  `lxqt2-transupdate` fixes that.

tsujan:
I regularly use `lupdate` with my apps; `&#xa0;` and normal spaces both remain as they are, in French and other languages.

stefonarch (approved):
Looks ok.

PR #1034, opened 2025-02-08
Replace `qMin`,`qMax`,`qBound`,`qRound`,`qAbs` with standard funcs

isf63:


tsujan (approved):
Nice. So fast!

Merge it if you can't find another case; otherwise, please add it and wait for my next review.

Thanks!

src/folderview.cpp
```diff
@@ -709,7 +709,7 @@ void FolderViewTreeView::layoutColumns() {
                     }
                 }
                 opt.section = columnId;
-                w = qMax(sizeHintForColumn(columnId),
+                w = std::max(sizeHintForColumn(columnId),
                          style()->sizeFromContents(QStyle::CT_HeaderSection, &opt, QSize(),
                                                    headerView).width());
```
tsujan, line 714: Please add four spaces before `style()->` and `headerView)` for them to be aligned with the changed line, like before.

src/folderview.cpp
```diff
@@ -726,7 +726,7 @@ void FolderViewTreeView::layoutColumns() {
 
                 // Compute the minimum acceptable width for the filename column, showing
                 // whole texts whose lengths are less than 30 times the average font width.
-                int filenameMinWidth = qMin(iconSize().width()
+                int filenameMinWidth = std::min(iconSize().width()
                                             + 30 * opt.fontMetrics.averageCharWidth(),
                                             sizeHintForColumn(filenameColumn));
```
tsujan, line 731: Again, four spaces before `+ 30` and `sizeHintForColumn`.

tsujan:
@isf63 
Although the compilation seems fine — due to what's included in some basic Qt headers — there is a logical problem: `std::max`, `std::min` and `std::clamp` are provided by `<algorithm>`, while `std::round` and `std::abs` by `<cmath>`. The former doesn't include the latter.

As you correctly included `<algorithm>` where needed, please also do so about `<cmath>`.

isf63:
Made PRs for the previous two commits. Will update lxqt-config/lxqt-panel ones soon.

tsujan:
Thanks! I'll check them later.

PR #1036, opened 2025-02-09
Add missing <cmath> headers

isf63:
To check:
```
grep -R -P "(std::abs|std::round)"
grep -R "#include <cmath>"
```

### DIFF
--- a/src/filepropsdialog.cpp
+++ b/src/filepropsdialog.cpp
@@ -30,6 +30,7 @@
 #include <QFileDialog>
 #include <sys/types.h>
 #include <ctime>
+#include <cmath>
 #include <algorithm>
 #include "core/totalsizejob.h"
 #include "core/folder.h"
--- a/src/filepropsdialog.cpp
+++ b/src/filepropsdialog.cpp
@@ -31,7 +31,6 @@
 #include <sys/types.h>
 #include <ctime>
 #include <cmath>
-#include <algorithm>
 #include "core/totalsizejob.h"
 #include "core/folder.h"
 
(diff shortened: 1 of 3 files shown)


tsujan:
I'll review all of them soon — being a little busy for a while — especially those related to libfm-qt and pcmanfm-qt. Thanks!

src/filepropsdialog.cpp
```diff
@@ -30,6 +30,7 @@
 #include <QFileDialog>
 #include <sys/types.h>
 #include <ctime>
+#include <cmath>
 #include <algorithm>
 #include "core/totalsizejob.h"
```
tsujan, line 35: `algorithm` isn't needed here.

src/proxyfoldermodel.cpp
```diff
@@ -23,6 +23,7 @@
 #include <QCollator>
 #include <QApplication>
 
+#include <cmath>
 #include <algorithm>
```
tsujan, line 27: Ditto.

tsujan:
@isf63 
Which version of Qt did you use to compile libfm-qt and pcmanfm-qt after these changes?

I have a reason to believe that including `<cmath>` may cause a problem with Qt 6.7 in Debian, because of an issue in that version. But I can't test, having Qt 6.8.2.

isf63:
If Qt has a problem with include <cmath> then it has a serious problem. I have the updated regular Qt5 and 6 versions in Arch.

tsujan:
> If Qt has a problem with include <cmath> then it has a serious problem.

I agree completely.

Sorry, it was a false alarm, caused by a totally unrelated problem in Debian.

PR #1040, opened 2025-03-01
ark: support zstandard

es20490446e:


tsujan (changes requested):
Please don't change the order — it isn't alphabetical intentionally, and we prioritize lxqt-archiver  — but only the Ark section.

es20490446e:
The shorting of this list doesn't affect which archiver is selected by default, I have tested it.

The default is always "file-roller", even if it is not installed or selected as handler of archives.

Hence I don't see how shorting them in an unspecified order is useful. It's just personal preference.

stefonarch:
> The default is always "file-roller", even if it is not installed or selected as handler of archives.

I've checked a clean user, default is `lxqt-archiver` here. Probably if it's not installed it's fileroller.

tsujan:
It's about how the apps are offered. We don't want to put Ark on the top inside the popup.

es20490446e:
@stefonarch I have tried:

1. Close pcman
2. Uninstall file-roller
3. Install lxqt-archiver
4. From the terminal: `rm -r "${HOME}/.config/pcmanfm-qt"`
5. Launch pcman and see which is the default archiver.

Result:
file-roller

So probably the code selecting the archiver is somewhere else, let me check 🔍

es20490446e:
You can see that `file-roller` is hard-coded in [settings.cpp](https://github.com/lxqt/pcmanfm-qt/blob/master/pcmanfm/settings.cpp#L218).

So what criteria do you want to use to short the archivers?

I personally think that, being that it doesn't affect the default, just shorting them alphabetically.

Alternatively putting one recommended one at the top... but then what should be the recommended one? `lxqt-archiver` or `file-roller`?

I personally see `lxqt-archiver` as inferior to a bunch of other options listed there. That's what I say skip the dilemma and short alphabetically.

And, maybe, make `pcman` autodetect which archivers are installed during first launch.

tsujan:
https://github.com/lxqt/pcmanfm-qt/blob/36e432b0058073e1ca4571beadc04c28061ab9be/config/pcmanfm-qt/lxqt/settings.conf.in#L5

es20490446e:
In my system this file is located at:
`usr/share/pcmanfm-qt/lxqt/settings.conf`

I guess this file is only loaded when you launch the `lxqt` desktop environment, isn't it?

tsujan:
Yes, it's for an LXQt session, whether under X11 or under Wayland.

es20490446e:
So what shorting do you want?

tsujan:
As mentioned above, a PR shouldn't do anything other than what it's about. Please just don't change the current sorting.

Also, are you sue about `--changetofirstpath`? Does it work for you as expected?

es20490446e:
From the [Ark Manual](https://linuxcommandlibrary.com/man/ark):

>  `--changetofirstpath`: Change the current directory to the first entry and add all other entries relative to this one.

Which is an obscure way to say that files won't be put within a sub-directory, named as the working directory. I tested it.

Let me adjust this shorting 🔧

es20490446e:
Ready 👌

tsujan:
Thanks, but still other parts of the file have been changed for no good reason — the place of `supports_uris=true` is changed and a comment is removed; why?! Please fix that too; there's only one step to merging.

tsujan:
There's also another problem:

I downloaded the ark 24.12.2-1 and checked its desktop file. It has this:
```
MimeType=application/x-deb;application/x-cd-image;application/x-bcpio;application/x-cpio;application/x-cpio-compressed;application/x-sv4cpio;application/x-sv4crc;application/x-rpm;application/x-compress;application/gzip;application/x-bzip;application/x-bzip2;application/x-lzma;application/x-xz;application/zlib;application/zstd;application/x-lz4;application/x-lzip;application/x-lrzip;application/x-lzop;application/x-source-rpm;application/vnd.debian.binary-package;application/vnd.efi.iso;application/vnd.ms-cab-compressed;application/x-xar;application/x-iso9660-appimage;application/x-archive;application/x-tar;application/x-compressed-tar;application/x-bzip-compressed-tar;application/x-bzip2-compressed-tar;application/x-tarz;application/x-xz-compressed-tar;application/x-lzma-compressed-tar;application/x-lzip-compressed-tar;application/x-tzo;application/x-lrzip-compressed-tar;application/x-lz4-compressed-tar;application/x-zstd-compressed-tar;application/x-7z-compressed;application/vnd.rar;application/zip;application/x-java-archive;application/x-lha;application/x-stuffit;application/x-arj;application/arj;
```
Please copy-paste it here. `mime_types` should be what the app provides.

es20490446e:
When I improve code I get into the habit of refactoring it to be clearer. Otherwise making a pull request just to a small clean up is a bit silly, and probably it won't be done.

I removed the comment because it no longer seems to hold true. If you use URI it seems to work.

I put `supports_uris=true` at the top, because the next line is the one that will be using them. So it follows a logical progression, and you can check right away if one line matches the other.

```
[lxqt-archiver]
supports_uris=true
create=lxqt-archiver --add %U
extract=lxqt-archiver --extract %U
extract_to=lxqt-archiver --extract-to %d %U
```

Shall I revert that too?

tsujan:
> Shall I revert that too?

Yes, please: everything that's not related to the purpose of the PR. It may seem trivial to you, but it helps a lot in finding probable regressions in general.

tsujan:
> I removed the comment because it no longer seems to hold true.

Bring it back too.

es20490446e:
Done 👌

tsujan (approved):

PR #1049, opened 2025-04-14
Try to stop or eject drive before ejecting volume from side-pane

tsujan:
Lack of drive stopping/ejection caused USB sticks to remain powered after they were ejected by the app.

Closes https://github.com/lxqt/libfm-qt/issues/1048

NOTE 1: As far as I know, only USB sticks and optical media support ejection, in different senses. An external hard drive can only be unmounted, partition by partition.

NOTE 2: Please don't merge before 2.2.0 is released.

### DIFF
--- a/src/mountoperation.cpp
+++ b/src/mountoperation.cpp
@@ -207,6 +207,24 @@ void MountOperation::onEjectVolumeFinished(GVolume* volume, GAsyncResult* res, Q
     delete pThis;
 }
 
+void MountOperation::onStopDriveFinished(GDrive* drive, GAsyncResult* res, QPointer<MountOperation>* pThis) {
+    if(*pThis) {
+        GError* error = nullptr;
+        g_drive_stop_finish(drive, res, &error);
+        (*pThis)->handleFinish(error);
+    }
+    delete pThis;
+}
+
+void MountOperation::onEjectDriveFinished(GDrive* drive, GAsyncResult* res, QPointer<MountOperation>* pThis) {
+    if(*pThis) {
+        GError* error = nullptr;
+        g_drive_eject_with_operation_finish(drive, res, &error);
+        (*pThis)->handleFinish(error);
+    }
+    delete pThis;
+}
+
 void MountOperation::onEjectFileFinished(GFile* file, GAsyncResult* res, QPointer<MountOperation>* pThis) {
     if(*pThis) {
         GError* error = nullptr;
--- a/src/mountoperation.h
+++ b/src/mountoperation.h
@@ -93,6 +93,21 @@ public:
             prepareUnmount(mnt);
             g_object_unref(mnt);
         }
+        // The current function is called when the volume can be ejected,
+        // but we first check if the drive can be stopped or ejected.
+        if(GDrive* drv = g_volume_get_drive(volume)) {
+            if(g_drive_can_stop(drv)) {
+                g_drive_stop(drv, G_MOUNT_UNMOUNT_NONE, op, cancellable_, (GAsyncReadyCallback)onStopDriveFinished, new QPointer<MountOperation>(this));
+                g_object_unref(drv);
+                return;
+            }
+            else if(g_drive_can_eject(drv)) {
+                g_drive_eject_with_operation(drv, G_MOUNT_UNMOUNT_NONE, op, cancellable_, (GAsyncReadyCallback)onEjectDriveFinished, new QPointer<MountOperation>(this));
+                g_object_unref(drv);
+                return;
+            }
+            g_object_unref(drv);
+        }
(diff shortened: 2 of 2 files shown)


isf63:
Will optical drives remain visible in `computer:///` with no medium? The "Eject" button there has the ability to open the tray, regardless of being mounted.

tsujan:
Yes, optical drives are different. But I'd appreciate it if you could test with optical drives; I have none here.

tsujan:
Oh, if you test it, don't eject from inside `computer:///`; *use the side-pane for ejection*.

The behavior of `computer:///` is like before because, apparently, GLib doesn't provide a way of finding the volume from a `GFile`, let alone the drive. I even wanted to remove the "Eject" menu item from `computer:///` but thought it's useful for media devices. This seems to me like a drawback in GIO.

isf63:
Testing, my USB thumbdrive doesn't have a "power on" light so it's identical as before. The "root" (`/dev/sdd` - not a partition) item along with its name still shows up in `computer:///` after unmounting the partition(s). `/dev/sr0` eject from `computer:///` doesn't disappear the item.

There's a possibility I'm not on your branch though. Transitioning to Ubuntu and have not yet re-learned everything.

tsujan:
> /dev/sr0 eject from computer:/// doesn't disappear the item.

See my previous comment: it shouldn't disappear from inside `computer:///`. Please use the side-pane for ejecting.

Moreover, I may have found a way of doing it from inside `computer:///` too, by using `g_file_find_enclosing_mount`, which I missed before. Will experiment with it…

tsujan:
OK, for whatever reason, `g_file_find_enclosing_mount` returns NULL inside `computer:///` — most probably, that's a drawback of GVFS. So, this works only by ejecting from the side-pane.

tsujan:
I tried it with another USB stick that magically auto-connected after being ejected — its icon disappeared and reappeared. The same happened with Thunar. I didn't know before that such a behavior was possible.

Anyway, pcmanfm-qt's ejection from its side-pane is exactly like Thunar's "safe removal" now; the difference is that pcmanfm-qt doesn't pretend that there's anything called "safe removal" (a phrase taken from MS Windows) apart from ejection. So, IMO, this is a good change.

tsujan:
I was lucky to be able to borrow 4 USB sticks, each of which showed a different behavior with this patch. With some, the item under Computer disappeared on ejecting from the side-pane, with others it didn't; and there was one USB stick that auto-connected after successful ejections. But the same thing always happened in Thunar.

Anyhow, this is the new behavior on ejecting from the side-pane, put in English:

 1. The app first tries to stop the drive (like Thunar and Nautilus);
 2. If not successful, it tries to eject the drive (like Thunar);
 3. If not successful, it uses its usual method, which is successful under all normal circumstances.

The difference is that, if step 1 or 2 succeeds, the stick will also be powered off. That didn't happen before, when only step 3 was used. So, IMO, this change is a minor improvement.

When or why 1 or 2 succeeds, I don't know, and I doubt that GLib devs know either. Someone showed me the different behaviors of his USB stick on "safe removal" (= drive stopping/ejection) under Windows7 and Windows11. It's like tossing dice…

stefonarch:
Rebased and tested, I can't see any diff, my luks disk for backups doesn't stop spinning when unmounted and the light is on, also on one other stick I've got.

tsujan:
> my luks disk for backups doesn't stop spinning

As I explained above, this is for USB sticks, not disks. Disks can't be ejected.

stefonarch:
It's an usb disk, so it can't power off?

tsujan:
> It's an usb disk

A USB stick (aka. USB flash drive, thumb drive. etc.) is a small, old-fashioned device, used in per-historic times, but still in use. It's usually formatted in FAT32. For whatever reason, it has an "Eject" option, like optical media.

What you have is an external hard drive that's connected to a USB port. It can't be "ejected" in this sense; it can only be unmounted.

Chiitoo:
I don't have thunar or udisks2 installed to test right now, but looking around a bit, perhaps `udisksctl power-off` is being used?

Or something similar maybe.

tsujan:
> perhaps udisksctl power-off is being used?

We use only GLib/GIO. GLib knows how to do it, and we trust it for file management.

stefonarch:
Just tested, ` udisksctl power-off -b /dev/sdb1` works.

stefonarch:
Just noticed that if I expell the usb disk using mount plugin it stops spinning.

tsujan:
Does your disk provide an "Eject" option when you right click its item in the side-pane?

stefonarch:
No, only "unmount".

tsujan:
Thanks! I needed the answer.

Please bear with me. I'm testing new ideas…

tsujan:
Superseded by https://github.com/lxqt/libfm-qt/pull/1050

PR #1050, opened 2025-04-19
Added "Safely Remove" to side-pane context menu when possible

tsujan:
This new item appears only if (`GIO` announces that) the drive is stoppable.

Also, when a volume can be ejected, the code first tries to stop or eject the *drive* instead, and only if unsuccessful, it tries to eject the volume itself.

Closes https://github.com/lxqt/libfm-qt/issues/1048

### DIFF
--- a/src/mountoperation.cpp
+++ b/src/mountoperation.cpp
@@ -207,6 +207,24 @@ void MountOperation::onEjectVolumeFinished(GVolume* volume, GAsyncResult* res, Q
     delete pThis;
 }
 
+void MountOperation::onStopDriveFinished(GDrive* drive, GAsyncResult* res, QPointer<MountOperation>* pThis) {
+    if(*pThis) {
+        GError* error = nullptr;
+        g_drive_stop_finish(drive, res, &error);
+        (*pThis)->handleFinish(error);
+    }
+    delete pThis;
+}
+
+void MountOperation::onEjectDriveFinished(GDrive* drive, GAsyncResult* res, QPointer<MountOperation>* pThis) {
+    if(*pThis) {
+        GError* error = nullptr;
+        g_drive_eject_with_operation_finish(drive, res, &error);
+        (*pThis)->handleFinish(error);
+    }
+    delete pThis;
+}
+
 void MountOperation::onEjectFileFinished(GFile* file, GAsyncResult* res, QPointer<MountOperation>* pThis) {
     if(*pThis) {
         GError* error = nullptr;
--- a/src/mountoperation.h
+++ b/src/mountoperation.h
@@ -93,6 +93,20 @@ public:
             prepareUnmount(mnt);
             g_object_unref(mnt);
         }
+        // first check if the drive can be stopped or ejected.
+        if(GDrive* drv = g_volume_get_drive(volume)) {
+            if(g_drive_can_stop(drv)) {
+                g_drive_stop(drv, G_MOUNT_UNMOUNT_NONE, op, cancellable_, (GAsyncReadyCallback)onStopDriveFinished, new QPointer<MountOperation>(this));
+                g_object_unref(drv);
+                return;
+            }
+            else if(g_drive_can_eject(drv)) {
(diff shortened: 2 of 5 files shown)


stefonarch:
At a first glance it looks perfect. With a normal usb stick I've "unmount" and "eject", with the usb disk "unmount" and "Safely remove". But actually also unmounting  is safe, so not sure about the wording.

isf63:
From the Microsoft website (not that Microsoft Windows is the authority, but):

>In the system tray area of the taskbar, right-click on the Safely Remove Hardware and Eject Media icon. If the Safely Remove Hardware and Eject Media icon isn't visible, select Show hidden icons first, and then right-click the Safely Remove Hardware and Eject Media icon.

tsujan:
Unmounting seems safe to me too, but apparently some other FMs as well as MS Windows use "Safely Remove" for stopping the drive — as you surely noticed the difference.

tsujan:
In the case of an external drive with multiple partitions, whether they're encrypted or not, if you use "Unmount" or just click the Eject button, the corresponding mounted partition will be unmounted but its item won't disappear. On ther hand, if you select "Safely Remove", provided that there's no file transference, `GIO` will take care of stopping the drive, such that the partition items will disappear and lxqt-panel's Removable Media shows notifications about the removal.

In short, with "Safely Remove", you should reinsert the drive if you want to use it again, or you could just disconnect it (its main purpose).

stefonarch:
I have to admit I never used the right click menu, always used the eject icon. Using that the drive remains visible, the stick not.

stefonarch (approved):

stefonarch:
Btw some users still would like the drives/sticks/cd showing up on the desktop like trash and network (which looks obsolete to me).

tsujan:
> Btw some users still would like the drives/sticks/cd showing up on the desktop …

I have no plan for it, because it requires lots of dirty codes. Most importantly, I don't want to make the desktop code more complex that it is now.

Remember the old Nautilus that drew the desktop in GNOME2? Its code became so unmaintainable that they removed it altogether in GNOME3. I absolutely don't want this for pcmanfm-qt.

tsujan:
> … used the eject icon. Using that the drive remains visible, the stick not.

I found an old USB stick that also remained visible after ejection — even lxqt-panel's Removable Media couldn't make it disappear, unless it was physically ejected. The same thing happened in Thunar and Dolphin with it. But my own stick disappears on soft ejecting.

stefonarch:
> > Btw some users still would like the drives/sticks/cd showing up on the desktop …
> 
> I have no plan for it, because it requires lots of dirty codes. Most importantly, I don't want to make the desktop code more complex that it is now.
> 
> Remember the old Nautilus that drew the desktop in GNOME2? Its code became so unmaintainable that they removed it altogether in GNOME3. I absolutely don't want this for pcmanfm-qt.

I've no problems with that. IMO the desktop isn't for launchers and similar, as it's buried under windows usually (stacking) or on an empty workspace (tiling) so the panel is a much better place to eject drives.

tsujan:
> so the panel is a much better place to eject drives.

Personally, I use only pcmanfm-qt's side-pane for drive jobs, but I like lxqt-panel's Removable Media because of its nice notifications.

stefonarch:
Just now after wake from suspend I had an issue, botu stick and drive are remounted on wake as usually, but I couldn't eject the stick anymore "no dbus interface" or similar.

tsujan:
I don't see anything in this commit that can cause that. Fortunately, we have lxqt-panel. What happens if you eject the stick from lxqt-panel?

stefonarch:
It happened once only, and the issue was the stick apparently. As I said I got the same message from both panel and sidepane.

tsujan:
I think I encountered it once or twice months ago. when trying to eject something that was already ejected. Why something ejected may appear to be ejectable, I don't know, but since it happened in both lxqt-panel and pcmanfm-qt, the cause should be more fundamental.

tsujan:
BTW, it needs a translation update for "Safely Remove".

PR #1055, opened 2025-05-06
Used `--extract-here` for "Extract Here" with lxqt-archiver

tsujan:
Also,
 * Updated Arqiver's mimetypes list; and
 * Silenced a new compilation warning about implicit capturing of `this` with `=`.


tsujan:
@stefonarch 
Could you check this please?

stefonarch:
Didn't follow really, what exactly should I check for?

tsujan:
"Extract Here" should work like a charm; you can test with just a few archives that worked before.

PR #1057, opened 2025-05-17
Add and make message more descriptive for emptying trash

isf63:
Previously the wrong message was prompted - "Do you want to delete the selected file(s)?"

Extra `FileOperation::` function was warranted because `FileOperation::deleteFiles` shouldn't be passed a message string.

Todo: Use the function in PCManFM-Qt desktop; it's coded separately there.

### DIFF
--- a/src/fileoperation.cpp
+++ b/src/fileoperation.cpp
@@ -416,6 +416,28 @@ FileOperation* FileOperation::deleteFiles(Fm::FilePathList srcFiles, bool prompt
 }
 
 //static
+FileOperation* FileOperation::emptyTrashFiles(bool prompt, QWidget* parent) {
+
+    Fm::FilePathList srcFiles;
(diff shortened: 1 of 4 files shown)


tsujan:
I'm afraid adding a new (and, *redundant*) file operation just for showing a separate message isn't good.

Emptying Trash is an instance of deleting the files, which is handled by `FileOperation::deleteFiles`. The code is compact and uncluttered, and the message doesn't have any problem.

isf63:
Is there any way to show a different string with no other function?

tsujan:
Since it's about the general concept of deletion, it makes no difference how it's done.

isf63:
Would a parameter `bool all = false` in `FileOperation::trashFiles` make any sense?

tsujan:
It's possible to add a new argument but, really, the code is clean and the message is correct; changing it isn't worth it.

PR #1058, opened 2025-05-19
Small UX improvement to emptying Trash inside folder

isf63:
Previously the menu action for emptying, inside the `trash:///` folder, disappeared when there was nothing to empty.

This makes the menu consistent with Desktop and 'Places' menus, which instead greyed the action. It also in my opinion is better UX.

**To test**
1. PCManFM-Qt → Places → Trash → Right click
2. PCManFM-Qt Desktop → Trash → Right click
3. PCManFM-Qt `trash:///` → Right click

### DIFF
--- a/src/foldermenu.cpp
+++ b/src/foldermenu.cpp
@@ -54,16 +54,15 @@ FolderMenu::FolderMenu(FolderView* view, QWidget* parent):
     }
     if(insideTrash) {
         if(auto folder = view_->folder()) {
-            if(!folder->isEmpty()) {
-                auto trashAction = new QAction(tr("Empty Trash"), this);
-                addAction(trashAction);
-                connect(trashAction, &QAction::triggered, []() {
-                    Fm::FilePathList files;
-                    files.push_back(Fm::FilePath::fromUri("trash:///"));
-                    Fm::FileOperation::deleteFiles(std::move(files), true);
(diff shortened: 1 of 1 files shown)


tsujan:
I made it so *intentionally* :)

> It also in my opinion is better UX.

I'm afraid I disagree.

isf63:
Why intentionally?

It's much better IMO if menus are static, and the disabled cue is everywhere but folder view.

tsujan:
> Why intentionally?

Didn't want a redundant item there. The context menu may already be populated by custom actions of all kinds. I know yours isn't, but it's a possibility. The less redundant items, the better.

> It's much better IMO…

Much?! Sometimes I wonder why you exaggerate :) Although I like it when you pay attention to details.

isf63:
Ha, it is a minor detail, there is not _much_ about anything.

Personally I think having a static menu (important items are the same, only enabled/disabled) + consistency with other locations, is worth one extra menu item in the case of an empty Trash folder.

isf63:
The point I was trying to get across for this small detail is: the pros, as described above, outweigh the one con - that someone may have a large number of custom actions for `inode/directory` and now they have one more menu action. The change may not be completely beneficial in _all use cases_, but I'd reckon the _median use case_ it is.

PR #1059, opened 2025-05-20
Workaround for GLib's problem with moving directory over directory

tsujan:
GLib doesn't allow moving a directory over another one with the same name. As a workaround, we copy and delete in this case.

### DIFF
--- a/src/core/filetransferjob.cpp
+++ b/src/core/filetransferjob.cpp
@@ -447,7 +447,13 @@ bool FileTransferJob::moveFile(const FilePath &srcPath, const GFileInfoPtr &srcI
     auto src_fs = g_file_info_get_attribute_string(srcInfo.get(), "id::filesystem");
     auto dest_fs = g_file_info_get_attribute_string(destDirInfo.get(), "id::filesystem");
     bool ret;
-    if(src_fs && dest_fs && (strcmp(src_fs, dest_fs) == 0 || g_str_has_prefix(src_fs, "trash"))) {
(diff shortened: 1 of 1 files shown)


tsujan:
@stefonarch 
To test it, make two directories with the same name *in the same partition* and try to move (not just copy) one over the other.

Previously, an error message was shown by GLib. No message was shown when directories were on different filesystems only because we did a copy+delete in that case.

stefonarch:
GTM
Content is merged, I'd never expect it to be honest

https://github.com/user-attachments/assets/b3ff4aac-d261-4844-bfee-e8b845e68566

tsujan:
>  Content is merged, I'd never expect it to be honest

I think many users don't do it.

Thanks.

isf63:
I've run into this limitation several times when making backups. The workarounds were difficult, so this is a welcome fix.

tsujan:
@isf63 
Your PR tests or reviews would be very appreciated in all cases. The more tests before merging, the better. Of course, tests that are done after merging are important too.

tsujan:
BTW, the original report was https://github.com/lxqt/pcmanfm-qt/issues/1816. It was closed because GLib was responsible. However, the workaround was small.

isf63:
Tested and works, also works across drives/partitions.

luebking:
Symptoms and patch and versions are simply too much on the nose …
Let's see whether there's a result to the revert attempt.

PR #1075, opened 2025-10-02
Make dialogs as modal only for the window

City-busz:
Without this, the dialogs block all opened windows of the application, not just the parent window.


tsujan:
→ https://github.com/lxqt/pcmanfm-qt/pull/2101#issuecomment-3362720632

PR #1076, opened 2025-10-02
Make dialogs as modal only for the window

City-busz:
Without this, the dialogs block all opened windows of the application, not just the parent window.


tsujan:
Please don't do this (→ https://github.com/lxqt/pcmanfm-qt/pull/2102#issuecomment-3363316731).

PR #1077, opened 2025-10-03
Fix 'Erase files on removable media instead of trash can creation' deleting files on internal drives.

Rabcor:
Fixes: https://github.com/lxqt/pcmanfm-qt/issues/2056

Makes it so that the 'Erase files on removable media instead of trash can creation' option only applies to actual removable media and not internal drives.

### DIFF
--- a/src/core/trashjob.cpp
+++ b/src/core/trashjob.cpp
@@ -31,7 +31,7 @@ void TrashJob::exec() {
             if(fm_config->no_usb_trash) {
                 GMountPtr mnt{g_file_find_enclosing_mount(gf.get(), nullptr, nullptr), false};
                 if(mnt) {
-                    ret = g_mount_can_unmount(mnt.get()); /* TRUE if it's removable media */
+                    ret = g_mount_can_eject(mnt.get()); /* TRUE if it's removable media */
                     if(ret) {
                         unsupportedFiles_.push_back(path);
                         break;  // don't trash the file


tsujan:
Neither reliable nor backward compatible. External hard drives aren't ejectable

Rabcor:
> Neither reliable nor backward compatible. External hard drives aren't ejectable

It was the only simple solution I could find, every drive that isn't the root drive will always be read by pcmanfm-qt as an external drive, preventing that seems like extremely basic functionality to me that any file manager should have, unlike what you said on that issue this is in fact an issue with pcmanfm-qt, not gvfs, I hope you will eventually get around to fixing it, and perhaps reopen that issue to reflect this fact.

tomty89:
@tsujan The question is, what exactly is option supposed to cover? "Anything USB"? (As internally the value is stored as `no_usb_trash`.)

Or, if the interface isn't the point here, what is "removable media" supposed to cover / refer to?

Either way, is there anything concrete that makes `g_mount_can_unmount` a better or even, a sensical choice? I mean, it doesn't seem to be supposed to have anything to do with USB / removable media anyway. (From what I can see, it pretty much returns true for everything.)

P.S. I'm just passing by. It's totally fine by me if this is like "I know it's hasn't made any sense from day one but I'd like to keep it untouched", but I'm really not seeing how avoiding this change could be a better idea for like, anyone.

tsujan:
See https://github.com/lxqt/libfm-qt/issues/643 for a similar (but not identical) conundrum.

As I mentioned in various places, this should be solved by GIO/GVFS. Until then, we have to stick with the most logical choice available to us, although it isn't ideal.

tsujan:
> what is it even assumed / expected to mean for you?

It's simple: removable ≈ un-mountable. Ejection is an additional job doable only for DVD ROMs (irrelevant) and USB sticks.

> Assuming you were the one that picked it.

The original author chose it. I found nothing better and kept it.

> what / how do you think that it should be "solved"?

I don't have a better idea, perhaps like GNOME devs. I didn't mean it's a problem to me; used the word "solve" to acknowledge that it was to some.

tomty89:
> removable ≈ un-mountable

I assume that by unmountable you mean like it is not supposed to be a part of the system (i.e. "data drive"). In that case I guess the choice does make sense, at least as per the actual behavior.

From what I can see, everything is `can_unmount=1`, except for those that are mounted under somewhere under `/media` or `/run/media` -- I don't know whether it is expected but mounts elsewhere simply won't get enumerated by gio/gvfs, apparently (which is why a "drive" would disappear from the drive list on the file manager once it is mounted elsewhere).

So at least practically speaking, if you want to tell gvfs/gio (and in turn the file manager) that the drive is "non-removable" (as per your definition), what you should do is that avoid mounting it under `/media` and `/run/media` (which implies you need to have an fstab entry for it instead of "easy mounting" it through udisks2) -- this is what I told @Rabcor on SuperUser too (apparently what really annoys him is that then the drive will be gone from the "drive list" -- but the list was pretty much meant to show "removable drives" from day one).

tomty89:
I see. So basically "I don't care".

Because to be frank, that makes no sense anyway, because whether a filesystem can be *actually* unmounted is "time-specific" (like whether it is in use, period).

And certainly it can't possibly be sensical to use to relate to "removable media", unless you are using it for a contextual meaning you desire for which its practical behavior *happen to" match.

Rabcor:
> Because to be frank, that makes no sense anyway, because whether a filesystem can be _actually_ unmounted is "time-specific" (like whether it is in use at the moment, period). (Of course it can be about permission as well.)

The user already can't unmount the drive actually, it's only unmountable by root.

PR #1092, opened 2026-02-10
Streamline file-props file count tr

isf63:
Use numeric count in all instances (we cannot assume all languages follow the same English convention,) plus they weren't capitalized. Also simplify the code.

### DIFF
--- a/src/filepropsdialog.cpp
+++ b/src/filepropsdialog.cpp
@@ -449,11 +449,10 @@ void FilePropsDialog::onFileSizeTimerTimeout() {
         if(ui->contentsLabel->isVisible()) {
             unsigned int fc =  totalSizeJob->fileCount(); // the directory is included
             if (fc <= 1)
-                str = tr("no file");
-            else if (fc == 2)
-                str = tr("one file");
+                fc = 0;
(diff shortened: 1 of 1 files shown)


tsujan:
> we cannot assume …

No assumption is made. Let's not waste a second on discussing capitalization either.

"no file" looks better than "0 file" (although with this patch it would become "0 files").

isf63:
IMO "no file" is odd as an English speaker, it should be "Zero files". Plus this sidesteps the capitalization point.

isf63:
Maybe just change, now that I think about translatability:
- "no file" -> "Zero files"
- "one file" -> "One file"

tsujan:
> I think about translatability:

There has been no problem with that either. "Zero files" (or "Zero file") is terrible, BTW ;)

PLEASE ignore capitalization!! It's good as it is.

tsujan:
> as an English speaker

And I wonder how you can be an English speaker and find "Contains no file" odd!

isf63:
Closing, because it's a fix without much of a problem. The specifics of the translations can be taken up (much) later.

PR #1100, opened 2026-04-15
Replaced invalidateFilter

stefonarch:
Fixes some of the warnings here https://github.com/lxqt/libfm-qt/issues/1099

The vfs-menu.c part is done with help from LLMs, so this needs a good review.

### DIFF
--- a/src/placesview.cpp
+++ b/src/placesview.cpp
@@ -59,7 +59,8 @@ void PlacesProxyModel::restoreHiddenItems(const QSet<QString>& items) {
             ++i;
         }
         hiddenItemsRestored_ = true;
-        invalidateFilter();
+        beginFilterChange();
+        endFilterChange();
     }
 }
 
@@ -72,12 +73,14 @@ void PlacesProxyModel::setHidden(const QString& str, bool hide) {
     else {
         hidden_.remove(str);
     }
-    invalidateFilter();
+    beginFilterChange();
+    endFilterChange();
 }
 
 void PlacesProxyModel::showAll(bool show) {
     showAll_ = show;
-    invalidateFilter();
+    beginFilterChange();
+    endFilterChange();
 }
 
 bool PlacesProxyModel::filterAcceptsRow(int source_row, const QModelIndex& source_parent) const {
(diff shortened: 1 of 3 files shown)


isf63:
There was a cleaner way to silence the `const` warnings, also in multiple files: https://github.com/lxqt/libfm-qt/pull/1103/

stefonarch:
Removed the  warnings already fixed by https://github.com/lxqt/libfm-qt/pull/1104.

tsujan:
It's good but needs version check (6.10). Adding version checks for 6.10 isn't a good idea; instead we should bump the min. required version of Qt to 6.10 everywhere and stop supporting very old distros in git and the next release. So, please wait for now …

tsujan (changes requested):
Oops! There's a problem that I missed: change(s) should be done between `beginFilterChange();` and `endFilterChange();`. So, this needs fixing.

src/placesview.cpp
```diff
@@ -59,7 +59,8 @@ void PlacesProxyModel::restoreHiddenItems(const QSet<QString>& items) {
             ++i;
         }
         hiddenItemsRestored_ = true;
-        invalidateFilter();
+        beginFilterChange();
+        endFilterChange();
```
tsujan, line 63: Changes should be done between `beginFilterChange()` and `endFilterChange()`, like this:

```c++
    if(!hiddenItemsRestored_ && !items.isEmpty()) {
        beginFilterChange();
        hidden_.clear();
        ....
        endFilterChange();
```

And also in the other cases.

stefonarch: Done hopefully right.

src/placesview.cpp
tsujan, line 67: `beginFilterChange();` is missing here:

```c++
    beginFilterChange();
    if(hide) {
```

src/placesview.cpp
```diff
@@ -59,25 +60,27 @@ void PlacesProxyModel::restoreHiddenItems(const QSet<QString>& items) {
...
         }
         hiddenItemsRestored_ = true;
-        invalidateFilter();
+        endFilterChange();
     }
 }
 
 void PlacesProxyModel::setHidden(const QString& str, bool hide) {
     if(hide) {
+    beginFilterChange();
```
tsujan, line 68: `beginFilterChange();` should be before `if(hide) {`.

stefonarch: shouldn't do this work when tired as now...

stefonarch: Should we rise Qt version in this PR too?

tsujan (approved):
Now it's OK, but please don't merge it until bumping of Qt version is decided and done.

PR #1101, opened 2026-04-17
Strange sorting on cyrillic locale

ded-maped:
On a system with cyrillic locale(i tested with ukrainian and russian ones) file names sorting is very strange.
It places cyrillic names before latin ones, like
__file
123
567
абв
где
abc
defgh
I cannot remember any program with such sorting, and it is couter intuitive because any non-latin characters have higher codes.
The bug is very annoying

### DIFF
--- a/src/proxyfoldermodel.cpp
+++ b/src/proxyfoldermodel.cpp
@@ -40,6 +40,7 @@ ProxyFolderModel::ProxyFolderModel(QObject* parent):
     setSortCaseSensitivity(Qt::CaseInsensitive);
 
     collator_.setNumericMode(true);
+    collator_.setLocale(QLocale::AnyLanguage);
 }
 
 ProxyFolderModel::~ProxyFolderModel() {


isf63:
The default Qt locale (before calling any `setLocale`) is the OS/system locale. `AnyLanguage` from what I can gather is a unspecified locale and Qt will resolve to `C` or similar.

Have you tried using `lxqt-config-locale` to set a system locale and does that fix your problem?

ded-maped:
System locale was set during OS installation, it looks fine in lxqt-config-locale and there is no other issues with it. What should i set and what's that for? Just to rewrite possibly corrupted settings?

Maybe QLocale::C is a better option, idk

ded-maped:
With QLocale::C NumericMode feature doesn't work. Looks like QLocale::AnyLanguage is the best option

tsujan:
→ https://lxqt-project.org/wiki/SettingLocale.html

PR #1103, opened 2026-04-20
Silence const warnings

isf63:


stefonarch:
That's @tsujan sacred field ;)

tsujan:
→ https://github.com/lxqt/libfm-qt/pull/1104

PR #1114, opened 2026-08-29
Enhance search experience with incremental file listing and be able to stop a search job

ahsand97:
## 🔍 Incremental search results + stoppable search jobs

`search://` listings were collected entirely in memory and only handed to the UI once the whole recursive enumeration finished — so a slow search looked completely frozen/empty the whole time it ran, and there was no way to cancel one in progress without destroying the `Folder` object outright (e.g. by closing the tab). This PR wires up the incremental-loading logic that already existed as dead/unfinished code, and adds a real stop API. It's the companion library change for the search-UX work in `pcmanfm-qt`.

- `DirListJob` now emits `filesFound()` in batches of 32 (⚙️ tunable — flagged inline as a lazy tradeoff between round-trip overhead and result latency) while enumerating a `search://` job, instead of silently collecting everything into one list and dumping it into the view only at the very end.
- 🔗 **Wired up `Folder`** — new `Folder::onDirListFilesFound()` slot forwards each batch straight into the existing `filesAdded` signal, over the same `Qt::BlockingQueuedConnection` pattern this file already uses for `error` (the job runs on its own thread via `Job::runAsync()`, so this keeps `Folder::files_` mutations serialized onto the main thread). `FolderModel` already listens to `filesAdded` unconditionally, so it picks up rows **live** with zero downstream changes needed.
- 🛑 **New `Folder::stopLoading()`** — cancels the in-flight `DirListJob` in place, without restarting it (unlike `reload()`, which cancels and immediately starts a fresh job). Needed no further plumbing: because results stream in as they're found rather than all at once, cancelling mid-search just stops it wherever it is, and the existing cancelled-job branch in `onDirListFinished()` already settles the folder into a correct "loaded" state with whatever was found so far.
- **Bug fixes**:
  - `emit_files_found` was never initialized in `DirListJob`'s constructor — reading it (via `incremental()`, and in `exec()`) was undefined behavior.
  - `DirListJob::setIncremental()` was declared in the header but never defined anywhere — calling it would have been a link error.
- **Deleted ~30 lines of dead `#if 0` GLib/C-era code** (`emit_found_files`, `queue_add_file`, `fm_dir_list_job_add_found_file`, and the `#if 0`'d incremental-connect block in `reallyReload()`) that was never compiled — this is exactly the functionality the new Qt-native `filesFound`/`onDirListFilesFound` implementation replaces.

Companion PR in `pcmanfm-qt` https://github.com/lxqt/pcmanfm-qt/pull/2190

### DIFF
--- a/src/core/dirlistjob.cpp
+++ b/src/core/dirlistjob.cpp
@@ -7,7 +7,11 @@
 namespace Fm {
 
 DirListJob::DirListJob(const FilePath& path, Flags _flags):
-    dir_path{path}, flags{_flags} {
+    dir_path{path}, flags{_flags}, emit_files_found{false} {
+}
+
+void DirListJob::setIncremental(bool set) {
+    emit_files_found = set;
 }
 
 void DirListJob::exec() {
@@ -79,6 +83,10 @@ _retry:
     }
 
     FileInfoList foundFiles;
+    FileInfoList incrementalBatch;
+    // batch size trades round-trip overhead against result latency;
+    // revisit if very large recursive searches feel laggy.
+    constexpr std::size_t incrementalBatchSize = 32;
     /* check if FS is R/O and set attr. into inf */
     // FIXME:  _fm_file_info_job_update_fs_readonly(gf, inf, nullptr, nullptr);
     err.reset();
@@ -126,7 +134,11 @@ _retry:
 #endif
                 auto fileInfo = std::make_shared<FileInfo>(inf, FilePath(), realParentPath);
                 if(emit_files_found) {
-                    // Q_EMIT filesFound();
+                    incrementalBatch.push_back(fileInfo);
+                    if(incrementalBatch.size() >= incrementalBatchSize) {
+                        Q_EMIT filesFound(incrementalBatch);
+                        incrementalBatch.clear();
+                    }
                 }
 
                 foundFiles.push_back(std::move(fileInfo));
@@ -143,6 +155,10 @@ _retry:
                 break;
             }
         }
+        if(emit_files_found && !incrementalBatch.empty()) {
+            Q_EMIT filesFound(incrementalBatch);
+            incrementalBatch.clear();
+        }
         err.reset();
         g_file_enumerator_close(enu.get(), cancellable().get(), &err);
     }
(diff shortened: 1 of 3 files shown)


tsujan:
Thanks!

It's a good idea and, at a glance, the implementation looks good. It might take a while before I find the time to examine all possible aspects of it.

As for your implementation in pcmanfm-qt, I have some doubts, but we'll discuss it later.

tsujan:
There's a fundamental issue: The detailed list view isn't compatible with an incremental search. Since setting of its columns layout is queued (as it should be), the user may see an empty folder several times before stopping the search.

tsujan:
The incremental loading should be decided inside the core code, while the mode setting belongs to the GUI code. So, the detailed list mode can't be excluded without very untidy changes to the core code.

On the other hand, seeing an empty folder defeats the purpose of incremental loading. We can't ignore this issue.

tsujan:
@ahsand97 
I found a clean and short way to rescue both PRs. Can I add commits to your PRs, or should I make separate PRs based on them? I ask because, in LXQt, no dev should add commits to another dev's PR without his permission, and it's quite understandable if the original dev prefers to keep his PR as it is (although commits can be reverted).

This is the safe way:

 1. Don't limit the incremental listing to `search://` in libfm-qt; make it optional for any folder, with a non-incremental listing by default.
 2. Use incremental listing with `search://` in pcmanfm-qt, but exclude the detailed list view mode.

A third step will make incremental listing with `search://` optional too. The reason is that incremental listing takes more CPU time — sometimes much more.

ahsand97:
> Can I add commits to your PRs

Yep, no problem, team work ;).

tsujan:
> Yep, no problem, team work ;)

So you agree with the general plan. Will do it tonight (I think our time zones are very different).

ahsand97:
> So you agree with the general plan.

Yep :)

tsujan:
Done. Please check both of them, although they worked fine here.

The third step (in pcmanfm-qt) is yours. Please make the incremental search optional, with non-incremental search as the default. You might also want to add a short tooltip to its option, saying something like "Does not work with the detailed list mode". The best place for the new option is Preferences → Advanced → Search.

ahsand97:
> Done. Please check both of them, although they worked fine here.
> 
> The third step (in pcmanfm-qt) is yours. Please make the incremental search optional, with non-incremental search as the default. You might also want to add a short tooltip to its option, saying something like "Does not work with the detailed list mode". The best place for the new option is Preferences → Advanced → Search.

Done

tsujan:
It looks good:

search

tsujan:
Everyone is welcome to test the PRs.

@ahsand97,  you cheated by involving me in this ;) Now I agree with the change, provided that no issue pops up.

## WARNING

*Since the ABI is changed, all apps that are based on libfm-qt should be recompiled.*

@stefonarch, in case I'm on Mars again when LXQt 2.5.0 is going to be released, please bump libfm-qt's ABI and "so" version to 18. Hopefully I'll be on Earth.

stefonarch:
It looks good here.

tsujan:
With the current code, we can have incremental listing everywhere and cheat like Thunar does, i.e., pretend that we're faster than GLib (= impossible). But over my dead body ;)

EDIT:
Just for the reference: https://github.com/lxqt/pcmanfm-qt/issues/462#issuecomment-753602825. 

There I said:

> Thunar "cheats" by quickly showing items and only then, doing intense file reading for a long time …

That's incremental listing.

tsujan (approved):

ahsand97:
> you cheated by involving me in this ;)

team work! hahahaha

tsujan:
I meant I saw how good your implementation was as soon as I started to generalize it. Before that, I only "felt" that it should be correct.

PR #1117, opened 2026-09-06
Functionallity to show volumes on desktop in pcmanfm-qt

### LINKED ISSUE #1192: [Feature request] Make it possible to show volumes on the desktop
##### Expected Behavior

There should be a way to show volumes/partitions on the `--desktop` view, like they already show up under "Devices" in the side panel of the the file manager.

Most desktops allow this. E.g., Xfce:

image

##### Current Behavior

Volumes/partitions cannot be shown in the `--desktop` view.

ahsand97:
Well, I was digging into the issues of `pcmanfm-qt` and this one seemed interesting to me:          
                                                                                                      
🔗 https://github.com/lxqt/pcmanfm-qt/issues/1192                                                     
                                                                                                      
And digging more into the conversation 🕵️ as mentioned in https://github.com/lxqt/pcmanfm-qt/issues/2130 usable `.desktop` files from `computer:///` was not a good approach, so I decided to dig more into the code and found out that the bug is **not** in `.desktop` file generation deciding `Type=Link` vs `Type=Application`. The generation code is generic and correct ✅; the "bug" is in the launch path, in a stale native-only check inside `FileInfo::isExecutableType()` — a stale `target.isNative()` check that returns `false` for any shortcut pointing at a non-local URI — which is every GVfs virtual scheme (`computer://`, `network://`, `trash://`, `mtp://`, etc.).                                
                                                                                                      
The launcher already has a scheme whitelist that handles exactly this — `handleShortcut()` in `basicfilelauncher.cpp` already special-cases `file`, `trash`, `network`, `computer`, `menu` and knows how to open them correctly 👌. It's just unreachable for non-native targets because `isExecutableType()` bails out before ever calling it. That's why you get `"Invalid desktop entry file"` instead of it working.                                                                                             

Hopefully with this implementation that just requires a very minimal change in `libfm-qt` solves it, also I added logic to trust the generated `.desktop` file, just like the other generated `.desktop` files already work.

Closes https://github.com/lxqt/pcmanfm-qt/issues/1192 and https://github.com/lxqt/pcmanfm-qt/issues/2130

Edit: Also I believe this would need a PR in `pcmanfm-qt`, like a configuration in desktop preferences something like "Show volume icons on connection" or something like that to automatically create the `.desktop` files on the desktop. But before proceeding to that I wanted to confirm the core functionallity in `libfm-qt` would be valid.

### DIFF
--- a/src/core/fileinfo.cpp
+++ b/src/core/fileinfo.cpp
@@ -409,6 +409,15 @@ bool FileInfo::isExecutableType() const {
                         return true;
                     }
                 }
+                else if(target.hasUriScheme("trash") || target.hasUriScheme("network")
+                        || target.hasUriScheme("computer")) {
+                    /* these virtual schemes are known-safe launch targets;
+                       see BasicFileLauncher::handleShortcut(). "menu" is deliberately
+                       excluded here: it's handled by its own special case in
+                       BasicFileLauncher::launchDesktopEntry(), which needs
+                       isExecutableType() to stay false for it. */
+                    return true;
+                }
             }
             else {
                 return true;
--- a/src/core/filetransferjob.cpp
+++ b/src/core/filetransferjob.cpp
@@ -653,6 +653,13 @@ bool FileTransferJob::createShortcut(const FilePath &srcPath, const GFileInfoPtr
                 }
             } while(!isCancelled() && retry);
             ret = true;
+
+            // the user explicitly chose "Create Link Here" on this location; trust it
+            // like the desktop's own auto-generated shortcuts.
+            GFileInfoPtr trustInfo{g_file_info_new(), false};
+            g_file_info_set_attribute_string(trustInfo.get(), "metadata::trust", "true");
+            g_file_set_attributes_from_info(destPath.gfile().get(), trustInfo.get(),
+                                            G_FILE_QUERY_INFO_NONE, cancellable().get(), nullptr);
         }
     }
     return ret;
--- a/src/core/filetransferjob.cpp
+++ b/src/core/filetransferjob.cpp
@@ -653,13 +653,6 @@ bool FileTransferJob::createShortcut(const FilePath &srcPath, const GFileInfoPtr
                 }
             } while(!isCancelled() && retry);
             ret = true;
-
-            // the user explicitly chose "Create Link Here" on this location; trust it
-            // like the desktop's own auto-generated shortcuts.
-            GFileInfoPtr trustInfo{g_file_info_new(), false};
-            g_file_info_set_attribute_string(trustInfo.get(), "metadata::trust", "true");
-            g_file_set_attributes_from_info(destPath.gfile().get(), trustInfo.get(),
-                                            G_FILE_QUERY_INFO_NONE, cancellable().get(), nullptr);
         }
     }
     return ret;


src/core/filetransferjob.cpp
```diff
@@ -653,6 +653,13 @@ bool FileTransferJob::createShortcut(const FilePath &srcPath, const GFileInfoPtr
                 }
             } while(!isCancelled() && retry);
             ret = true;
+
+            // the user explicitly chose "Create Link Here" on this location; trust it
+            // like the desktop's own auto-generated shortcuts.
+            GFileInfoPtr trustInfo{g_file_info_new(), false};
+            g_file_info_set_attribute_string(trustInfo.get(), "metadata::trust", "true");
+            g_file_set_attributes_from_info(destPath.gfile().get(), trustInfo.get(),
+                                            G_FILE_QUERY_INFO_NONE, cancellable().get(), nullptr);
```
tsujan: > the user explicitly chose "Create Link Here" on this location; trust it

A desktop entry shouldn't be trusted so easily. We don't know what may happen to a non-native file.

> like the desktop's own auto-generated shortcuts.

There's no concept of Desktop on this level either (in contrast to desktop entry).

tsujan (changes requested):

src/core/filetransferjob.cpp
ahsand97: done!

src/core/fileinfo.cpp
```diff
@@ -409,6 +409,15 @@ bool FileInfo::isExecutableType() const {
...
                 }
+                else if(target.hasUriScheme("trash") || target.hasUriScheme("network")
+                        || target.hasUriScheme("computer")) {
+                    /* these virtual schemes are known-safe launch targets;
+                       see BasicFileLauncher::handleShortcut(). "menu" is deliberately
+                       excluded here: it's handled by its own special case in
+                       BasicFileLauncher::launchDesktopEntry(), which needs
+                       isExecutableType() to stay false for it. */
+                    return true;
+                }
```
tsujan, line 420: We don't execute non-native files. Why do you think a file inside `computer:///` or `network:///` is safe? `BasicFileLauncher::handleShortcut()` is a private method that gives the path.

ahsand97: You're right that I shouldn't lean on handleShortcut()'s whitelist as the safety argument. What about narrowing this to `trash` and `computer` only — both enumerate purely local system state, nothing insecure. Dropping `network` since that one can reflect names discovered on the LAN, which isn't something I can call safe with the same confidence.

tsujan: Please (re-)read https://github.com/lxqt/pcmanfm-qt/issues/1192#issuecomment-687646017. Shortcuts aren't a solution.

I left that report open because I knew that it could be done by making Desktop something more than an `Fm::FolderView`, but thought a "magic" could happen — which isn't impossible with coding ;) I emphasize this again: IMO, it's a bad idea to make Desktop's code more complicated than it is now, especially for a duplicate feature that can be found in the side-pane.

ahsand97: Oh okay I understand, so this PR wouldn't close https://github.com/lxqt/pcmanfm-qt/issues/1192, but it would close https://github.com/lxqt/pcmanfm-qt/issues/2130 then?,   drag-a-drive and it throws an error is a genuine bug (isn't it?), unrelated to whether volumes-on-desktop with .desktop files is a good feature.

tsujan: > but it would close https://github.com/lxqt/pcmanfm-qt/issues/2130 then?

It was already closed ;) `computer:///` is just one example for `Invalid desktop entry file`. Create a link by DND from a file inside a folder that's inside Trash to anywhere, and then click the created shortcut.

I've made a rule of thumb for myself: When dealing with fundamental codes, follow the logic as closely as possible, without being mislead by casual cases; otherwise, unforeseen things could happen.

EDIT: BTW, the above rule of thumb was my main reason to generalize your previous libfm-qt PR.

ahsand97: @tsujan Thank you so much for the context, closing this PR since it's not necessary anymore nor closes the mentioned issue. But since I'm a very passionate coder haha I'd like to keep the conversation there.

tsujan: > I'd like to keep the conversation there.

Discussing things with you has been proven to be productive. I enjoy it regardless of the final result.

ahsand97: > Discussing things with you has been proven to be productive. I enjoy it regardless of the final result.

So do I!. I tend to be very enthusiast with crazy ideas and sometimes you need more direction and understanding, so I'm always open to a discussion.

PR #1118, opened 2026-09-16
Show and hide file dialog immediately before showing it with delay

tsujan:
This makes it possible to get access to our file dialog immediately when needed. For example, such an access is needed for setting the dialog's layer on Wayland.

WARNING: This patch should be tested thoroughly, under both X11 and Wayland.

Without the delayed showing, the file dialog would be completely blocked and might block the window of the calling app too, in which case killing the app would be the only solution. So, the delayed showing is kept intact here. But I'm not sure about probable side effects of showing and hiding. Theoretically no side effect comes to my mind, but who knows? Hence the need for testing.


tsujan:
Superseded by https://github.com/lxqt/libfm-qt/pull/1119

PR #1119, opened 2026-09-17
remove bogus hack and just use the QPA API as was meant to

luebking:
And nobody gets to call this victory: I know that you're trying to trick me into installing a DE.
Doesn't count :P


tsujan (approved):
It is a victory.

PR #1122, opened 2026-09-20
Removed version check for Qt < 6.8

stefonarch:


tsujan:
Thanks, but please don't remove the comment. It could be useful and even prevent headaches later.

